PR #210, opened 2015-08-16
Fix register form

gnomus:
Please Test before merge


msquare:
Please name the email form helper to "form_email", it's still "form_text".

gnomus:
Done

Issue #271, opened 2016-11-02
admin_active (Active Angles) -> Unable to load event config.


anykraus:
The Active Angles page (admin_active.php) currently does not show any content. Only "Unable to load event config."
Apache's error.log is pretty full, main culprit seems to be:
`PHP Notice:  Undefined variable: shift_sum_formula in /srv/http/engelsystem/includes/pages/admin_active.php on line 108, referer: http://localhost/engelsystem/public/?p=admin_user&action=save_groups&id=3&set_locale=en_US.UTF-8`
Subsequently the SQL-Query fails as ${shift_sum_formula} is empty:

SELECT \`User\`.*, COUNT(\`ShiftEntry\`.\`id\`) as \`shift_count\`,  as \`shift_length\` \n      FROM \`User\` LEFT JOIN \`ShiftEntry\` ON \`User\`.\`UID\` = \`ShiftEntry\`.\`UID\` \n      LEFT JOIN \`Shifts\` ON \`ShiftEntry\`.\`SID\` = \`Shifts\`.\`SID\` \n      WHERE \`User\`.\`Gekommen\` = 1\n      AND (\`Shifts\`.\`end\` < 1478047417 OR \`Shifts\`.\`end\` IS NULL`

-> nothing before **as \`shift_length\`**

msquare:
Works for me. Does the problem still exist?

anykraus:
Tested with commit 1a3b4e2a334c57cc403be1d8e21fd8e162d9df5c
Works for me.

PR #272, opened 2016-11-02
fix admin_active.php not displaying (issue #271)

anykraus:
Reverts 53ad1b5110e22bdbb57960aec7d29b824dda7b5b "fix codacy unused code" partially. (where $shift_sum_formula was removed from global)

### DIFF
--- a/includes/pages/admin_active.php
+++ b/includes/pages/admin_active.php
@@ -5,7 +5,7 @@ function admin_active_title() {
 }
 
 function admin_active() {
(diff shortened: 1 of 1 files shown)


anykraus:
Why is it unused according to codacy? There is a SQL Query where it is used:
SELECT \`User\`.*, COUNT(\`ShiftEntry\`.\`id\`) as \`shift_count\`, ${shift_sum_formula} as \`shift_length\`

Issue #235, opened 2016-01-01
selection users should be possible by some fancy javascript search


jwacalex:
currently every user-selection is done by a very long selectbox which contains every user in the system. finding an user is done by the browser. 

If some jQuery addon like select2 is used, it is quite a very small amount of work to resolve this (cp. [example](https://select2.github.io/examples.html#single)).

PR #292, opened 2016-12-13
Gh 235 user selection by javascript

jwacalex:
frist draft to fix issue #235. needs styling for dark themes


includes/pages/user_shifts.php
```diff
@@ -167,7 +167,7 @@ function view_user_shifts() {
   return page([
       div('col-md-12', [
           msg(),
-          template_render('../templates/user_shifts.html', [
+          template_render('../templates/user_shtifts.html', [
```
MyIgel: ?

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "vendor/bootstrap"]
 	path = vendor/bootstrap
 	url = https://github.com/twbs/bootstrap.git
+[submodule "vendor/select2"]
```
MyIgel: I removed the submodules in feature-igel-rewrite / b839e401062b294292fdcbd7e30b79bc149fab6f so we should load the [select2/select2](https://packagist.org/packages/select2/select2) via composer

jwacalex:
fully agree with your suggestions. this was a quick and dirty approach to have a working proof of concept. if you want to integrate it in your rewrite i would suggest to close this request

msquare:
I will close the request as suggested.

PR #306, opened 2017-01-01
Fixes "Constant already defined" notice

MyIgel:
With the current config CRYPT_ALG and MIN_PASSWORD_LENGTH can not be configured in config.php


msquare:
Why move configuration away from config file?

msquare:
Then make it configurable please, by changing the constants into variables.

MyIgel:
@msquare Done

PR #314, opened 2017-02-03
Add config no reply mail

Scriptkiddi:
The default no reply email was to noreply@engelsystem.de now changed


msquare:
Thank you for coding this feature.
For what purpose is the $base_url in this patch? It seems to be an unused variable.
Could you please fix that?

Scriptkiddi:
hey removed it wanted to adapt all used urls but never got to it

Issue #372, opened 2017-12-03
Feature: Only allow shift signup a few hours in advance


der-michik:
We are curently discussing how it might be possible to distribute access to popular shifts more fairly to angels without too much manual work involved. Currently, this could either be done by disabling self-signup and scheduling manually like the VOC does it or by creating shifts ad-hoc every day, which would increase workload on heaven during the event and is error-prone.

The `signup_requires_arrival` config option already is a nice idea as it requires people to show up in person but it still favors people showing up early on day -1 or so and allows them to block all the cool shifts for days.

So the idea came up to allow signup for shifts only a certain time in advance, e.g. 24 hours, or maybe even less for those shifts which will fill up within seconds after creation anyway (like bar shifts). This would ensure that popular shifts on later days will still be available for people arriving late and in addition, it would probably reduce freeloading due to people signing up for shifts early and then forgetting about them.

As a start, a global config uption similar to `signup_requires_arrival` would be nice, but having this configurable on a per-shift basis would be even better (with configurable defaults for rooms or shift types).

msquare:
My ideas:
* Config per angeltype and/or per shifttype (leads to more complex shift state calculation since some angeltypes might be able to sign up earlier than others)
* Display countdown for signup on shift view
* Shift calendar filters free and full shifts. Is this still enough? Or do we need free+signup possible?
* Check the admin view - are admins able to signup everybody? An admin should not see the "not yet" state.
* If this feature is wanted, development and testing on own branch and test instance

cbeneke:
It might also help to only allow up to 6 hours (or so) to be signed up to in advance. (E.g. the user is signed up for a shift t+0h until t+2h and one t+4h until t+8h; then the signup for future shifts will be blocked until t+2h and then allow another signup again)

MyIgel:
The global "Signup only x hours before" has already been implemented in #613

PR #373, opened 2017-12-03
POC: Add a config option for maximum signup in advance

der-michik:
We now have `signup_advance_hours` in the configuration that only allows signup for shifts so many hours in the future. A notice is displayed in the shift view for shifts in the far future, together with an info notice on when signup for that shift will be possible.

In addition, we now have a new shift state `NOT_YET` for this purpose so we do not have to abuse `SHIFT_ENDED` anymore for `signup_requires_arrival` as well.

This is a proposed approach to #372. I hope that I did not miss anything since my PHP is not as fluent as it could be.


MyIgel:
This PR needs a rebase (or better rebuild?) to the current master before merging

msquare:
I think this was a proof of concept. Its open as a reminder to do this feature.

msquare:
Will close the pr for now. We have to redo this.

PR #379, opened 2017-12-07
fixed "add more angels" link target

claudehohl:
**Before:**

When you click "Add more angels", the sign-up form for the shift appears. I think this is not intended...?

**After:**

"Add more angels" takes you to the form where you can add or delete angels under "Needed angels".


msquare:
The behavior is acturally as intended. The idea of the button is to add any user for any angeltype to the shift, wether it is needed or not. The shift edit page can be reached with the edit icon.
Maybe this is not intuitive enough?

claudehohl:
Ah I see! I first thought of "Add more angeltypes", but yea, it makes sense as it is now.

msquare:
Ok!

PR #380, opened 2017-12-07
Code donation: shifts.js

claudehohl:
**JavaScript Shiftplan Solution(tm) using distributed WebSQL technology in the browser-based cloud - empowering angels leveraging their visible return**

TL;DR: Our secret weapon against angelsystem-overload.

You can merge and apply the updates to the live angelsystem right now, or we can sit together and have a Mate at the congress and do the updates then, when the angelsystem is probably already under heavy load and in demand. Or you can reject the pull request at all and make me feel sad... (But I had to do and finish it anyway, since I already invested so much work ☺). Besides, I enjoyed it.

## How to upgrade

1. Merge
2. execute db/websql.sql
3. enable websql in your config.php
4. ...
5. Profit!

* I already incorporated MyIgels big rewrite
* The changes to the database are in update.sql as well, so for new installations, install.sql and update.sql is sufficient
* I didn't touch the translations.po. Some strings defined here: https://github.com/claudehohl/engelsystem/blob/27da2f520999e0d649c4f86b8d6e3772f358e3e8/includes/pages/user_shifts_browser.php#L39-L61 need some translation. (Most of them are translated already).

## Changes to the database
Every table relevant for synchronisation has an additional column "updated_microseconds". Synchronisation works by telling the db "give me every record after 1512570499. Now, what happens if at the same time some records had been added *in the same second*? They'd been lost. That's why microseconds, i.e. 1512570499.293834.

PHP's default precision is 14, which should be enough to differentiate between the single timestamps. Set "precision = 16" to actually have real microseconds.

### New Table: DeleteLog
Every delete-operation needs to be synced to the browsers as well - records that have been deleted are being tracked in this table.

I think I found every db call (some calls are stored outside models/, for example updates to the Rooms in includes/page/admin_rooms.php).

Should you decide to switch back and not use websql, you can safely do so. Nothing is deleted database-wise, just some colums and a table added.


claudehohl:
You are right, the Angelsystem runs fine. It was a bad idea in the first place to even try making it faster, I shouldn't have invested so much time for...nothing. Sorry for bothering, I'm not much of use on this world.

PR #381, opened 2017-12-07
Fix sign-up to shifts by supporter

der-michik:
Any user with the `shiftentry_edit_angeltype_supporter` privilege was able to sign up any users of the correct angeltype to any shift that they could sign up themselves because the shift entry controller only checks for the global privilege an not the fact that the user is indeed supporter for the angeltype in question.

How to reproduce:

Create two unprivileged users and an angeltype. Give supporter rights for the angeltype to user 1. Give both users the `shiftentry_edit_angeltype_supporter` privilege. Create some shifts. Both users will be able to sign up each other to these shifts when they are allowed to sign up themselves. Only user 1 should be able to sign up user 2 because user 1 is supporter, but not the other way around.


msquare:
Thanks!

PR #387, opened 2017-12-25
spelling fixes

ka7:
patch contains some spelling fixes ( just in comments ) as found by a bot ( http://www.misfix.org, https://github.com/ka7/misspell_fixer ).
...might the "angels share" be with you. :)


ka7:
it seem the "error(_('You are not allowed to remove this shift entry. it seems the ..." is not used in any translation.

PR #402, opened 2017-12-28
Advertise iCal subscription options

katrinleinweber:
Hello! It took me a while to figure out how to get the iCal feed onto my phone. Maybe the solution I found will also be helpful to others. Cheers :-)

### DIFF
--- a/includes/pages/user_shifts.php
+++ b/includes/pages/user_shifts.php
@@ -287,7 +287,7 @@ function ical_hint()
 
     return heading(_('iCal export'), 2)
         . '<p>' . sprintf(
-            _('Export your own shifts. <a href="%s">iCal format</a> or <a href="%s">JSON format</a> available (please keep secret, otherwise <a href="%s">reset the api key</a>).'),
+            _('Export of shown shifts. <a href="%s">iCal format</a> or <a href="%s">JSON format</a> available (please keep secret, otherwise <a href="%s">reset the api key</a>). On Android, you can use <a href="https://icsdroid.bitfire.at/download/">ICSdroid</a> (<a href="https://f-droid.org/packages/at.bitfire.icsdroid/">F-Droid</a>), to subscribe to your shifts.'),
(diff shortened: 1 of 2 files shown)


MyIgel:
The text should be translated to german in the [default.po](https://github.com/engelsystem/engelsystem/blob/master/locale/de_DE.UTF-8/LC_MESSAGES/default.po) file

katrinleinweber:
Ah, one more question: There is a difference between `msgid "" "Export your own shifts.` and `sprintf(_('Export of shown shifts.`. Probably introduced by rebasing my local branch onto upstream/master. What is the reason for the duplication of the displayed text here? Is another commit needed, to align both strings?

msquare:
Hey Katrin, nice idea. Would you like to write a manual for engelsystem.de / https://github.com/engelsystem/engelsystem.de to link as user documentation? I don't want links to other software in the engelsystem itself. Please let me know.

katrinleinweber:
> don't want links to other software

That's not really a motivation for me to start writing documentation about a problem, because it isn't merged into exactly that GUI element it IMHO fits best ;-)   I can remove the DAVDroid snippet if you want, but I think this PR would become less useful then.

MyIgel:
Imho it's a good decision to keep the Engelsystem free of links to other products as they need to be maintained / updated / removed if the linked products change or there is a better suitable one.
In a documentation that's another point, there would be enough room for a list of potential clients on all other platforms like Apple devices and laptops.

katrinleinweber:
How is maintaining hyperlinks (or letting volunteers to do so) _within_ the system different from in the docu?

MyIgel:
The documentation gets deployed to [something].engelsystem.de so its under the control of the maintainers and can be updated. In contrast to the Engelsystem which gets released once and then stays at that version with potential outdated / wrong links?

msquare:
I put the links to the doc and added a help button to the engelsystem wich opens the doc at this point.

katrinleinweber:
Thank you :-)

PR #427, opened 2018-02-08
Use npm to handle frontend dependencies

marudor:
Now with 50% more memory consumption! (in dev)
Also another dependency management tool! \o/

On a serious note:
composer isn't really up to date with dependencies for frontend stuff. NPM obviously is.
Also this modernises some of the code. And also handles old browsers better (babel preset env)
So older browser work and we dont have to care! \o/

PR #428, opened 2018-02-08
Add Docker container support

marudor:
Docker is good for deployment.
We should use it!
Probably needs some fine tuning. 
Best to do this once https://github.com/engelsystem/engelsystem/pull/427 is merge.


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+COPY --from=composer /app/vendor /var/www/vendor
+COPY --from=themes /app/public/ /var/www/html
+COPY src/ /var/www/src/
+COPY includes/ /var/www/includes/
+COPY config/ /var/www/config/
+COPY locale/ /var/www/locale
+COPY templates/ /var/www/templates
+
+# Symlink gets copied so we delete the symlink.
```
MyIgel: Hm, the files should be build and copied by webpack

Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM composer AS composer
```
MyIgel: I think the docker files should be moved to a `contrib` folder to clean up the root folder?

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN npm install -g less
+COPY --from=composer /app/vendor /app/vendor
+COPY public/ /app/public
+COPY themes/ /app/themes
+RUN sh /app/themes/build-themes.sh
+
+FROM php:7-fpm-alpine
+RUN apk add --no-cache icu-dev
+RUN docker-php-ext-install intl
+RUN apk add --no-cache gettext-dev
```
MyIgel: Dev?

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+  db:
+    image: mariadb:latest
+    environment:
+      MYSQL_DATABASE: engelsystem
+      MYSQL_USER: engelsystem
+      MYSQL_PASSWORD: engelsystem
+      MYSQL_RANDOM_ROOT_PASSWORD: 1
+    volumes:
+      - db:/var/lib/mysql
+  # used for initial database stuff
```
MyIgel: Hm, we should import the database when the php container starts (init script) or document the docker exec .... commands to import the files

docker-compose.yml
```diff
@@ -0,0 +1,37 @@
+version: "3.3"
```
MyIgel: The containers could be build with gitlab ;)

config/config.default.php
```diff
@@ -47,7 +47,7 @@
     'display_news'            => 6,
 
     // Users are able to sign up
-    'registration_enabled'    => true,
+    'registration_enabled'    => env('REGISTRATION_ENABLED', 1),
```
MyIgel: This should be a boolean and only a 0 when disabled with an environment variable

MyIgel:
When building with `docker-compose build` the last step fails:

```
Step 25/25 : COPY vendor/twbs/bootstrap/dist/ /var/www/html/vendor/bootstrap/
ERROR: Service 'engelsystem' failed to build: COPY failed: stat /var/lib/docker/tmp/docker-builder382974979/vendor/twbs/bootstrap/dist: no such file or directory
```

MyIgel:
And another interesting Point: what about mails? The php mail() function does not work that great inside a docker container ;)

```
WARNING: [pool www] child 7 said into stderr: "sendmail: can't connect to remote host (127.0.0.1): Connection refused"
```

MyIgel:
I'm currently working on it :)

MyIgel:
I think this can be merged now :)

I will work on the GitLab build pipeline in another PR

PR #431, opened 2018-05-10
Hotfix php7 syntax error

flyingapfopenguin:
die is a reserved keyword and php7 throws a syntax error

PR #442, opened 2018-08-04
Fixes tests after #431 (Hotfix php7 syntax error)

MyIgel:
This commit fixes the tests after #431 renamed the function without running tests what causes them to terminate while running


src/Exceptions/Handler.php
```diff
@@ -76,7 +76,7 @@ public function exceptionHandler($e)
     protected function terminateApplicationImmediately($message = '')
     {
         echo $message;
-        die();
+        die(1);
```
MyIgel: This lets at least break the tests (exit code 1)

PR #443, opened 2018-08-05
UrlGeneratorInterface and "Legacy URL" support

MyIgel:
Some changes that I have laying around, build for #425 (404 on every site) and never vreated a pull request.

Should we support the old "legacy" link structure? It could be interesting for servers that have no mod_rewrite (or similar configuration) active.

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -43,6 +43,9 @@ return [
         '1' => 'Engelsystem dark'
     ],
 
+    // Rewrite URLs with mod_rewrite
+    'rewrite_urls'            => true,
+
     // Number of News shown on one site
     'display_news'            => 6,
 
--- a/src/Routing/LegacyUrlGenerator.php
+++ b/src/Routing/LegacyUrlGenerator.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Engelsystem\Routing;
+
+class LegacyUrlGenerator extends UrlGenerator
+{
+    /**
+     * @param string $path
+     * @param array  $parameters
+     * @return string
+     */
+    public function to($path, $parameters = [])
+    {
+        $page = ltrim($path, '/');
+        if (!empty($page)) {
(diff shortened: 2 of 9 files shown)


MyIgel:
@msquare: Why did you changed the `to` method to `linkTo` in 2832fe48eeff7335e734ac17b1b7d9f3c52621a8? This is really annoying while using the class (a `$url->to('...')` call is really more intuitive than a `$url->linkTo('...')`, its easier to remember and type and [other fraeworks](https://github.com/laravel/framework/blob/5.6/src/Illuminate/Contracts/Routing/UrlGenerator.php#L22) are using it too.
Yes i know that sometimes two characters are not enough but in this use case it's imho the better choice.

BTW: You broke the [tests](https://chaos.expert/engelsystem/engelsystem/-/jobs/21683), didn't change the [docstring](https://github.com/engelsystem/engelsystem/commit/2832fe48eeff7335e734ac17b1b7d9f3c52621a8#diff-1a29c774763588652842fe842190dccfR14) and adding a space between a [negation](https://github.com/engelsystem/engelsystem/commit/86b29370786de35cc40e3424e09c3cf48551ee28#diff-55a6beebd1f927788b4ac71280642cadR16) doesn't really make it more [readable](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#6-operators) ;)

msquare:
@MyIgel: Thanks for checking. I try to use codacy to ensure consistant source code. Since I am not familiar with other frameworks and the method `linkTo` only beeing used one time in the whole project I wouldn't call it annoying. The codacy check on the pull request failed and I (at least tried) to fix it.

I ran the tests locally and everything was fine(!). Guess I have to check again how to use phpunit on this project. I have no idea, why it worked for me.

Your last point, the space, was unintentional done by the eclipse autoformatter (which is configured to the PSR-2 template).

Issue #343, opened 2017-08-30
New MVC architecture


MyIgel:

MyIgel:
Mostly implemented, the rest is just even more refactoring 🎉

PR #452, opened 2018-08-20
Rebuild database

MyIgel:
* `Db` class refactoring to use `illuminate/database` (required for #343)
* Implemented database migrations
* Moved the deploy script to `bin/` folder
* Fixed potential issues with session deserialization for shiffts filter if class does not exist / is not loaded / included
* Config extends from `Illuminate\Support\Fluent` for better access
* More refactorings


includes/model/Shifts_model.php
```diff
@@ -182,7 +182,7 @@ function NeededAngeltypes_by_ShiftsFilter(ShiftsFilter $shiftsFilter)
 /**
  * @param array $shift
  * @param array $angeltype
- * @return array|null
+ * @return array
  */
 function NeededAngeltype_by_Shift_and_Angeltype($shift, $angeltype)
 {
```
msquare: This is a selectOne - if nothing found, null should be returned.

MyIgel:
I'll have a look after fixing some code in my current working branch ;)

Issue #312, opened 2017-01-21
Indequate translation


mrm2m:
If you are admin and want to put someone else into a group of angel types you have to click on "join", which is quite missleading as it suggests to join yourself (at least that's what I would have in mind). Never the less you could understand it as "join user" which would include all users. So in English you can join (yourself) or you can join someone else. 

The German translation says "mitmachen" which can be translated as "join" in the meaning of participate. So you yourself can "mitmachen", but you can't "mitmachen" someone else.

So I'd suggest "join angel" and the German translation "Engel hinzufügen" for this button.

opensamba:
Hi,
we need to edit the german translation in line 108 here in 

https://github.com/engelsystem/engelsystem/blob/master/locale/de_DE.UTF-8/LC_MESSAGES/default.po:108

I tried with poedit but I've got several issue and I stopped



![screenshot from 2017-04-25 14-48-28](https://cloud.githubusercontent.com/assets/1799909/25386057/49368a36-29c6-11e7-9b6f-6bb9847466c7.png)


`ASSERT INFO:
./src/gtk/dataview.cpp(211): assert "gtk_tree_selection_get_select_function(m_selection) == func" failed in CheckCurrentSelectionFunc(): selection function has changed unexpectedly, review this code!

BACKTRACE:
[1] wxDataViewCtrl::DoSetCurrentItem(wxDataViewItem const&)
[2] wxString wxString::Format<wxString>(wxFormatString const&, wxString)
[3] wxEvtHandler::TryHereOnly(wxEvent&)
[4] wxEvtHandler::ProcessEventLocally(wxEvent&)
[5] wxEvtHandler::ProcessEvent(wxEvent&)
[6] wxEvtHandler::ProcessPendingEvents()
[7] wxAppConsoleBase::ProcessPendingEvents()
[8] wxApp::DoIdle()
[9] g_main_context_dispatch
[10] g_main_loop_run
[11] gtk_main
[12] wxGUIEventLoop::DoRun()
[13] wxEventLoopBase::Run()
[14] wxAppConsoleBase::MainLoop()
[15] wxEntry(int&, wchar_t**)
[16] __libc_start_main

`

Nos-:
Hi,
is the translation error (the error shown above) solved yet?

> So I'd suggest "join angel" and the German translation "Engel hinzufügen" for this button.

I would like to do improve the translation this way.

opensamba:
[...]

PR #456, opened 2018-08-27
Improvement of the german translation

Nos-:
See also my comment according to issue #312 .


locale/de_DE.UTF-8/LC_MESSAGES/default.po
```diff
@@ -2341,6 +2342,7 @@ msgstr "Eintragen"
 msgid "ended"
 msgstr "vorbei"
 
+# Wie ist dies zu verstehen bitte?
```
jens1o: It means `Engels` need to go to the event so they are marked for being there. They only can apply for shifts (if this mode is enabled) when it is proven that they are at the event.

Nos-:
Ok, I try to make a better PR.

PR #504, opened 2018-11-27
Modernize JS Parts

marudor:
I've also updated bootstrap and charts.js to the latest version (within their major version)


msquare:
`error An unexpected error occurred: "https://registry.yarnpkg.com/event-stream/-/event-stream-3.3.6.tgz: Request failed \"404 Not Found\"".`

msquare:
event-stream was that package with malware problems right?

marudor:
Yes it was.

PR #512, opened 2018-12-03
styling 35c3: Lighten the font color

MyIgel:


zichy:
Do you have a screenshot of how `lighten(@gray, 15%)` would look like? The contrast ratio should be at least 4.5:1 to meet the [WCAG](https://www.w3.org/TR/WCAG21/) criteria. `#757574` would be fine, `#BBBBB9` is even better.

MyIgel:
The `lighten(@gray, 15%)` sets the color to `#737372` so it should be enough ;)
Imho a `#BBBBB9` would look (at that point) like white and that would not represent the color scheme of the congress.

msquare:
Ok, lets go. Usability before design specifications again.

MyIgel:
At least the c3groc does not see any strict design requirements ;)

> Feel free to choose a light gray :)
> Bad contrast is "form fucks function" but so aesthetic ;)
>
> Source: https://twitter.com/c3groc/status/1069689215313694721

PR #514, opened 2018-12-05
Use another, more capable date picker, get rid of some inline JS and make the time inputs in the shifts view more pretty

der-michik:
In my browser, the time input fields over the shifts table render quite ugly:

schichten-input-type-time

This is due to `<input type="time">`. Therefore, I switched the date picker for a more capable date and time picker, did the same trick as in the date input fields in #509 to keep the current behaviour and HTML 5 semantics when JS is disabled and added a time picker. If a user now clicks on the input field, a time picker widget appears (of course, entering the time manually is still possible).

In addition, I threw away all that inline JS around the date and time pickers and put it into static JS. The min/max date and locale are fed to the JS via data attributes. The date pickers are initialized lazily on the first click into the input field since there are some fields (mainly departure date) that can be empty and if we instantiate the date picker right away, it would not leave the field empty but set the value of the input to the first valid date or so.

The "Now" button still works as intended, the time period buttons below the time inputs work as well.

I have a few more frontend tricks up my sleeve, especially with regard to the ever growing list of rooms and angel types. Please let me know if you would like a PR for that as well.


MyIgel:
I like the idea (and already thought about fixing all this ugly inline mess) -> so :+1: for that!

I would say: All improvements for the room/angeltype selects are welcome? ;)

PR #532, opened 2018-12-21
WIP: Add div and css to expand shift panel on hover fixes #526

### LINKED ISSUE #526: Not all angels shown in shift overview
In the user-shift overview, when the event is too short, only the first angeltypes are shown and the other are getting cut off.
grafik

wonko21:
Implementation idea stolen from http://jsfiddle.net/6K7t4/24/

### DIFF
--- a/includes/view/ShiftCalendarShiftRenderer.php
+++ b/includes/view/ShiftCalendarShiftRenderer.php
@@ -38,24 +38,25 @@ class ShiftCalendarShiftRenderer
 
         return [
             $blocks,
-            div(
-                'shift panel panel-' . $class . '" '
-                . 'style="height: '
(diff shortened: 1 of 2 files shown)


msquare:
Is it possible to do this on click and not on mouseover? Mouseover happens accidently and you may not be able to acces the shift below.

wonko21:
Very possible but might involve some js.Do you have a db dump somewhere that I can use to go through "real" use cases? 
@MyIgel  already wihed for the cards not beeing cut of on overflow.

msquare:
Create 2h shifts that need 3 different angel types for your testcase. JS for this would be ok if without js the shift panel is not extended.

wonko21:
closing in favor of new pr

PR #546, opened 2018-12-27
WiP: Expand shift panel

wonko21:


MyIgel:
The changes look and work great (even mobile).
I have only point: The dotted line is shown a centimeter under every shift even if it is not expandable which is irritating.

image

msquare:
No hurry! Thank you for contributing!

msquare:
@wonko21 any new ideas on this?

Issue #582, opened 2019-03-18
Gitlab CI broken due to failing test job


k127:
Hi there, after forking this nice project and trying to build it on my own Gitlab space, sadly the `test` job is failing for a **deprecation reason** [1], and the very same problem happens also on Gitlab `engelsystem/engelsystem` CI [2]:
> An <del>un</del>expected error occurred, a team of untrained monkeys has been dispatched to deal with it.Exception: Code: 0, Message: The "Symfony\Bridge\PsrHttpMessage\Factory\DiactorosFactory" class is deprecated since symfony/psr-http-message-bridge 1.2, use PsrHttpFactory instead., File: vendor/symfony/psr-http-message-bridge/Factory/DiactorosFactory.php:14, Trace: …

Since the respective committed code changes [3] do not explicitly hint to a connection with the problem, I'd consider this a package incompatibility issue due to an unversioned `composer.lock` file.

EDIT:
This problem came up with a version change from `v1.1.1` to `v1.2.0` of package `symfony/psr-http-message-bridge` [4] that sneaked into this project within the last 7 days.

---
[1] <https://chaos.expert/engelsystem/engelsystem/-/jobs/24541>
[2] <https://chaos.expert/engelsystem/engelsystem/pipelines/21641>
[3] <https://chaos.expert/engelsystem/engelsystem/commit/69ac1b2d2115a41afe2bd4527abadda70063019a>
[4] <https://github.com/symfony/psr-http-message-bridge/compare/v1.1.1...v1.2.0#diff-7e5b3956b6ae2a5924719b9d4eecb8cdR14>

MyIgel:
Jup, i know this (Got an email from my ci some days ago for this and another one). The "problem" is indeed the added warning in twig >=2.7 and could be fixed (have a half pr for that but this requires some [upstream changes](https://github.com/twigphp/Twig-extensions/issues/239)) with the version pinning.

The reason for not versioning the composer.lock is that it needs to be updated and working on the newest bugfix releases of software is normally a good thing while developing (and the master is normally the "should be kinda stable" branch). The release packages have a working composer.lock ;)

k127:
Ok, looks like a good strategy since version pinning (of those two packages) provides the fix :)

PR #583, opened 2019-03-18
Fix CI test job failures induced by deprecation errors

k127:
This fixes the `test` CI job
* https://github.com/engelsystem/engelsystem/issues/582
* another deprecation error as of twig/twig v2.7.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "psr/log": "^1.0",
         "swiftmailer/swiftmailer": "^6.1",
         "symfony/http-foundation": "^3.3",
-        "symfony/psr-http-message-bridge": "^1.0",
+        "symfony/psr-http-message-bridge": "~v1.1.0",
         "twig/extensions": "^1.5",
         "twig/twig": "^2.5",
         "zendframework/zend-diactoros": "^1.7"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -33,9 +33,9 @@
         "psr/log": "^1.0",
         "swiftmailer/swiftmailer": "^6.1",
         "symfony/http-foundation": "^3.3",
-        "symfony/psr-http-message-bridge": "^1.0",
+        "symfony/psr-http-message-bridge": "~v1.1.0",
```
MyIgel: This pinning is not needed to fix the error message?

composer.json
```diff
@@ -33,9 +33,9 @@
...
         "twig/extensions": "^1.5",
-        "twig/twig": "^2.5",
+        "twig/twig": "~v2.6.0",
```
MyIgel: just using `"~2.6.0"` is imho more readable ;)

composer.json
k127: It is. One deprecation error comes with version `v1.2.0`…

composer.json
k127: …while the other one comes with `twig/twig` `v2.7.0`.

composer.json
MyIgel: Ah, ok, now i see ^^
Thats the code that i fixed at the weekend and didn't create a PR because it was not tested ;)
-> Will create one soon

k127: Fyi, on my Gitlab fork all CI jobs pass for this PR.

MyIgel: But its a preventable workaround :P

PR #610, opened 2019-06-04
Use ini timezone setting

### LINKED ISSUE #586: Arrival date seems to be timezone dependent
If the php date.timezone is unset/set to UTC, the arrival date selector shows the configured buildup date als selectable, but shows an error, if that date is selected as date of arrival.
Setting date.timezone to the same timezone the event is in fixes this.

MyIgel:
This should work around some potential problems when setting the timezone.

For #586 (Arrival date seems to be timezone dependent) to be fixed this timezone should be somehow interpreted by the client. @marudor?

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -115,7 +115,7 @@ return [
     'max_freeloadable_shifts' => 2,
 
     // Local timezone
-    'timezone'                => env('TIMEZONE', 'Europe/Berlin'),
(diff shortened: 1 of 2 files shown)


marudor:
No Idea about the PHP Setting. But it depends how you transfer the time to the client.
All Formats but ISO 8601 will end up being interpreted as local time of the user.

MyIgel:
The PHP part should be fine (now), the interesting question is how to solve the timing on the client selects which use `chart.js`

PR #614, opened 2019-06-07
Ansible

foosinn:
This adds a ansible role with a basic playbook to setup engelsystem.

Please review.


foosinn:
if you want i can squash the commits.

MyIgel:
I am not shure if that's the right place to put Ansible roles, IMHO an own repository is the better way to go.
Another point is: Who will maintain them in the future? I don't use Ansible so I can't test it nor change things when it needs to be changed.

msquare:
Thank you for that pull request! Since I don't use Ansible either, I agree with @MyIgel to place these config files into another repository.
Would you mind to take a look at https://github.com/engelsystem/engelsystem-ansible, a brandnew empty repo for the Ansible stuff you did and reopen your pull request there? I would also like to give you rights for the repo if you want to maintain Ansible support!

foosinn:
Sounds great! Seems like it's not posssible to open a pull request on an empty repo.

> I would also like to give you rights for the repo if you want to maintain Ansible support!

I am happy to maintain it. Thanks.

Can you give me access to the repo. I will add some tests and upload it afterwards. Thanks!

msquare:
You should have an invite to the repo. Thank you. Please close this pull request, if you are ready.

MyIgel:
Another Point is if this rules are enough for the most usecases (in meaning of complexity vs. features) as for example internal or external database / mail /  whatever server etc

foosinn:
I noticed you are using gitlab for ci? is this a private instance or the gitlab.com?

MyIgel:
We are using chaos.expert which is a GitLab run by members of the Hackerspace Frankfurt ;)

PR #627, opened 2019-07-25
Users overview: Order by planned arrival/departure dates

MyIgel:


MyIgel:
> No time needed
Hm, fair point ;)

PR #638, opened 2019-08-18
Fix CI

marudor:
Less broke itself with a new version. (https://github.com/less/less.js/issues/3414)
I've also removed redundant yarn install in the dockerfile and changed the node version to 10.
(Installing yarn with apk also installed node 10 so nothing changed)

Also added yarn.lock to the dockerfile => wont happen again.


MyIgel:
Adding the ´yarn.lock` file to docker means that it has to be versioned which is currently not the case as it is excluded by `*.lock` in the `.gitignore` file

PR #651, opened 2019-09-06
Save time of past shifts as worklog if they get deleted

### LINKED ISSUE #640: Don't delete work when deleting shifts
When Rooms are deleted shifts are deleted too, even past shifts with work that has been done by volunteers. Past work shouldn't be deleted (even more so, as this isn't logged in a way, that allows recovery). I propose to automatically add a Work Log on shift deletion if the shift starts in the past.

Right now we can't recover from this issue and actual work has been lost.

MyIgel:
Fixes #640 (Don't delete work when deleting shifts)


msquare:
Let's add something like this to the shift deletion mail text:
Since the deleted shift was already done, we added a worklog entry instead, to keep your work hours correct.

includes/model/UserWorkLog_model.php
```diff
@@ -121,6 +121,41 @@ function UserWorkLog_create($userWorkLog)
...
+    foreach ($shift['ShiftEntry'] as $entry) {
+        if ($entry['freeloaded']) {
+            continue;
+        }
+        $type = AngelType($entry['TID']);
+
+        UserWorkLog_create([
+            'user_id'        => $entry['UID'],
+            'work_timestamp' => $shift['start'],
+            'work_hours'     => ($shift['end'] - $shift['start']) / 60 / 60,
```
msquare: This should be doubled on night shifts using the configured rules.

MyIgel: Hmm, fair point but this would mean the worklog hours wont change even if the admin changes the night shifts times but as we have no data about it but that's better than not adding them ;)

PR #659, opened 2019-10-11
Sets Monday as first day of the week

### LINKED ISSUE #630: Datepickers for planned arrival/departure date starts with sunday
Calendar for planned arrival/depature date starts with Sunday when browser language is English.

This might lead to problems because angels from Germany assume that Monday is the first day of the week, when they do not check their calendar or the description.

weeman1337:
Moment.js introduced a locale update function in 2.12. So I've updated Moment to this version.

Monday is now first day of the week - even if English has been selected.

Closes #630

### DIFF
--- a/resources/assets/js/vendor.js
+++ b/resources/assets/js/vendor.js
@@ -10,6 +10,13 @@ require('./forms');
 require('./sticky-headers');
 require('./moment-countdown');
 
+moment.updateLocale('en', {
+    week : {
+        dow : 1, // Monday is the first day of the week.
+        doy : 4  // The week that contains Jan 4th is the first week of the year.
+    }
+});
+
 $.ajaxSetup({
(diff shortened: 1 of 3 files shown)


resources/assets/js/vendor.js
```diff
@@ -10,6 +10,13 @@ require('./forms');
 require('./sticky-headers');
 require('./moment-countdown');
 
+moment.updateLocale('en', {
```
MyIgel: Shouldn't `const moment = require('moment');` be used first?

weeman1337: `require('imports-loader?this=>window&define=>false&exports=>false!moment');` makes `moment` globally available.

MyIgel: Hm, ok, TIL ;)

PR #663, opened 2019-10-15
Migrates the News items to an Eloquent model

weeman1337:
Just want to make this visible to other devs.

Also I want to use this PR to discuss some points / answer questions.


weeman1337:
@MyIgel do you have any idea why the user test complains about the `news` table to be absent? The debugger says that the test setup runs into the "news" migration. Is there any trick for the tests?

weeman1337:
> do you have any idea why the user test complains about the news table to be absent?

I answer myself :) SQLite is case insensitive for identifiers. It deletes the news table right after creating it :smiley:

MySQL is case insensitive in most cases. We should be aware of this for future migrations.

db/migrations/2019_10_15_000000_create_news_table.php
```diff
@@ -0,0 +1,145 @@
...
+
+    private function createNewNewsTable(): void
+    {
+        $this->schema->create('news', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('title', 150);
+            $table->text('text');
+            $table->boolean('is_meeting');
+            $table->unsignedInteger('user_id');
+            $table->foreign('user_id')
```
MyIgel: There is a `$this->referencesUser();` for that usecase ;)

db/migrations/2019_10_15_000000_create_news_table.php
```diff
@@ -0,0 +1,145 @@
...
+        $previousNewsRecords = $this->schema
+            ->getConnection()
+            ->table('PreviousNews')
+            ->get();
+
+        foreach ($previousNewsRecords as $previousNews) {
+            $date = Carbon::createFromTimestamp($previousNews->Datum)
+                ->format(self::DB_DATETIME_FORMAT);
+
+            DB::table('news')->insert([
```
MyIgel: Better to use the News model here (`$news = new News([data]); $news->save();`)

src/Models/News/News.php
```diff
@@ -0,0 +1,37 @@
...
+ * This class represents a news item.
+ */
+class News extends Model
+{
+    protected $casts = [
+        'is_meeting' => 'boolean',
+    ];
+
+    protected $attributes = [
+        'is_meeting' => false,
```
MyIgel: What?

db/migrations/2019_10_15_000000_create_news_table.php
weeman1337: I've seen that one. The reason I didn't use it is `->onDelete('cascade')`. We can do that but then all news written by a user are gone as well when deleting him. ¯\\\_(ツ\)_/¯ 

One solution could be moving shifts, news etc. to another user on delete. But that should be done somehere else.

db/migrations/2019_10_15_000000_create_news_table.php
weeman1337: Are you sure? In a migration you only want to do plain database stuff. Why?: Imagine the News models changes that much or we decide to rename or delete it in the future. On setup all migrations run one after another. But because of the missing/modifies News model it crashes.

src/Models/News/News.php
weeman1337: What what? :)

db/migrations/2019_10_15_000000_create_news_table.php
MyIgel: We could allow `null` as the user ID here (and add an option `$setNull=false` to the method) but I personally don't think that's a relevant point as not that many users get deleted during a event

db/migrations/2019_10_15_000000_create_news_table.php
MyIgel: On the other side in that way you can validate the data inside the model, hmm

src/Models/News/News.php
MyIgel: Why is it required to set the value here?

db/migrations/2019_10_15_000000_create_news_table.php
weeman1337: As discussed, I'll set it to delete cascade.

src/Models/News/News.php
weeman1337: It's the default value when you create a new instance using the model functions

weeman1337:
@MyIgel what is the reason for this method https://github.com/engelsystem/engelsystem/blob/master/src/Models/BaseModel.php#L20 ? Since there is a static `create` as well that creates and stores a model it is now no longer possible to call it.

I it's just because static is evil: I agree you shouldn't use static stuff in general - but that's the way how the Laravel DB / Model stuff works.

src/Models/News/News.php
```diff
@@ -0,0 +1,37 @@
...
+        'title',
+        'text',
+        'is_meeting',
+        'user_id',
+    ];
+
+    /**
+     * @return BelongsTo
+     */
+    public function user(): BelongsTo
```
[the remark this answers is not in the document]
weeman1337: ✓ done

db/migrations/2019_10_15_000000_create_news_table.php
```diff
@@ -0,0 +1,142 @@
...
+use stdClass;
+
+/**
+ * This migration creates the news table and copies the existing News table records to the new one.
+ */
+class CreateNewsTable extends Migration
+{
+    use ChangesReferences, Reference;
+
+    private const DB_DATETIME_FORMAT = 'Y-m-d H:i:s';
```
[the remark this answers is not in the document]
weeman1337: ✓ done

db/migrations/2019_10_15_000000_create_news_table.php
```diff
@@ -0,0 +1,142 @@
...
+                'created_at' => $date,
+                'updated_at' => $date,
+            ]);
+        }
+    }
+
+    /**
+     * Recreates the previous News table, copies back the data and drops the new news table.
+     */
+    public function down(): void
```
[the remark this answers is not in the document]
weeman1337: ✓ done

db/migrations/2019_10_15_000000_create_news_table.php
```diff
@@ -0,0 +1,142 @@
...
+        $newsRecords = $this->schema
+            ->getConnection()
+            ->table('new_news')
+            ->get();
+
+        foreach ($newsRecords as $newsRecord) {
+            $date = Carbon::createFromFormat(self::DB_DATETIME_FORMAT, $newsRecord->created_at)
+                ->getTimestamp();
+
+            DB::table('News')->insert([
```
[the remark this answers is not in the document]
weeman1337: ✓ done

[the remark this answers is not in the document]
weeman1337: ✓ done

db/migrations/2019_10_15_000000_create_news_table.php
```diff
@@ -0,0 +1,142 @@
...
+    {
+        /** @var stdClass[] $previousNewsRecords */
+        $previousNewsRecords = $this->schema
+            ->getConnection()
+            ->table('PreviousNews')
+            ->get();
+
+        foreach ($previousNewsRecords as $previousNews) {
+            $date = Carbon::createFromTimestamp($previousNews->Datum)
+                ->format(self::DB_DATETIME_FORMAT);
```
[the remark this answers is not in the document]
weeman1337: ✓ done

MyIgel:
There is a static `::create()`?

weeman1337:
> There is a static ::create()?

yes: https://laravel.com/docs/6.x/eloquent#mass-assignmen

 It has been also available in former versions.

> Once we have made the attributes mass assignable, we can use the create method to insert a new record in the database. The create method returns the saved model instance:

MyIgel:
> yes: https://laravel.com/docs/6.x/eloquent#mass-assignmen

Jeah indeed it gets forwarded to `\Illuminate\Database\Eloquent\Builder::create` :D

I have fixed that in the current master

weeman1337:
@MyIgel thanks for merging. Tiny side note: Because I want to lean the "engelsystem code style" please just send a note next time what to change :)

MyIgel:
> [...] just send a note next time what to change :)

Will do that but here I had to test stuff out by myself so it was easier (as i already had it changed) to just commit it ;)

PR #664, opened 2019-10-19
added symfony/console as dev dependency for var dump server

allentje:
#### What does this PR do?
This PR allows the var dump server to be used in cli by adding the required "symfony/console" dev dependency .
Composer does not install dependencies dev-dependencies.

#### Where should the reviewer start?
Install composer packages and then start var dump server in the cli

#### How should this be manually tested?
- [ ] `composer install`
- [ ]  `./vendor/bin/var-dump-server`

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -263,4 +263,11 @@ return [
             . 'you want to to contribute, have found any [bugs](https://github.com/engelsystem/engelsystem/issues) '
             . 'or need help.'
     ],
+
+    // var dump server
+    'var_dump_server' => [
+        'host' => '127.0.0.1',
+        'port' => '9912',
+        'enable' => false,
+    ],
 ];
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -54,6 +54,27 @@ at least `phpdbg -qrr`(which has problems with switch case statements) as using 
 php -d pcov.enabled=1 vendor/bin/phpunit --testsuite Unit --coverage-text
 ```
 
+### Var Dump server
+Symfony Var Dump server is configured to allow for easier debugging. It is not meant as a replacement for xdebug but can actually be used together with xdebug.
+This Var Dump Server is especially useful for when you want to debug a request without messing up the output e.g API calls ot HTML layout.
+
+To use simply call the method `dump` and pass the arguments in exactly the same way you would when using `var_dump`.
+
+This will send the output to the Var Dump server which can be viewed in the terminal. 
+This does however require that you start the var-dump-server otherwise the output will be printed in your browser
+
+You can also `dump` and `die` if you wish to not let your code continue any further by calling the `dd` method
+
+To view the output of `dump` call the following commands:
+
+```bash
+vendor/bin/var-dump-server
+# or for running in docker
+docker exec -it engelsystem_dev_es_php_fpm_1 vendor/bin/var-dump-server
+```
+
+For more information check out the Var Dump Server documentation: [Symfony VarDumper](https://symfony.com/components/VarDumper)
+
 ## Translation
 We use gettext. You may use POEdit to extract new texts from the sourcecode.
 Please config POEdit to extract also the twig template files using the following settings: https://gist.github.com/jlambe/a868d9b63d70902a12254ce47069d0e6
--- a/config/app.php
+++ b/config/app.php
@@ -33,6 +33,7 @@ return [
         \Engelsystem\Helpers\VersionServiceProvider::class,
         \Engelsystem\Mail\MailerServiceProvider::class,
         \Engelsystem\Http\HttpClientServiceProvider::class,
+        \Engelsystem\Helpers\DumpServerServiceProvider::class
     ],
 
     // Application middleware
--- a/src/Helpers/DumpServerServiceProvider.php
+++ b/src/Helpers/DumpServerServiceProvider.php
@@ -1,0 +1,55 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Engelsystem\Helpers;
+
+use Engelsystem\Config\Config;
+use Engelsystem\Container\ServiceProvider;
+use Symfony\Component\VarDumper\Cloner\VarCloner;
+use Symfony\Component\VarDumper\Dumper\CliDumper;
+use Symfony\Component\VarDumper\Dumper\ContextProvider\CliContextProvider;
+use Symfony\Component\VarDumper\Dumper\ContextProvider\SourceContextProvider;
+use Symfony\Component\VarDumper\Dumper\HtmlDumper;
+use Symfony\Component\VarDumper\Dumper\ServerDumper;
+use Symfony\Component\VarDumper\VarDumper;
+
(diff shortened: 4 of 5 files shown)


MyIgel:
TIL: There is a var dump server :tada: ;)

Some documentation how to use it would be nice, esp. how to set it up (also in composer) and how to configure it.

Imho `xdebug` is normally the better option for debugging than `dump()` ^^

allentje:
@MyIgel I have released that it was added to the require dev in composer but I don't think it was configured.
I guess this needs to be configured first and then I have to figure out how to run it in docker before requesting a PR.

I will close the PR for now.

allentje:
I have configured the var dump server and added a small documentation to the README.md

config/config.default.php
```diff
@@ -193,5 +193,11 @@
         'Contribution' => 'Please visit [engelsystem/engelsystem](https://github.com/engelsystem/engelsystem) if '
             . 'you want to to contribute, have found any [bugs](https://github.com/engelsystem/engelsystem/issues) '
             . 'or need help.'
+    ],
+    // Var dump server
+    'vardump_server' => [
+        'host' => '127.0.0.1',
+        'port' => '9912',
+        'enable' => true,
```
MyIgel: This enables the functionality by default what should slow down development, i would just disable it by default

composer.json
```diff
@@ -44,7 +44,8 @@
         "dms/phpunit-arraysubset-asserts": "^0.1.0",
         "filp/whoops": "^2.3",
         "phpunit/phpunit": "^8.1",
-        "symfony/var-dumper": "^4.3"
+        "symfony/var-dumper": "^4.3",
+        "symfony/console": "~3.4|~4.0"
```
MyIgel: `v` comes after `c` please :stuck_out_tongue_winking_eye:

includes/application.php
```diff
@@ -33,6 +33,24 @@
     $app->bind(HandlerInterface::class, 'error.handler.development');
     ini_set('display_errors', true);
     error_reporting(E_ALL);
+
```
MyIgel: I would definitely move this to a `VarDumpServerServiceProvider` to keep this legacy and bootstrapping code clean

includes/application.php
```diff
@@ -33,6 +33,24 @@
...
+    // setup var dump server to use for easier debugging
+    $vardumpServerConfig = config('vardump_server');
+    if ($vardumpServerConfig['enable'] === true) {
+
+        $cloner = new Symfony\Component\VarDumper\Cloner\VarCloner();
+        $fallbackDumper = \in_array(\PHP_SAPI, ['cli', 'phpdbg']) ? new Symfony\Component\VarDumper\Dumper\CliDumper() : new Symfony\Component\VarDumper\Dumper\HtmlDumper();
```
MyIgel: You couldnt you just use `\Symfony\Component\VarDumper\VarDumper` here?

And using `imports` shortens line lengths :grin:)

config/config.default.php
allentje: My thought was to always have this prepared during development.
As for the performance side, this does not really have any significant impact.
I can however invert it

includes/application.php
allentje: sure, will refactor this

config/config.default.php
MyIgel: That would be nice as i personally don't like it if software connects randomly to some ports before doing some fallback magic

MyIgel:
I see that this PR has only smaller changes, do you want to fix them or should i have a look?

allentje:
Sure feel free, I don't have my laptop with me this weekend so can't commit anyway. :)

MyIgel:
It should be something like

```
<?php

namespace Engelsystem\Helpers;

use Engelsystem\Config\Config;
use Engelsystem\Container\ServiceProvider;
use Symfony\Component\VarDumper\Cloner\VarCloner;
use Symfony\Component\VarDumper\Dumper\CliDumper;
use Symfony\Component\VarDumper\Dumper\ContextProvider\CliContextProvider;
use Symfony\Component\VarDumper\Dumper\ContextProvider\SourceContextProvider;
use Symfony\Component\VarDumper\Dumper\HtmlDumper;
use Symfony\Component\VarDumper\Dumper\ServerDumper;
use Symfony\Component\VarDumper\VarDumper;

class DumpServerServiceProvider extends ServiceProvider
{
    public function register()
    {
        /** @var Config $config */
        $config = $this->app->get('config');

        if (
            !$config->get('environment') == 'development'
            || !class_exists(ServerDumper::class)
        ) {
            return;
        }

        // setup var dump server to use for easier debugging
        $varDumpServerConfig = config('var_dump_server');
        if (!$varDumpServerConfig['enable']) {
            $cloner = new VarCloner();
            $fallbackDumper = in_array(PHP_SAPI, ['cli', 'phpdbg']) ? new CliDumper() : new HtmlDumper();

            $dumper = new ServerDumper(
                'tcp://' . $varDumpServerConfig['host'] . ':' . $varDumpServerConfig['port'],
                $fallbackDumper,
                [
                    'cli'    => new CliContextProvider(),
                    'source' => new SourceContextProvider(),
                ]
            );

            VarDumper::setHandler(
                function ($var) use ($cloner, $dumper) {
                    $dumper->dump($cloner->cloneVar($var));
                }
            );
        }
    }
}
```

But as i currently don't have the time to rewrite a testable version of that (and move all the changes to the new files) it would be great if you could maybe find the time to re-add it to the current master?

allentje:
Ok I'll do it this evening.

allentje:
I am running on a new machine so I had to compose and build everything (docker) but unfortunately I do not know how to access the site through the browser.
Which url/ip do I need to use and do I need to setup anything on my machine directly?

MyIgel:
Hi,
the Engelsystem can be accesses using http://localhost:5000 (or better http://egelsystem.localhost:5000 ;) )

The error means that something is not working, here i would suspect that you have to run the database migrations (`docker exec -it engelsystem_dev_es_workspace_1 bin/migrate up`).

To see the error you can use `docker logs -f engelsystem_dev_es_php_fpm_1`

allentje:
ah ok, you guys don't push lock files for composer. 
Apparently my branch still had an old lock file that was installing some old version of libs which prevented me from updating the db.

I now have a running system :+1:

allentje:
I moved the code into a helper. Tested it and all works as expected. Just have to remember to change the config to allow dumping into the server

MyIgel:
It looks great so far, the only missing thing are tests for the class and passing style checks ;)
(looking at the removed new line in application.php) ^^

allentje:
>passing style checks
Fixed the application.php style. I had to convert the first doc block to an inline block to completely validate the file ;)
I'll write a test tomorrow for the helper class

allentje:
so I have created the test. Its not perfect but it tests if the server can be enabled and disabled which should be sufficient.
Its a but difficult to test the actual var dump server code.

allentje:
Hi there,
This PR is still open. Have you managed to check the latest changes?

MyIgel (approved):

PR #667, opened 2019-10-24
Fix some translation-related issues

lujoga:
As the `default.po` files are now located in `$locale/` instead of `$locale/LC_MESSAGES/`, one `../` had to be removed from the X-Poedit-Basepath. Also, the `auth.not-found` and `validation.password.confirmed` strings should be resolved before being passed to the template, so Poedit recognizes the `__('<msgid>')` signature.


resources/views/pages/login.twig
```diff
@@ -34,7 +34,7 @@
                     <div class="panel-body">
                         {{ msg() }}
                         {% for message in errors|default([]) %}
-                            {{ m.alert(__(message), 'danger') }}
+                            {{ m.alert(message, 'danger') }}
```
MyIgel: This breaks the translation output of automatically generated validation messages in the form of `validation.[field].[type]`

resources/views/pages/password/reset.twig
```diff
@@ -9,7 +9,7 @@
         <h1>{{ __('Password recovery') }}</h1>
 
         {% for message in errors|default([]) %}
-            {{ m.alert(__(message), 'danger') }}
+            {{ m.alert(message, 'danger') }}
```
MyIgel: Same as `resources/views/pages/login.twig` (this needs a cleanup anyway to prevent code duplication..)

resources/views/pages/credits.twig
```diff
@@ -17,15 +17,10 @@
                 <h2>{{ __('Source code') }}</h2>
                 <p>{{ __('Version: _%s_', [version])|markdown }}</i></p>
                 <p>
-                    {{ __('The original engelsystem was written by
-[cookie](https://github.com/cookieBerlin/engelsystem).
-It was then completely rewritten and enhanced by [msquare](https://notrademark.de) (maintainer) and
-[MyIgel](https://myigel.name).')|markdown }}
+                    {{ __('The original engelsystem was written by [cookie](https://github.com/cookieBerlin/engelsystem). It was then completely rewritten and enhanced by [msquare](https://notrademark.de) (maintainer) and [MyIgel](https://myigel.name).')|markdown }}
```
MyIgel: I hate those long lines but it seems that it can't be parsed otherwise (or are there any options for that?)

MyIgel: Generally I prefer unique IDs like `credits.credit`, `credits.contributors` and so on instead of splitting them up more (which is imho useless here)

lujoga:
Now there is a separate `validation.po` file, so one can update the `default.po` file from sources without the auto-generated messages being removed.

resources/views/pages/login.twig
lujoga: (see comment below)

resources/views/pages/password/reset.twig
lujoga: (see comment below)

resources/lang/de_DE/validation.po
```diff
@@ -0,0 +1,36 @@
...
+"Project-Id-Version: Engelsystem 2.0\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"Last-Translator: \n"
+"Language: de_DE\n"
+
+msgid "auth.not-found"
```
MyIgel: Is not really a "validation" module output so imho this should be splittet further? In that case it would be something like `auth.po` (or something more general like `pages.po`? but that would mean that we have a ton of files in the end which could be problematic? Eventually we should have a look at the performance impact of that)

resources/lang/en_US/default.po
```diff
@@ -22,29 +22,5 @@ msgstr ""
 #~ msgid "auth.no-password"
 #~ msgstr "Please enter a password."
 
-msgid "auth.not-found"
```
MyIgel: See above

resources/lang/de_DE/default.po
```diff
@@ -1536,10 +1536,10 @@ msgstr "Pflichtfeld!"
 #~ msgstr "Gib bitte ein Passwort ein."
 
 #: includes/pages/guest_login.php:418
-msgid "auth.not-found"
```
MyIgel: Imho this should be removed to keep the file clear

src/Helpers/Translation/TranslationServiceProvider.php
```diff
@@ -93,11 +99,12 @@ public function getTranslator(string $locale): GettextTranslator
 
     /**
      * @param string $locale
+     * @param string $name
      * @return string
      */
-    protected function getFile(string $locale): string
+    protected function getFile(string $locale, string $name = 'default'): string
```
MyIgel: It would be nice if you could also extend the unit tests to load multiple files

resources/lang/de_DE/validation.po
lujoga: According to [this](https://www.php.net/manual/de/book.gettext.php#115370) `*.mo` files are cached, so in a production environment the performance impact of having multiple translation files is probably neglectable. In my opinion, though, two files should be sufficient: a `default.po` and (as a more general alternative to naming it `validation.po`) an `errors.po`.

MyIgel: If we would use phps own `gettext` that would be true but we are using the [gettext/gettext](https://github.com/php-gettext/Gettext/) Package which allows to differentiate between "No value translated" and "Translated value is the same".

I think that even `errors.po` won't be the right wording as validation messages could be also positive feedback

lujoga: I mistook the `Engelsystem\Helpers\Translation\GettextTranslator` for the `Gettext\GettextTranslator`, the latter of which uses PHP's `gettext`. My bad.

Still, I need a name for the auto-gen messages file. Would a general name like `additional.po` or `other.po` be a solution?

MyIgel: At least `additional.po` would be much clearer ;)

src/Helpers/Translation/TranslationServiceProvider.php
lujoga: Done

src/Helpers/Translation/TranslationServiceProvider.php
```diff
@@ -82,6 +83,11 @@ public function getTranslator(string $locale): GettextTranslator
             } else {
                 $translations->addFromPoFile($file);
             }
+            if (Str::endsWith($validationFile, '.mo')) {
```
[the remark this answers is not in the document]
lujoga: The files are now added in a foreach loop over an array of filenames (`['default', 'additional']` as you suggested).

PR #672, opened 2019-10-31
Implement registration as controller and Twig view

lujoga:
These commits implement the user registration with a Controller class to handle requests on `/register` and a Twig template for the registration form.
Apart from porting registration from a function to a Controller class, I added the option to enable "pending registrations", i.e. authenticated users don't have to set a password when creating new users. The newly created user is sent a password reset link, so they can set their password themselves. For this, I changed 'password recovery' to 'password reset', as it is a more neutral wording when a password reset can not only happen for lost passwords, but also to set one in the first place.


resources/views/macros/form.twig
```diff
@@ -1,7 +1,12 @@
 {% macro input(name, label, type, required) %}
     <div class="form-group">
         {% if label %}
-            <label for="{{ name }}">{{ label }}</label>
+            <label for="{{ name }}">
+                {{ label }}
+                {% if required|default(false) %}
+                    <span class="text-info glyphicon glyphicon-warning-sign"></span>
```
MyIgel: I don't think that we should use the warning sign here, a smaller asterisk would be a better choice as this makro will be used every where and some input fields will be obviously required like a search input, so eventually use another parameter like `show_required` and only append an asterisk to the name (as this should be common knowledge)?

And PS: You should have used your own `entry_required` macro ;)

resources/views/macros/form.twig
```diff
@@ -16,3 +21,18 @@
 {% macro submit(label) %}
     <button type="submit" class="btn btn-default">{{ label|default(__('form.submit')) }}</button>
 {% endmacro %}
+
+{% macro entry_required() %}
```
MyIgel: As its only used on one page its not a global makro, only a page one

resources/views/pages/register.twig
```diff
@@ -0,0 +1,133 @@
...
+        <form action="" enctype="multipart/form-data" method="post">
+            {{ csrf() }}
+            <div class="row">
+                <div class="col-md-6">
+                    <div class="row">
+                        <div class="col-sm-4">
+                            {{ f.input('login', __('Nick'), 'text', true) }}
+                            <span class="help-block">
+                                {{ m.glyphicon('info-sign') }}
+                                {{ __('Use up to 23 letters, numbers, connecting punctuations or spaces for your nickname.') }}
```
MyIgel: One point of moving stuff to templates is to change the texts to use an abbreviation text like `[page_name].[part]` so that not all translations have to be updated only because the English text changes.
And another point is that (more or less changeable) numbers/names should be generalized to use placeholders like `{{ __('The user has %u bananas!', count(user.bananas)) }}` so that refactoring is less painful.

resources/views/pages/register.twig
```diff
@@ -0,0 +1,133 @@
...
+                    {% endif %}
+                    <h4>
+                        {{ f.entry_required() }}
+                        = {{ __('Entry required!') }}
+                    </h4>
+                </div>
+            </div>
+            <div class="form-group">
+                <button class="btn btn-primary" type="submit" name="submit">
+                    {{ __('Register') }}
```
MyIgel: As above: This should be `form.register` or (as it's only used here) `register.register`

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+    /**
+     * @param Request $request
+     * @return Response
+     */
+    public function postRegister(Request $request): Response
+    {
+        $isPending = $this->config->get('enable_pending_registrations') && $this->auth->user() != null;
+
+        if (!($isPending || $this->config->get('registration_enabled'))) {
+            error(__('registration.disabled'));
```
MyIgel: This is a legacy method and as such should not be used in new code in `src/` especially because it creates a dependency to not tested code.

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+        }
+
+        $rules = [
+            'login'     => 'required|between:1:23|regex:/^([\p{L}\p{N}\-_. ]+)$/ui',
+            'email'     => 'required|email',
+            'mobile'    => 'optional|phone',
+        ];
+
+        if ($this->config->get('enable_tshirt_size')) {
+            $rules = array_merge($rules, [
```
MyIgel: Why is `array_merge` used here? A simple `$rules[] = ...` does the same.

Same for all following.

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+                'firstname'   => 'optional|alpha',
+                'lastname'  => 'optional|alpha',
+            ]);
+        }
+
+        $data = $this->validate($request, $rules);
+
+        if (User::whereName($data['login'])->count() > 0) {
+            $this->session->set('errors', array_merge($this->session->get('errors', []), ['registration.exists.login']));
```
MyIgel: I really have to implement proper error message handling :joy:

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+            return $this->showForm();
+        }
+
+        if (!$isPending && $data['password'] !== $data['password_confirmation']) {
+            $this->session->set('errors', array_merge($this->session->get('errors', []), ['validation.password.confirmed']));
+
+            return $this->showForm();
+        }
+
+        $user = new User([
```
MyIgel: This part can be moved to a factory or user provider or whatever, it's just to much code and logic for a controller which should be just better glue code

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+        $state = new State([]);
+        if ($this->config->get('autoarrive')) {
+            $state->arrived = true;
+            $state->arrival_date = new Carbon();
+        }
+        $state->user()
+            ->associate($user)
+            ->save();
+
+        DB::insert('INSERT INTO `UserGroups` (`uid`, `group_id`) VALUES (?, -20)', [$user->id]);
```
MyIgel: This is not testable behavior and thus direct queries have to be avoided, esp. es it makes this controller only testable against a real Database and not only a sqlite in memory one.

And another point is that database queries in a controller are a no go...

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+                'lastname'  => 'optional|alpha',
+            ]);
+        }
+
+        $data = $this->validate($request, $rules);
+
+        if (User::whereName($data['login'])->count() > 0) {
+            $this->session->set('errors', array_merge($this->session->get('errors', []), ['registration.exists.login']));
+
+            return $this->showForm();
```
MyIgel: Returning that early means that the user has to send the form multiple times, a better way would be to run all checks and after that show the page again if any check fails

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+            ->save();
+
+        $state = new State([]);
+        if ($this->config->get('autoarrive')) {
+            $state->arrived = true;
+            $state->arrival_date = new Carbon();
+        }
+        $state->user()
+            ->associate($user)
+            ->save();
```
MyIgel: As the User model has `withDefault()` on the model, saving of this is optional

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+use Engelsystem\Models\User\Settings;
+use Engelsystem\Models\User\State;
+use Engelsystem\Models\User\User;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Collection;
+use Parsedown;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
+
+class RegistrationController extends BaseController
```
MyIgel: This controller needs unit tests

src/Controllers/RegistrationController.php
```diff
@@ -0,0 +1,268 @@
...
+            $this->mail->sendViewTranslated(
+                $user,
+                'Password reset',
+                'emails/password-reset',
+                ['username' => $user->name, 'reset' => $reset]
+            );
+        }
+
+        $userAngelTypesInfo = [];
+        foreach (AngelTypes() as $angelType) {
```
MyIgel: Same here for example: A dependency on older code.

resources/lang/de_DE/default.po
```diff
@@ -10,7 +10,7 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "X-Generator: Poedit 1.8.11\n"
-"X-Poedit-KeywordsList: __;_e:1,2;translate;translatePlural\n"
+"X-Poedit-KeywordsList: __;_e:1,2;translate;translatePlural;sendViewTranslated:2\n"
```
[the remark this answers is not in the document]
lujoga: `additional.po` should only contain messages that can't be extracted from code, i.e. when message ids appear in places other than function calls (adding them to the `errors` array, dynamically generated ids). As `sendViewTranslated` *can* be used as a keyword for gettext, it should be added to the list and messages translated by it should go into `default.po`.

MyIgel:
Hi, did you had time to have a look at the remaining issues?

msquare:
Ping @lujoga?

msquare:
Will be done with #1160.

PR #675, opened 2019-11-12
Introduce the NewsComments model

weeman1337:
Next model migration..


weeman1337:
I've seen that you have added PSR-12 compliance. Cool :) Less discussion regarding code style.

PR #679, opened 2019-11-20
WIP: Basic (temporary) API implementation

cbluoss:
# Content
This implements some basic api features around shift data.
To be more precise the following endpoints are implemented by now:
* /api/angeltypes/my
* /api/angeltypes
* /api/shifts/my
* /api/shifts/free/{start:.+}/until/{stop:.+}
* /api/shifts/by/angeltype/{angeltypeid:.+}

All endpoints are protected by the already implemented api_key and the new privilege 'view_api' which is not a default privilege for any user or group (yet).

## Motivation
We at c3cert spending quite some time on all large events (where we _have to_ provide service 24/7) to fill up all our shifts across different rooms and angel types.
The web view (on smaller devices) isn't optimal in this regard  and won't offer a proper way to poke our team automagically.
I'm sure this feature is useful for other OCs as well.

### DIFF
--- a/config/routes.php
+++ b/config/routes.php
@@ -30,5 +30,6 @@ $route->addGroup('/api/v2019-alpha', function (RouteCollector $route) {
     $route->get('/shifts/my', 'ApiController@getMyShifts');
     $route->get('/shifts/free/{start:.+}/until/{stop:.+}', 'ApiController@getShiftsFree');
     $route->get('/shifts/by/angeltype/{angeltypeid:\d+}', 'ApiController@getShiftsByAngelType');
+    $route->get('/shift/{shiftid:\d+}', 'ApiController@getShiftById');
     $route->get('[/{resource:.+}]', 'ApiController@index');
 });
--- a/config/routes.php
+++ b/config/routes.php
@@ -24,9 +24,11 @@ $route->get('/metrics', 'Metrics\\Controller@metrics');
 $route->get('/stats', 'Metrics\\Controller@stats');
 
 // API
-$route->get('/api/angeltypes/my', 'ApiController@getMyAngelTypes');
-$route->get('/api/angeltypes', 'ApiController@getAngelTypes');
-$route->get('/api/shifts/my', 'ApiController@getMyShifts');
-$route->get('/api/shifts/free/{start:.+}/until/{stop:.+}', 'ApiController@getShiftsFree');
-$route->get('/api/shifts/by/angeltype/{angeltypeid:.+}', 'ApiController@getShiftsByAngelType');
-$route->get('/api[/{resource:.+}]', 'ApiController@index');
+$route->addGroup('/api/v2019-alpha', function (RouteCollector $route) {
+    $route->get('/angeltypes/my', 'ApiController@getMyAngelTypes');
+    $route->get('/angeltypes', 'ApiController@getAngelTypes');
+    $route->get('/shifts/my', 'ApiController@getMyShifts');
+    $route->get('/shifts/free/{start:.+}/until/{stop:.+}', 'ApiController@getShiftsFree');
+    $route->get('/shifts/by/angeltype/{angeltypeid:.+}', 'ApiController@getShiftsByAngelType');
+    $route->get('[/{resource:.+}]', 'ApiController@index');
+});
--- a/config/routes.php
+++ b/config/routes.php
@@ -24,4 +24,9 @@ $route->get('/metrics', 'Metrics\\Controller@metrics');
 $route->get('/stats', 'Metrics\\Controller@stats');
 
 // API
+$route->get('/api/angeltypes/my', 'ApiController@getMyAngelTypes');
+$route->get('/api/angeltypes', 'ApiController@getAngelTypes');
+$route->get('/api/shifts/my', 'ApiController@getMyShifts');
+$route->get('/api/shifts/free/{start:.+}/until/{stop:.+}', 'ApiController@getShiftsFree');
+$route->get('/api/shifts/by/angeltype/{angeltypeid:.+}', 'ApiController@getShiftsByAngelType');
 $route->get('/api[/{resource:.+}]', 'ApiController@index');
--- a/config/routes.php
+++ b/config/routes.php
@@ -29,6 +29,6 @@ $route->addGroup('/api/v2019-alpha', function (RouteCollector $route) {
     $route->get('/angeltypes', 'ApiController@getAngelTypes');
     $route->get('/shifts/my', 'ApiController@getMyShifts');
     $route->get('/shifts/free/{start:.+}/until/{stop:.+}', 'ApiController@getShiftsFree');
-    $route->get('/shifts/by/angeltype/{angeltypeid:.+}', 'ApiController@getShiftsByAngelType');
+    $route->get('/shifts/by/angeltype/{angeltypeid:\d+}', 'ApiController@getShiftsByAngelType');
     $route->get('[/{resource:.+}]', 'ApiController@index');
 });
--- a/db/migrations/2019_11_19_000000_add_api_privilege.php
+++ b/db/migrations/2019_11_19_000000_add_api_privilege.php
@@ -12,15 +12,11 @@ class AddApiPrivilege extends Migration
     public function up()
     {
         $connection = $this->schema->getConnection();
-
         // Add api permissions
-        $connection->insert(
-            '
-                INSERT INTO `Privileges` (`id`, `name`, `desc`) VALUES (NULL, \'view_api\', \'API accessible.\');
-            ',
+        $connection->table('Privileges')->insert(
             [
-                -40, // Shift Coordinator
-                43,  // admin_user_worklog
+                'name' => 'view_api',
+                'desc' => 'API accessible.'
             ]
         );
     }
@@ -30,11 +26,8 @@ class AddApiPrivilege extends Migration
      */
     public function down()
     {
+        $connection = $this->schema->getConnection();
         // Remove view_api permission
-        $this->schema->getConnection()->delete(
-            'DELETE FROM `Privileges`
-                WHERE `name` = \'view_api\'
-                )'
-        );
+        $connection->table('Privileges')->where('name', '=', 'view_api')->delete();
     }
 }
--- a/db/migrations/2019_11_19_000000_add_api_privilege.php
+++ b/db/migrations/2019_11_19_000000_add_api_privilege.php
@@ -11,6 +11,9 @@ class AddApiPrivilege extends Migration
      */
     public function up()
     {
+        if (!$this->schema->hasTable('Privileges')) {
+            return;
+        }
         $connection = $this->schema->getConnection();
         // Add api permissions
         $connection->table('Privileges')->insert(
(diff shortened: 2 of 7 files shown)


config/routes.php
```diff
@@ -24,4 +24,9 @@
 $route->get('/stats', 'Metrics\\Controller@stats');
 
 // API
+$route->get('/api/angeltypes/my', 'ApiController@getMyAngelTypes');
```
MyIgel: `$route->addGroup('/api', function (RouteCollector $route) { [...]` should be used here

MyIgel: Also as mentioned above: Atm the API can't be implemented in a stable way so it should use a version namespace (normally `/api/v[\d+]/`), so in that case it should be `/api/01-alpha` or something as it will change unexpectedly.

db/migrations/2019_11_19_000000_add_api_privilege.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * Run the migration
+     */
+    public function up()
+    {
+        $connection = $this->schema->getConnection();
+
+        // Add api permissions
+        $connection->insert(
```
MyIgel: `$this->schema->getConnection()->table('User')->insert([ [...]` should be used here instead of a less flexible sql query, see other migrations for reference

db/migrations/2019_11_19_000000_add_api_privilege.php
```diff
@@ -0,0 +1,40 @@
...
+        );
+    }
+
+    /**
+     * Reverse the migration
+     */
+    public function down()
+    {
+        // Remove view_api permission
+        $this->schema->getConnection()->delete(
```
MyIgel: Same here, this should NOT be a sql query

src/Controllers/ApiController.php
```diff
@@ -3,6 +3,18 @@
 namespace Engelsystem\Controllers;
 
 use Engelsystem\Http\Response;
+use Engelsystem\Http\Exceptions\HttpForbidden;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+
+# Legacy functions
+/* Doing their job, but adding a proper namespaces breaks legacy code :( */
```
MyIgel: That's the point where this code can't be fully unit tested as it depends on legacy code that has no code coverage and depends on database entries that can't be used in a in memory sqlite database. The only thing where i could see them be used would be some kind of factory class but that feels more like a hack than a good workaround as  is the main point why no API exists at the moment

src/Controllers/ApiController.php
```diff
@@ -27,4 +39,114 @@ public function index()
             ->withHeader('content-type', 'application/json')
             ->withContent(json_encode(['error' => 'Not implemented']));
     }
+
+
+    /**
+     * @return User
+     *
+     * Warning: I am no php coder. This code isnt intended for eternity but making our lives @36c3 a bit easier.
```
MyIgel: Not shure if thats a proper comment to leave in the code? ;)

src/Controllers/ApiController.php
```diff
@@ -27,4 +39,114 @@ public function index()
...
+     * Code is mostly harvested from /includes/controller/shifts_controller.php/shifts_json_export_controller()
+     */
+    protected function doApiAuth($key)
```
MyIgel: Using the `$permissions` attribute would be the better option (and make the `ErrorHandler` Middleware api aware as in format the error page as JSON instead of HTML)

src/Controllers/ApiController.php
```diff
@@ -27,4 +39,114 @@ public function index()
...
+
+        return $this->response
+            ->setStatusCode(200)
+            ->withHeader('content-type', 'application/json')
+            ->withContent(json_encode($shifts));
+    }
+
+    /**
+     * @Route(“/api/shifts/my”)
+     * doesnt work? dunno why result is empty.
```
MyIgel: The it should be a WIP PR

src/Controllers/ApiController.php
```diff
@@ -27,4 +39,114 @@ public function index()
...
+        $user = $this->doApiAuth($key);
+
+        // Shifts_by_angeltype uses only the id, required a AngelType though.
+        $angelTypeId = (int)$request->getAttribute('angeltypeid');
+        $angelType = AngelType($angelTypeId);
+        $shifts = Shifts_by_angeltype($angelType);
+
+        return $this->response
+            ->setStatusCode(200)
+            ->withHeader('content-type', 'application/json')
```
MyIgel: Using something like `$this->createResponse([...])` prevents useless code duplication ;)

src/Controllers/ApiController.php
```diff
@@ -27,4 +39,114 @@ public function index()
...
+
+        if (!auth()->can('view_api')) {
+            throw new HttpForbidden('{"error":"Not allowed"}', ['content-type' => 'application/json']);
+        }
+
+        return $user;
+    }
+
+    /**
+     * @Route(“/api/shifts/my”)
```
MyIgel: That should be `"` instead of `“`,  same for following

src/Controllers/ApiController.php
```diff
@@ -27,4 +39,114 @@ public function index()
...
+     * doesnt work? dunno why result is empty.
+     */
+    public function getShiftsByAngelType(Request $request)
+    {
+        $key = $request->input('key');
+        $user = $this->doApiAuth($key);
+
+        // Shifts_by_angeltype uses only the id, required a AngelType though.
+        $angelTypeId = (int)$request->getAttribute('angeltypeid');
```
MyIgel: Imho just using `\d+` in the route definition should be enough

src/Controllers/ApiController.php
```diff
@@ -27,4 +39,114 @@ public function index()
...
+
+    /**
+     * @Route(“/api/shifts/free”)
+     */
+    public function getShiftsFree(Request $request)
+    {
+        $key = $request->input('key');
+        $user = $this->doApiAuth($key);
+        // Shifts_by_angeltype uses only the id, required a AngelType though.
+        $start = (int)$request->getAttribute('start');
```
MyIgel: Using timestamps might be the easiest thing (as the legacy database model uses that) but using a datetime is more readable and thus should be imho preferred

src/Controllers/ApiController.php
```diff
@@ -27,4 +39,114 @@ public function index()
...
+     * @return User
+     *
+     * Warning: I am no php coder. This code isnt intended for eternity but making our lives @36c3 a bit easier.
+     * Code is mostly harvested from /includes/controller/shifts_controller.php/shifts_json_export_controller()
+     */
+    protected function doApiAuth($key)
+    {
+        $user = auth()->apiUser('key');
+        if (
+            !preg_match('/^[\da-f]{32}$/', $key)
```
MyIgel: This check is useless, the user gets already authenticated over the API by using the `apiUser` method (which has the benefit that a logged in user isn't required to provide an API key)

db/migrations/2019_11_19_000000_add_api_privilege.php
cbluoss: I actually copied code from another migration (2018_01_01_000004_cleanup_group_privileges.php)  ;)
You still got a point though.

cbluoss:
Will work on the issues the next days. Thanks for your feedback, your work is very appreciated.

I totally get why a unstable/temporary api isn't perfect at all and I'm happy when the legacy code is finally gone for good and we can work on new stuff.

db/migrations/2019_11_19_000000_add_api_privilege.php
MyIgel: Yeah, some of the old migrations need to be cleaned up too ;)

src/Controllers/ApiController.php
cbluoss: Just copied from shifts_json_export_controller(). 
Should be likewise removed for user convenience (after resolving this PR)

src/Controllers/ApiController.php
cbluoss: I'm still uncertain about this. While it's unreadable for humans, it's easy to use in any kind of code and you instantly recognize it as a timestamp.
String representations of datetimes are often annoying due to a quadrillion different format variations even though human readable (but not necessary unambiguously)

src/Controllers/ApiController.php
cbluoss: $permissions gets validated before any authentication (with an api key) can happen.

cbluoss: Works, somehow.
* can't get rid of doApiAuth() for getMy[Shifts|AngelTypes] because $this->user is NULL if no api_key provided (and i honestly do not understand what makes the difference here)
* the resulting 403  page will be rendered as html (with proper content_type) and I've no idea where that happens.

MyIgel: That gets added by the `src/Middleware/ErrorHandler.php` (as stated above) that needs to be changed to accept json responses

src/Controllers/ApiController.php
MyIgel: Jeah, thats some old dead code that never got removed as changing everything thats inside the `includes/` path it is a pain

src/Controllers/ApiController.php
MyIgel: As thats only an API i think we could just define that a datetime has to be `DateTimeInterface::RFC3339` formatted and thus look like `'Y-m-d\TH:i:sP'` which should solve the interpretation problem

db/migrations/2019_11_19_000000_add_api_privilege.php
```diff
@@ -0,0 +1,33 @@
...
+class AddApiPrivilege extends Migration
+{
+    /**
+     * Run the migration
+     */
+    public function up()
+    {
+        $connection = $this->schema->getConnection();
+        // Add api permissions
+        $connection->table('Privileges')->insert(
```
MyIgel: I Would format it as
```php
$this->schema->getConnection()
->table('Privileges')
->[...]
```
but thats only my personal preference

src/Controllers/ApiController.php
```diff
@@ -27,4 +41,83 @@ public function index()
             ->withHeader('content-type', 'application/json')
             ->withContent(json_encode(['error' => 'Not implemented']));
     }
+
+
+    /**
+     * @return User
```
MyIgel: `Undefined class 'User'`

src/Controllers/ApiController.php
cbluoss: Why not both ;)

db/migrations/2019_11_19_000000_add_api_privilege.php
cbluoss: I don't. But you could collapse it in this case with just a single insert.

src/Controllers/ApiController.php
cbluoss: That's correct. We won't get test coverage for this, but the controller has no side-effects to anything else. 
Should it break the api isn't working, which is always the case if the api isn't implemented at all.
The controller could be moved to /includes as well if 100% coverage for /src is more important than the project structure.

cbluoss:
83454d3 Is the only change which may effect other parts of the engelsystem, as it adds another field to the query result.
This should be useful in general as Shifts are usually displayed with all ShiftEntries, including the UserID  for links and the associated username which required an additional query for every ShiftEntry.

msquare:
This will be reimplemented in #1158

PR #681, opened 2019-11-25
Add models ide helper script

weeman1337:
I've frickeled the [ide model helper](https://github.com/barryvdh/laravel-ide-helper) into a script. Since it's not 100 % laravel I write a separate file and not into the models directly. It's quite convenient during development.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
         "vlucas/phpdotenv": "^3.3"
     },
     "require-dev": {
+        "barryvdh/laravel-ide-helper": "^2.6",
         "dms/phpunit-arraysubset-asserts": "^0.1.0",
         "filp/whoops": "^2.3",
         "phpunit/phpunit": "^8.1",
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ _vimrc_local.vim
 /docker/dev/.env
 
 # Project files
+/_ide_helper_models.php
 /config/config.php
 /test/coverage
 /public/coverage
--- a/README.md
+++ b/README.md
@@ -193,6 +193,10 @@ The `bin/migrate` script can be used to import and update the database of the en
 
 For more information on how to use it call `./bin/migrate help`
 
+##### [bin/gen_ide_helper_models](bin/gen_ide_helper_models)
+
+This script dumps the ide helper model docs to [_ide_helper_models](./_ide_helper_models).
+
 ### Translation
 We use gettext. You may use POEdit to extract new texts from the sourcecode. Please config POEdit to extract also the twig template files using the following settings: https://gist.github.com/jlambe/a868d9b63d70902a12254ce47069d0e6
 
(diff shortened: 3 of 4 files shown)


bin/gen_ide_helper_models.php
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env php
```
MyIgel: It would be nicer if the file has no `.php` extension ;)

bin/gen_ide_helper_models.php
```diff
@@ -0,0 +1,23 @@
...
+require_once __DIR__ . '/../includes/application.php';
+
+/**
+ * This script generates the models helper file "_ide_helper_models.php" into the project root.
+ */
+
+$config = $app->get('config');
+$config['ide-helper.model_locations'] = ['src/Models'];
+
+$fileSystem = new \Illuminate\Filesystem\Filesystem();
```
MyIgel: I'm  a big friend of the `use` statement

composer.json
```diff
@@ -49,7 +49,8 @@
         "filp/whoops": "^2.3",
         "phpunit/phpunit": "^8.1",
         "squizlabs/php_codesniffer": "^3.5",
-        "symfony/var-dumper": "^4.3"
+        "symfony/var-dumper": "^4.3",
+        "barryvdh/laravel-ide-helper": "^2.6"
```
MyIgel: `b` should be somewhere at the start of the list

bin/gen_ide_helper_models.php
```diff
@@ -0,0 +1,23 @@
...
+$command = new \Barryvdh\LaravelIdeHelper\Console\ModelsCommand($fileSystem);
+$command->setLaravel($app);
+$command->run(
+    new \Symfony\Component\Console\Input\ArrayInput([
+        '--nowrite' => true,
```
MyIgel: Is there a reason why it should not be written directly?

weeman1337: Yes like you found out in the other PR. This means writing to the model files. But then also stuff like the id property will be written. The idea here is: use it for generation to a file where you can grab the parts you need from.

bin/gen_ide_helper_models.php
weeman1337: ✓ done

composer.json
weeman1337: ✓ done

bin/gen_ide_helper_models.php
MyIgel: Well, that makes sense :D

MyIgel:
This commit is merge ready but I would wait with merging until https://github.com/barryvdh/laravel-ide-helper/commit/25b61c56fed54cdec2459045bbae04ade5c8fe2c is in a `barryvdh/laravel-ide-helper` release (should be `v2.6.6`)

cbluoss:
v2.6.6 of barryvdh/laravel-ide-helper was released last week.

PR #682, opened 2019-11-29
PHP 7.4 compatibility, PHP min version 7.2, Twig 3

MyIgel:
The goal is to support to also run on PHP 7.4 as thats the newest (and thus best) PHP version out there :heart:

* Changed type of `PASSWORD_DEFAULT` from `int` to `?string`
* Older xdebug code coverage
* Twig - can now be updated: The extension package can be removed as it's not used anymore (and deprecated anyway)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ To report bugs use [engelsystem/issues](https://github.com/engelsystem/engelsyst
 ## Installation
 
 ### Requirements
- * PHP >= 7.1
+ * PHP >= 7.2
    * Required modules:
      * dom
      * json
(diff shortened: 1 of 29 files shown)


weeman1337:
What is the supported PHP version range of the engelsystem?  
Should it be backward compatible or can I now use stuff that is >= PHP 7.4 only?

MyIgel:
> Should it be backward compatible or can I now use stuff that is >= PHP 7.4 only?

It should be definitive backwards compatible to the version thats stated in the readme and composer.json so at the moment 7.1, these changes are just needed to fully work with php 7.4 (esp. the password algorithm part is pretty anoying and definitely unexpected...) and updating twig something that i had planned for a long time (the commit is from 10 Months ago...).

But changing the minimum version to php 7.2 (as 7.1 is [outdated](https://www.php.net/supported-versions.php) from tomorrow on) is possible if required

MyIgel:
I changed it to 7.2 as supporting a not maintained version is useless.
If  there is a great usecase for a higher version that can be discussed ;)

PR #684, opened 2019-11-29
Questions model migration

weeman1337:
The next model migration

### DIFF
--- a/db/migrations/Reference.php
+++ b/db/migrations/Reference.php
@@ -3,6 +3,7 @@
 namespace Engelsystem\Migrations;
 
 use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Database\Schema\ColumnDefinition;
 
 trait Reference
 {
@@ -20,10 +21,11 @@ trait Reference
      * @param string    $targetTable
      * @param string    $fromColumn
      * @param bool      $setPrimary
+     * @return ColumnDefinition
      */
-    protected function references(Blueprint $table, $targetTable, $fromColumn, $setPrimary = false)
+    protected function references(Blueprint $table, $targetTable, $fromColumn, $setPrimary = false): ColumnDefinition
     {
-        $table->unsignedInteger($fromColumn);
+        $col = $table->unsignedInteger($fromColumn);
 
         if ($setPrimary) {
             $table->primary($fromColumn);
@@ -33,5 +35,7 @@ trait Reference
             ->references('id')->on($targetTable)
             ->onUpdate('cascade')
             ->onDelete('cascade');
+
+        return $col;
     }
 }
--- a/src/Models/Question.php
+++ b/src/Models/Question.php
@@ -1,0 +1,64 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Engelsystem\Models;
+
+use Engelsystem\Models\User\User;
+use Engelsystem\Models\User\UsesUserModel;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+/**
+ * @property integer   $id
+ * @property string    $text
+ * @property string    $answer
+ * @property integer   $answerer_id
+ * @property-read User $answerer
+ * @method static Builder|Question whereAnswer($value)
+ * @method static Builder|Question whereAnswererId($value)
+ * @method static Builder|Question whereId($value)
+ * @method static Builder|Question whereQuestion($value)
+ */
+class Question extends BaseModel
+{
+    use UsesUserModel;
+
+    /** @var string[] */
+    protected $fillable = [
+        'user_id',
+        'text',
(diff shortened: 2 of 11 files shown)


db/migrations/Reference.php
```diff
@@ -33,5 +35,7 @@ protected function references(Blueprint $table, $targetTable, $fromColumn, $setP
             ->references('id')->on($targetTable)
             ->onUpdate('cascade')
             ->onDelete('cascade');
+
+        return $col;
```
weeman1337: I'm returning the column definition here.  
For example the question answerer field should be nullable.

db/migrations/Reference.php
```diff
@@ -17,13 +18,14 @@ protected function referencesUser(Blueprint $table, $setPrimary = true)
 
     /**
      * @param Blueprint $table
-     * @param string    $targetTable
-     * @param string    $fromColumn
-     * @param bool      $setPrimary
+     * @param string $targetTable
+     * @param string $fromColumn
+     * @param bool $setPrimary
+     * @return ColumnDefinition
```
MyIgel: Hmm, i think that using spaces is better readable but thats another point for a code style discussion ;)

src/Models/Question.php
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public function answerer(): BelongsTo
+    {
+        return $this->belongsTo(User::class, 'answerer_id');
+    }
+
+    /**
+     * @return Builder
+     */
+    public static function findUnAnswered(): Builder
```
MyIgel: I would name it `findNotAnswered` but thats only my personal preference

src/Models/Question.php
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Engelsystem\Models;
+
+use Engelsystem\Models\User\User;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+/**
+ * @property integer $id
+ * @property integer $enquirer_id
```
MyIgel: I would name that just user?

src/Models/Question.php
```diff
@@ -0,0 +1,75 @@
...
+ * @property string $question
```
MyIgel: Or maybe text? That would be my personal preference and is imho better readable (as its shorter) than `$question->question`

src/Models/Question.php
```diff
@@ -0,0 +1,75 @@
...
+ * @property string $answer
+ * @property integer $answerer_id
+ * @property-read \Engelsystem\Models\User\User $answerer
+ * @property-read \Engelsystem\Models\User\User $enquirer
+ * @method static Builder|\Engelsystem\Models\Question newModelQuery()
```
MyIgel: These are (now) in the parent model so no need for these methods ;)

src/Models/Question.php
```diff
@@ -0,0 +1,75 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+
+/**
+ * @property integer $id
+ * @property integer $enquirer_id
+ * @property string $question
+ * @property string $answer
+ * @property integer $answerer_id
+ * @property-read \Engelsystem\Models\User\User $answerer
```
MyIgel: `use Engelsystem\Models\User\User;` and so on ^^

src/Models/User/User.php
```diff
@@ -116,4 +120,22 @@ public function newsComments(): HasMany
...
+    public function questions(): HasMany
+    {
+        return $this->hasMany(Question::class, 'enquirer_id')
+            ->where('enquirer_id', $this->id);
+    }
+
+    /**
+     * @return HasMany
+     */
+    public function answers(): HasMany
```
MyIgel: Hmm, would probably use `answeredQurestions()` here as it interferes less with potential "answers" to other stuff like comments?

MyIgel:
Btw, also `src/Controllers/Metrics/Stats.php` needs to be changed accordingly (searching for `Questions` helps ^^)

db/migrations/Reference.php
weeman1337: I think that as well ✓

src/Models/Question.php
weeman1337: ✓

src/Models/Question.php
weeman1337: ✓

src/Models/Question.php
weeman1337: ✓

src/Models/Question.php
weeman1337: Event shorter `unanswered` ;)

src/Models/User/User.php
weeman1337: I'll make this more specific

PR #685, opened 2019-11-30
Rebuild Schedule import

MyIgel:
* Reimplemented old import
* Use URL instead of file for identification/mapping
* Only imports rooms (no deletion/change)
* Use `.xml` instead of `.ics`
* Use data objects to represent the schedule export
* Allow for more than one import to support multiple sources
* Added talk language after the shift title
* Replaced `redirect('url')` with `throw_redirect('url')`
* added `back()` and `redirect($url)` methods that return a `Response` object
* Add `->with($name, $data)` and `->withInput($formData)` to `Response` class

Only caveat: This removes the connection from old imports as the schedule URL is not available during migration


resources/lang/de_DE/default.po
```diff
@@ -2777,3 +2777,60 @@ msgstr ""
...
+msgstr "Programm laden"
+
+msgid "form.import"
+msgstr "Importieren"
+
+msgid "schedule.import.title"
+msgstr "Programm importieren"
+
+msgid "schedule.import.text"
+msgstr "Dieser Import erstellt Räume and erstellt, aktualisiert und löscht Schichten anhand des schedule.xml exportes."
```
msquare: *und

msquare:
I get
```
ErrorException thrown with message "Trying to get property 'id' of non-object"

Stacktrace:
#25 ErrorException in /var/www/includes/pages/schedule/ImportSchedule.php:460
#24 Engelsystem\Exceptions\Handler:errorHandler in /var/www/includes/pages/schedule/ImportSchedule.php:460
#23 Engelsystem\Controllers\Admin\Schedule\ImportSchedule:shiftsDiff in /var/www/includes/pages/schedule/ImportSchedule.php:374
#22 Engelsystem\Controllers\Admin\Schedule\ImportSchedule:getScheduleData in /var/www/includes/pages/schedule/ImportSchedule.php:123
```
when i rename an imported room in the engelsystem and then try the import again. Test schedule was the schedule.xml from unit tests, but altered to 01.01.2020 (because 2042 explodes the unix timestamp fields)

msquare:
When i change the duration of an event in schedule.xml the shift end is not updated.

msquare:
"The only caveat" means we have to do a major version release after merging due to breaking changes (losing frab import informations on imported shifts and rooms).

config/app.php
```diff
@@ -26,10 +26,12 @@
         \Engelsystem\Middleware\RequestHandlerServiceProvider::class,
         \Engelsystem\Middleware\SessionHandlerServiceProvider::class,
         \Engelsystem\Http\Validation\ValidationServiceProvider::class,
+        \Engelsystem\Http\RedirectServiceProvider::class,
 
         // Additional services
         \Engelsystem\Helpers\VersionServiceProvider::class,
         \Engelsystem\Mail\MailerServiceProvider::class,
+        \Engelsystem\Http\GuzzleServiceProvider::class,
```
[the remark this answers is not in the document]
MyIgel: Thats also ok ;)

resources/lang/de_DE/default.po
MyIgel: Uff :D

MyIgel:
I'll have a look at the error & releasing a new major release should not be a problem @ congress ;)

MyIgel:
Funny enough that the timestamps can only reach 12.1.2038 03:14:07 UTC (and another reason why the rewrite should be done) :D

MyIgel:
That fixes the bugs above so it should be now mergeable ;)

msquare:
The bug seems still to exist in line 461 (Trying to get property 'id' of non-object).

MyIgel:
Hmmm, should be now fixed it by adding additional braces...
Anyhow, its rebased on the master (had a merge conflict)

msquare:
Nice :)

PR #687, opened 2019-12-01
Add CONTRIBUTING.md

weeman1337:
I've started a *CONTRIBUTING.md* .

I want to make it easier and less frustrating for others to join the development. For me personally it would have been much more enjoyable having a file telling all the things instead of n+1 comments in PRs where changes based on implicit knowledge are requested.

If you want to see other things there I would be happy to add them @MyIgel @msquare

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,10 @@
+# Contributing
+
+## Code style
+
+* Make sure your code follows the [PSR-12](https://www.php-fig.org/psr/psr-12/) code style.
+You may use `composer run phpcs` to verify that.
+* Use `use` statements wherever possible instead of writing the fully qualified name.
+* Order the composer/npm dependencies alphabetically.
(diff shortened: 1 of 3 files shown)


msquare:
Yeah. @MyIgel and I discussed this thought a few days ago. I would also like to see both of our development chapters (Yeah!) from the readme in the contributing file.

weeman1337:
> I would also like to see both of our development chapters (Yeah!) from the readme in the contributing file.

It's only my opinion but in the CONTRIBUTING file I would only expect the guide lines on how to contribute. The development chapters are more technical parts. If the README becomes too cluttered what about putting that into a DEVELOPMENT.md?

MyIgel:
I also would see that in a development file than in the contributions file.
Maybe we should create (at least the development file) together sometime?

rugk:
I would put the dev part from the Readme into the contributing part, too.

This PR would also fix #687

Issue #714, opened 2019-12-31
Needs favicon


rugk:
The website has no favicon… :cry: 

Maybe you can just use the same icon you always use for angels?
Or you just use some variant/image of the :angel: emoji.

I could pick this up, if you tell me which favicon/image to choose. (preferably SVG)

MyIgel:
The angel icon is available as `resources/assets/icons/angel.svg`, feel free to use that :)

MyIgel:
Implemented in b842466b3a842652662a0b354752bf5973c2b0cc

PR #717, opened 2020-01-03
Add favicon & Web Manifest etc.

rugk:
Fixes https://github.com/engelsystem/engelsystem/issues/714

generated by https://realfavicongenerator.net/
with these settings: https://gist.github.com/rugk/1a241f2fbc4c0c5d493119b688a0917d

It ensures proper look for all OSes, allows to add the website as a "simple" PWA via the [web manifest](https://developer.mozilla.org/docs/Web/Manifest) etc.

See also the settings I selected and their preview here (I selected general settings, so they are not influenced by any congress design, i.e. at 37c3 you could/may e.g. change the theme color or so):

grafik

**Important:** I did try to setup this locally for testing, but did not know how to proceed after `composer install`. As I think this is a simple change not involving code logic, I hope your review is enough to catch issues (maybe wrong icon path? I just assumed it works like this.) here…
Theoretically™ it should™ work.


MyIgel:
Hmm, I don't think that its a great idea to save multiple binary blobs in the engelsystem just to have different resolutions of the icon. They should be generated while running `yarn build` to save space and make development easier (for example changing the icon for an event means having multiple files that should not be versioned...).
Also some of the manifest files have dynamic content (like the string "Engelsystem") and eventually also the color and thus should imho be generated on the fly?

@marudor do you have an opinion on that?

rugk:
Okay, I don't have any experience with this generation system, but yes, theoretically an integration would be great… :man_shrugging:

marudor:
https://github.com/jantimon/favicons-webpack-plugin Could be included.

MyIgel:
Hi @rugk, could you have a look at it and try to implement the generator?

MyIgel:
Just as a side note: The favicon part has already been added by using the SVG in b842466b3a842652662a0b354752bf5973c2b0cc -> can SVGs be used here too eventually? That would fix the need for further static files (Which eventually also could be generated using php from one source file and then cached)

marudor:
most (if not all) platforms that use the webmanifest will use the favicon as fallback. So having a svg favicon will be enough for most platforms.
I doubt we need a specific webmanifest for the engelsystem.

MyIgel:
Thats right so i'll close this one & it can be recreated if there is a ned for it

Issue #723, opened 2020-03-26
Inconsistent date formatting with german language settings


chriskbach:
Language settings unfortunately do not affect the format of dates in all views.

Event dates are shown as  DD.MM.YYYY in the footer and on the login page. The dates of news, or in the shift view are all shown as YYYY-MM-DD. For "normal" people this can be confusing.

I found no settings in the config.php or in the language file. Is this something that could be changed easily somewhere in PHP just for the view?

I see dates being handled in many places in the scripts. Unfortunately my PHP knowledge is very limited and without an overview of the project, I got lost immediately. It would be highly appreciated, if someone with enough overview could tell me if this is possible with a small change in the code or if this requires change too many functions in the code.

chriskbach

msquare:
Fixing the news page is easy. Fixing the shifts filter is a bit more complicated.

MyIgel:
This should be "just" searching for the `'Y-m-d'` (and similar) formatting options through the code base and wrapping them in `__([...])` to get translated (in in the best case verifying that a translation exists)

msquare:
Nope, this will break the datepickers.

MyIgel:
Thats right and needs to be configured in the js lib but the other occurrences could be fixed that way

chriskbach:
Translating the ticks in the shifts view table required me to remove the linebreak, see here:
chriskbach/engelsystem@bdca96c3ae9b94254ff3971e11b05a2feefc926d

for me this works, somehow the linebreak is still present in the view. If you have any hints for me how to do this properly, I can try and then create a pull-request here.

msquare:
I guess the divs are to small to display d-m H:I in one row, so the browser breaks the line.
What about  __('d-m') . '<br>' . __('H:i')?

chriskbach:
thank you, I made the changes and created a pull request, works as proposed by you, except a space was needed for the line break to work

MyIgel:
I guess this issue can be closed now?

johnjohndoe:
:thinking: Just to double check: Is it correct that the date format in the JSON export stays the same?

msquare:
Yes.

PR #731, opened 2020-04-10
Chriskbach germandates

chriskbach:
this is a fix for the shift view to show correct german dates in the first column of the table, see #723

### DIFF
--- a/includes/view/ShiftCalendarRenderer.php
+++ b/includes/view/ShiftCalendarRenderer.php
@@ -222,14 +222,14 @@ class ShiftCalendarRenderer
                 return div('tick day');
             }
             return div('tick day', [
-                date('m-d<b\r />H:i', $time)
+                date(__('m-d'), $time) .'<br>'.date(__('H:i'), $time)
             ]);
         } elseif ($time % (60 * 60) == 0) {
             if (!$label) {
                 return div('tick hour');
             }
             return div('tick hour', [
-                date('m-d<b\r />H:i', $time)
+                date(__('m-d'), $time) .'<br>'.date(__('H:i'), $time)
             ]);
         }
         return div('tick');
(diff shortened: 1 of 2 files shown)


includes/view/ShiftCalendarRenderer.php
```diff
@@ -222,14 +222,14 @@ private function renderTick($time, $label = false)
                 return div('tick day');
             }
             return div('tick day', [
-                date('m-d<b\r />H:i', $time)
+                date(__('m-d').'<br>'.__(' H:i'), $time)
```
MyIgel, line 225: There was a good reason for using `\r` here as the `r` gets interpreted by the date formatter.
Also instead of using `' H:i'` as the translation string the space before it should be added after the break to make it better translatable (and circumvents double translations, ones with a space and once without it).

The clearer way would be to use `date(__('m-d'), $time) .'<br>'.date(__('H:i'), $time)` here (as  the whitespace can be omitted)

includes/view/ShiftCalendarRenderer.php
```diff
@@ -222,14 +222,14 @@ private function renderTick($time, $label = false)
...
             ]);
         } elseif ($time % (60 * 60) == 0) {
             if (!$label) {
                 return div('tick hour');
             }
             return div('tick hour', [
-                date('m-d<b\r />H:i', $time)
+                date(__('m-d').'<br>'.__(' H:i'), $time)
```
MyIgel, line 232: Same here

MyIgel:
Fixed the merge problems for you ;)

Looks like you have merged the changes onto the squash commit, using `git push -f` would had helped here (instead of merging).

What i did was `git push chriskabach a668b8ec:chriskbach-germandates -f`

PR #736, opened 2020-04-25
Clarified license definition as GPL 2.0 or later is supported

MyIgel:
This defines that GPL-2.0 or later (and thus 3.0) applys


weeman1337:
Why not licensing the Engelsystem under AGPLv3 or later?  
That also covers providing the software as a service.

msquare:
Thanks for asking, yes, I know. At first we want to clarify the actual license.

PR #749, opened 2020-07-31
refactor(lang): Put all english translations into the language file

haslersn:
This makes it easier to make Engelsystem forks which solely change some language strings.


MyIgel:
Hi,
this seems to be a valid approach but is definitely something that should be avoided as it adds just another file that needs to be changed when editing translations in source code.
To fix that we are currently transforming the code to use more "abstract" representations of string but until then its just more useless files to carry around and take care of which I would definitely avoid.

A more workable way is the one I implemented for an event a while ago (https://github.com/MyIgel/engelsystem/commit/4c2a0bef5a5ac8bd9fa88aa09e6fe983b42b25f5) which adds a wrapper to the translation class and allows to transparently replace text before it gets displayed. That should imho be enough for most use cases were only some words need to be replaced.

I cherrypicked the commit on the current master on https://github.com/MyIgel/engelsystem/tree/translation-mapping so you could have a look there.

If you have more questions feel free to ask :)

haslersn:
> Hi,
> this seems to be a valid approach but is definitely something that should be avoided as it adds just another file that needs to be changed when editing translations in source code.

You misunderstood. After this PR, you don't edit translations in the source code. You only edit the `msgstr`.

> To fix that we are currently transforming the code to use more "abstract" representations of string

That can still be done. It simply means changing the `msgid` to something more abstract.

MyIgel:
> You misunderstood. After this PR, you don't edit translations in the source code. You only edit the `msgstr`. For every language except for english, this is already the case.

If I only change them in the translation it would make the whole old engelsystem even worse to maintain as the texts will not be the ones the user sees at the end which is a nightmare in debugging and just finding things in the legacy code. Been there, tested that, won't recommend.

> That can still be done. It simply means changing the `msgid` to something more abstract.

And that's on the roadmap but its work which can be better investigated into rebuilding components imho so they'll go if they get replaced

msquare:
Apologies, since this pr introduces more confusion between texts in codebase and english language file, we decided that we won't merge it.
Thank you for your contribution.

PR #752, opened 2020-09-01
Add option for minutes after shift start where signup is allowed

neumantm:
For example it might be reasonable to allow signup up to 5 minutes after the shift has started.


MyIgel:
Thank you for your PR! This could be a great feature sometimes and eventually reduce the work for heaven on smaller events ;)

PR #754, opened 2020-09-13
Smaller improvements

MyIgel:
* Show "Mark as arrived" button on profile page
* Add "Edit" and "Delete" buttons on room pages
* User: Fixed default arrival/departure dates
* User: Allow group admins to edit their own groups
* Added "Welcome Angel" permissions group

### DIFF
--- a/includes/view/Rooms_view.php
+++ b/includes/view/Rooms_view.php
@@ -53,7 +53,19 @@ function Room_view($room, ShiftsFilterRenderer $shiftsFilterRenderer, ShiftCalen
     return page_with_title(glyph('map-marker') . $room['Name'], [
         $assignNotice,
         $description,
-        tabs($tabs, $selected_tab)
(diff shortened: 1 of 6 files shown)


includes/view/Rooms_view.php
```diff
@@ -53,7 +53,19 @@ function Room_view($room, ShiftsFilterRenderer $shiftsFilterRenderer, ShiftCalen
     return page_with_title(glyph('map-marker') . $room['Name'], [
         $assignNotice,
         $description,
-        tabs($tabs, $selected_tab)
+        auth()->can('admin_rooms') ? buttons([
+            button(
+                page_link_to('admin_rooms', ['show' => 'edit', 'id' => $room->id]),
```
msquare, line 58: $room->id is still in array here, but that is no problem when merging the room model rewrite.

msquare (approved):

PR #755, opened 2020-09-13
Rooms Model

MyIgel:
Added a `Room` model and thus upgraded the database structure and cleaned up migrations


msquare (changes requested):
Nice :)
There is a conflict to resolve to merge this pr.

msquare (approved):

PR #760, opened 2020-10-11
Add more config options via ENV & Docker

christian-heusel:
Since I set up a version of the engelsystem via docker (:whale:) just now, I noticed that some of the config vars support being set via env and some dont.
I expanded the amount of variables that can be set and moved them into an extra env file.

I am very happy to discuss if people think a different approach would be more suitable! :blush:


MyIgel:
Hi, thank you for the PR, looks good so far. I'm not sure about how useful the `.env` file is as it effectively doubles the places where the options have to be changed to not override them, is there another benefit of having it? (besides changing an env file instead the compose file).
Also the dockerignore part seems to be not that beneficiary?

christian-heusel:
> I'm not sure about how useful the .env file is as it effectively doubles the places where the options have to be changed to not override them, is there another benefit of having it?

Yeah I also noticed that ... I will probably just move the env vars back into the `docker-compose.yml` .. Its just that I am currently stuck having some options set via env vars and some as a config change and I think only one of them at a time would be better / more consistent.

I mean that is probably a design decision, but we could also remove some of the currently existing env options and remove this line from the `Dockerfile` and say that creating a custom config is the way to go:
https://github.com/engelsystem/engelsystem/blob/dbf373717063ba356226261dc55870227b58de7e/docker/Dockerfile#L29

I would like more env vars, thats why I wrote the PR this way :D

MyIgel:
> Yeah I also noticed that ... I will probably just move the env vars back into the `docker-compose.yml` .. Its just that I am currently stuck having some options set via env vars and some as a config change and I think only one of them at a time would be better / more consistent.

Imho the only config in the compose file should be that one which needs to be changed to run there (so only mysql).

> I mean that is probably a design decision, but we could also remove some of the currently existing env options and remove this line from the `Dockerfile` and say that creating a custom config is the way to go

Baking config files into the container never was and never will be a good idea and should be avoided so the line should remain (and config files linked to containers on start).

> I would like more env vars, thats why I wrote the PR this way :D

Thats definitely a good way, i agree -> more env vars are better!

The plan is that at some time in the future most of them can be changed using the webinterface and the other more or less dynamically replaced by env vars but thats another issue ^^

MyIgel:
I should had used added *requests changes ;)

PR #774, opened 2020-11-21
Schedule import: Add overview

MyIgel:
Created an overview page from where multiple schedules can be imported

### DIFF
--- a/resources/views/admin/schedule/index.twig
+++ b/resources/views/admin/schedule/index.twig
@@ -6,35 +6,56 @@
 
 {% block content %}
     <div class="container">
-        <h1>{% block content_title %}{{ title }}{% endblock %}</h1>
+        <h1>
+            {% block content_title %}{{ title }}{% endblock %}
 
-        {% for message in errors|default([]) %}
-            {{ m.alert(__(message), 'danger') }}
-        {% endfor %}
-        {% for message in success|default([]) %}
-            {{ m.alert(__(message), 'success') }}
-        {% endfor %}
+            {% if is_index|default(false) %}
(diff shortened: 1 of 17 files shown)


resources/views/admin/schedule/index.twig
```diff
@@ -6,35 +6,56 @@
...
+        <h1>
+            {% block content_title %}{{ title }}{% endblock %}
 
-        {% for message in errors|default([]) %}
-            {{ m.alert(__(message), 'danger') }}
-        {% endfor %}
-        {% for message in success|default([]) %}
-            {{ m.alert(__(message), 'success') }}
-        {% endfor %}
+            {% if schedules|default([]) %}
```
msquare, line 12: This shows the add button only if there is at least one schedule configured. The button should always be visible.

resources/views/admin/schedule/index.twig
```diff
@@ -6,35 +6,56 @@
...
+                            <tbody>
+                            {% for schedule in schedules %}
+                                <tr>
+                                    <td>{{ schedule.name }}</td>
+                                    <td>{{ schedule.url }}</td>
+                                    <td>
+                                        <div class="btn-group">
+                                            <a
+                                                href="{{ url('/admin/schedule/load/' ~ schedule.id) }}"
+                                                class="btn btn-default">
```
msquare, line 43: Please add btn-xs

msquare (changes requested):
Works out of the box as well. Very nice!
Only some minor corrections needed. Thank you.

resources/views/admin/schedule/index.twig
MyIgel: Whops, äh, will fix that :D

msquare (approved):

PR #776, opened 2020-11-24
Settings

msquare:
* Applied new settings layout
* Added settings section to change password according to new layout

### DIFF
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -7,6 +7,7 @@ use Engelsystem\Models\Worklog;
 use Engelsystem\Renderer\Renderer;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
+use Engelsystem\Controllers\SettingsController;
 
 /**
  * Renders user settings page
@@ -44,6 +45,7 @@ function User_settings_view(
         'pages/settings/settings.twig',
         [
             'title' => 'settings.profile',
+            'settings_menu' => app()->make(SettingsController::class)->settingsMenu(),
             'content' =>
                 msg()
                 . div('row', [
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -108,13 +108,6 @@ function User_settings_view(
                     : ''
                 ),
                 form([
-                    form_info(__('Here you can change your password.')),
-                    form_password('password', __('Old password:')),
-                    form_password('new_password', __('New password:')),
-                    form_password('new_password2', __('Password confirmation:')),
-                    form_submit('submit_password', __('Save'))
-                ]),
-                form([
                     form_info(__('Here you can choose your color settings:')),
                     form_select('theme', __('Color settings:'), $themes, $user_source->settings->theme),
                     form_submit('submit_theme', __('Save'))
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -37,23 +37,22 @@ function User_settings_view(
     $enable_dect = config('enable_dect');
     $enable_planned_arrival = config('enable_planned_arrival');
 
-    $showOauth = false;
-    foreach ($oauth2_providers as $name => $config) {
-        $showOauth = $showOauth || !isset($config['hidden']) || !$config['hidden'];
-    }
-
     return page_with_title(settings_title(), [
         msg(),
         div('row', [
-            div('col-md-6', [
+            div('col-md-3 settings-menu', [
+                '<ul class="nav nav-pills nav-stacked">',
+                '<li class="active"><a href="' . url('/user-settings') . '">' . __('settings.profile') . '</a></li>',
+                '<li><a href="' . url('/settings/password') . '">' . __('settings.password') . '</a></li>',
+                '<li><a href="' . url('/settings/oauth') . '">' . __('settings.oauth') . '</a></li>',
+                '</ul>'
+            ]),
+            div('col-md-9', [
+                heading(__('settings.profile'), 2),
                 form([
                     form_info('', __('Here you can change your user details.')),
                     form_info(entry_required() . ' = ' . __('Entry required!')),
                     form_text('nick', __('Nick'), $user_source->name, true),
-                    form_info(
-                        '',
-                        __('Use up to 23 letters, numbers, connecting punctuations or spaces for your nickname.')
-                    ),
                     $enable_pronoun
                         ? form_text('pronoun', __('Pronoun'), $personalData->pronoun, false, 15)
                         . form_info('', __('Will be shown on your profile page and in angel lists.'))
@@ -99,14 +98,7 @@ function User_settings_view(
                     ) : '',
                     form_info('', __('Please visit the angeltypes page to manage your angeltypes.')),
                     form_submit('submit', __('Save'))
-                ])
-            ]),
-            div('col-md-6', [
-                ($showOauth ?
-                    form_info(__('oauth.login'))
-                    . button(url('/settings/oauth'), __('settings.oauth'), 'btn-primary')
-                    : ''
-                ),
+                ]),
                 form([
                     form_info(__('Here you can choose your color settings:')),
                     form_select('theme', __('Color settings:'), $themes, $user_source->settings->theme),
@@ -119,7 +111,7 @@ function User_settings_view(
                 ]),
             ])
         ])
-    ]);
+    ], true);
 }
 
 /**
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -4,6 +4,7 @@ use Carbon\Carbon;
 use Engelsystem\Models\Room;
 use Engelsystem\Models\User\User;
 use Engelsystem\Models\Worklog;
+use Engelsystem\Renderer\Renderer;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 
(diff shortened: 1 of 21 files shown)


resources/lang/en_US/default.po
```diff
@@ -181,8 +181,17 @@ msgstr "Message"
...
 
+msgid "settings.profile"
+msgstr "Profile"
+
+msgid "settings.password"
+msgstr "Password"
+
 msgid "settings.oauth"
-msgstr "OAuth Settings"
+msgstr "Single-Sign-On"
```
MyIgel, line 191: Same here

resources/views/macros/form.twig
```diff
@@ -77,3 +77,8 @@
 {% macro submit(label, opt) %}
     {{ _self.button(label|default(__('form.submit')), opt|default({})|merge({'type': 'submit'})) }}
 {%- endmacro %}
+
+{% macro info(text) %}
+{% import 'macros/base.twig' as m %}
```
MyIgel, line 82: Hmm, imho it should be imported at the top? And i would also move it to the base macros as its not that form specific imho

resources/views/pages/news/edit.twig
```diff
@@ -62,7 +62,7 @@
             {% if news %}
                 <div class="row">
                     <div class="col-md-12">
-                        <h2>Preview</h2>
+                        <h2>{{ __('Preview') }}</h2>
```
MyIgel, line 65: Oof, that should be a `news.preview` :D

includes/view/User_view.php
```diff
@@ -37,23 +37,22 @@ function User_settings_view(
...
-    foreach ($oauth2_providers as $name => $config) {
-        $showOauth = $showOauth || !isset($config['hidden']) || !$config['hidden'];
-    }
-
     return page_with_title(settings_title(), [
         msg(),
         div('row', [
-            div('col-md-6', [
+            div('col-md-3 settings-menu', [
+                '<ul class="nav nav-pills nav-stacked">',
```
MyIgel, line 44: That means that this navigation has to be changed in two places when editing it which is not that great, there has to be a better option using the twig template and adding the content directly to it

resources/views/pages/settings/password.twig
```diff
@@ -0,0 +1,36 @@
...
+
+{% block title %}{{ __('settings.password') }}{% endblock %}
+
+{% block row_content %}
+        <form action="" enctype="multipart/form-data" method="post">
+            {{ csrf() }}
+
+            <div class="row">
+                <div class="col-md-12">
+                    {{ f.info(__('Here you can change your password.')) }}
```
MyIgel, line 13: These messages should be replaced by for example `settings.change-password`, `settings.old-password` etc

src/Controllers/SettingsController.php
```diff
@@ -21,11 +34,54 @@ class SettingsController extends BaseController
...
+     */
+    public function savePassword(Request $request): Response
+    {
+        $user = $this->auth->user();
+
+        if (
+            !$request->has('password')
+            || !$this->auth->verifyPassword($user, $request->postData('password'))
+        ) {
+            $this->addNotification('-> not OK. Please try again.', 'errors');
```
MyIgel, line 72: These should definitely be translation strings (and esp. this message should be reworded as its not that descriptive)

src/Controllers/SettingsController.php
```diff
@@ -21,11 +34,54 @@ class SettingsController extends BaseController
...
     public function __construct(
+        Authenticator $auth,
         Config $config,
+        LoggerInterface $log,
+        Redirector $redirector,
         Response $response
     ) {
-        $this->config = $config;
-        $this->response = $response;
+            $this->auth = $auth;
```
MyIgel, line 43: Interesting formatting which looks odd when comparing it tot he rest of the file

src/Controllers/SettingsController.php
```diff
@@ -21,11 +34,54 @@ class SettingsController extends BaseController
...
+        return $this->response->withView(
+            'pages/settings/password.twig',
+            $this->getNotifications()
+        );
+    }
+
+    /**
+     * @return Response
+     */
+    public function savePassword(Request $request): Response
```
MyIgel, line 64: I'm missing phpdoc tags for the Parameter ;)

src/Controllers/SettingsController.php
```diff
@@ -21,11 +34,54 @@ class SettingsController extends BaseController
...
+    {
+        $user = $this->auth->user();
+
+        if (
+            !$request->has('password')
+            || !$this->auth->verifyPassword($user, $request->postData('password'))
+        ) {
+            $this->addNotification('-> not OK. Please try again.', 'errors');
+        } elseif (strlen($request->postData('new_password')) < config('min_password_length')) {
+            $this->addNotification('Your password is to short (please use at least 6 characters).', 'errors');
```
MyIgel, line 74: The number should be replaced with a `%d` here to show the real configured value

tests/Unit/Controllers/SettingsControllerTest.php
```diff
@@ -10,34 +10,237 @@
...
+    protected $user;
+
+    /**
+     * @covers \Engelsystem\Controllers\SettingsController::password
+     */
+    public function testPassword()
+    {
+        /** @var Response|MockObject $response */
+        $this->response->expects($this->once())
+        ->method('withView')
```
MyIgel, line 52: These should be indented by 4 spaces to be better readable, will have a look at how the linter can find those ;)

tests/Unit/Controllers/SettingsControllerTest.php
```diff
@@ -10,34 +10,237 @@
...
+    public function testSavePassword()
+    {
+        $body = [
+            'password' => 'password',
+            'new_password' => 'newpassword',
+            'new_password2' => 'newpassword'
+        ];
+        $this->request = $this->request->withParsedBody($body);
+
+        $this->auth->expects($this->once())
```
MyIgel, line 76: :information_source: There is a `$this->setExpects(...)` method which shrinks it to one line

src/Controllers/SettingsController.php
```diff
@@ -21,11 +34,54 @@ class SettingsController extends BaseController
...
+    }
+
+    /**
+     * @return Response
+     */
+    public function savePassword(Request $request): Response
+    {
+        $user = $this->auth->user();
+
+        if (
```
MyIgel, line 68: I would recommend using the internal validator here

```php
$minLength = config('min_password_length');
$data = $this->validate($request, [
    'password' => 'required',
    'new_password' => 'required|min:' . $minLength,
    'new_password2' => 'required'
]);
```

Eventually the is equal to can be validated, that has to be checked. I would definitely rename the field to password_confirmation ^^

resources/views/pages/settings/password.twig
```diff
@@ -0,0 +1,36 @@
...
+                        'password',
+                        __('Old password:'),
+                        'password',
+                        {'required': true}
+                    ) }}
+                    {{ f.input(
+                        'new_password',
+                        __('New password:'),
+                        'password',
+                        {'required': true}
```
MyIgel, line 24: I would definitely also add a html `minlength` attribute here (as `min` opt) which gets configured by the config which should mean that one sentence has to be translated less :upside_down_face:

MyIgel (changes requested):

resources/views/macros/form.twig
msquare: That didn't work. Any other ideas?

MyIgel: 0o

Thats interesting (and looks more like a bug than a feature..)

But would be irrelevant if moved to the `base.twig` file

msquare: According to https://stackoverflow.com/questions/9548601/calling-a-macro-inside-another-macro-in-twig, the import was actually the correct way. It's even needed when moved to `base.twig`.

src/Controllers/SettingsController.php
msquare: This is now done by Validator and browser input length.

resources/views/macros/form.twig
MyIgel: When referencing macros from the same file the `_self` keyword can be used ;)

resources/views/pages/settings/password.twig
msquare: Added minlength to registration and password reset as well.

includes/view/User_view.php
msquare: Sorry, won't happen until settings are completely rewritten. Tried 5 things that didn't work.

MyIgel: It works for me, see 35aff139f7c0d140cf1751b104fa2e5320ea6dba ;)

resources/lang/de_DE/additional.po
```diff
@@ -27,7 +27,10 @@ msgid "validation.email.email"
 msgstr "Die E-Mail-Adresse ist nicht gültig."
 
 msgid "validation.password.min"
-msgstr "Dein angegebenes Passwort ist zu kurz."
+msgstr "Dein neues Passwort ist zu kurz."
```
MyIgel, line 30: That also affects the registration pssword but as it should be never used (the form has the minlnength requirement) the wording is not that important here :shrug:

resources/lang/de_DE/default.po
```diff
@@ -238,11 +238,6 @@ msgid "We sent you an email containing your password recovery link."
 msgstr ""
 "Wir haben dir eine eMail mit einem Link zum Passwort-zurücksetzen geschickt."
 
-#: resources/views/pages/password/reset-success.twig:9
-#: includes/pages/user_settings.php:120
-msgid "Password saved."
```
MyIgel, line 243: This string is used in `resources/views/pages/password/reset-success.twig`

src/Controllers/SettingsController.php
```diff
@@ -39,10 +107,27 @@ public function oauth(): Response
...
         );
     }
+
+    /**
+     * @return array
+     */
+    public function settingsMenu(): array
+    {
+        $menu = [
+            url('/user-settings') => 'settings.profile',
```
MyIgel, line 124: This should be moved to `additional.po` as it has no direct translation reference

MyIgel (changes requested):

MyIgel (approved):

PR #782, opened 2020-12-18
Questions, FAQ and settings fixes

MyIgel:
* Rebuild the Questions frontend
* Added a FAQ feature
* Use maxlength on user settings to prevent 500 errors


msquare (changes requested):
Nice work, as always!
Since the faq is now included in the footer per default config and therefore also visible for visitors, the migration script should add the view faq permission to the guest role as well.

And all "icon only" buttons should get a title, but not in this pr i guess.

msquare (approved):

PR #798, opened 2021-03-30
PHP 8.0 compatibility

herzmut:
Hello there,

I am not sure how far the project has advanced to establish PHP 8.0 compatibility before my PR. I managed to execute PHPUnit tests (with the astonishing original coverage of 100% 😲) on PHP 8 - the project is PHP 8-compatible 🎉, and I updated the Dockerfiles to reflect that. 

The unit tests had a problem with a recent migration script, as where current versions of IlluminateDb seem to require "nullable" to be explicitly set on columns if they are not created with a default value, as the assumed default value for all columns seems to be... null? And yet they're not nullable by default? ... Who in Laravel's team had that idea? Nevermind.

I think this problem might not have occured if `composer.lock` were included in the project, so everyone could share the same dependency fingerprints as [Conposer's Basic Usage Guide suggests](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control) - you can switch tags, you can rename branches (as that happened for the main branch between my last fork and now), but you can't temper (easily) on commit hashes, that's why the Composer people recommend `composer.lock` to be included in the repo. However, I kept `.gitignore` as it is, because I believe that's not my decision to make. 😬

To make everything even more weird: the migrate script of your project did not seem to have any problem with missing default values when executed on the container and not in the context of PHPUnit. Maybe setting the columns nullable is the better idea than setting unexpressive default values like I did?

Thank you for considering this PR,
~ Herzmut


MyIgel:
Thanks for the PR, i'll have a look :D
The container uses a mariadb server while the unit tests run mostly on a sqlite database (as its in memory engine is way more faster than a dedicated sql server), so that seems to be a engine dependent problem which is interesting

MyIgel:
(and also this didn't happen when i added the compatibility in https://github.com/engelsystem/engelsystem/pull/791) -> i'll have a look ^^)

MyIgel:
I tested the current engelsystem on a clean installation an was not able to trigger the nullable error, where did you get it?

With PHP 8 already active this PR only changes the minimum required PHP version to 8 while 7.4 still gets maintained, that's a change that i would like to make earliest at the end of the year if i want to explicitly use php 8 features. @msquare whats your opinion on that?

PS: The composer.lock is now also versioned ;)

msquare:
Please no PHP 8 requirement before ubuntu lts has it. (It's on 7.4 currently.)

Would you like to test the behavior again with the actual main codebase, @herzmut?

herzmut:
@MyIgel: 
> I tested the current engelsystem on a clean installation an was not able to trigger the nullable error, where did you get it?

I really got it by executing the PHPUnit tests on the Docker environment.

> PS: The composer.lock is now also versioned ;)

Excellent! :D 

@msquare: 
> Please no PHP 8 requirement before ubuntu lts has it. (It's on 7.4 currently.)

Is the non-virtualized (native) execution of engelsystem the preferred way? Otherwise we could maintain major version branches for PHP 7.4 and PHP 8 each at the same time, the latter for Docker with IP outbound until Ubuntu LTS gets PHP 8.

> Would you like to test the behavior again with the actual main codebase, @herzmut?

Of course! Please stand by...

MyIgel:
So that's great news, (even if it would be really interesting to know how to trigger that interesting error ^^)

Having to maintain an additional branch is imho not the best option, esp. as there are so few maintainers to do so ;)

Basically changing the required version is a consideration between "Which versions get official support", "What can be installed on the Engelsystem Hosted Servers" and "I really want to use that new feature that got introduced".

The primary option should be definitively running them on freshly build docker containers with the latest and greatest PHP version to benefit from performance improvements. But as long as i haven't completely convinced msquare and lea that they should use a docker setup (and to also allow more seamless "classical" hosting), above points have to be thought about.

herzmut:
> So that's great news, (even if it would be really interesting to know how to trigger that interesting error ^^)

I started Docker via docker-compose in ./docker/dev and created a custom config to use the Docker database. As PhpUnit executes migrations on its own, it failed on the first test until I defined default values for the columns.

> Would you like to test the behavior again with the actual main codebase, @herzmut?

I tested against the current code base and there were no errors occuring. However, the migration script `2020_11_20_000000_add_name_minutes_and_timestamps_to_schedules.php` is the only migration script _not_ containing default values or nullable marks for the columns.

If PHP 8 as a requirement is too early for the Engelsystem user base for now, we should postpone it.

msquare:
@herzmut Thank you for your feedback!

> > Please no PHP 8 requirement before ubuntu lts has it. (It's on 7.4 currently.) 
> Is the non-virtualized (native) execution of engelsystem the preferred way? Otherwise we could maintain major version branches for PHP 7.4 and PHP 8 each at the same time, the latter for Docker with IP outbound until Ubuntu LTS gets PHP 8.
At least for the chaos events, we deploy the engelsystem natively. There are also a lot of people hosting small instances on web hosters which are usuallly a bit behind with PHP upgrades.

Which of the PR topics are open?
- [ ] PHP 8 requirement
- [ ] PHP 8 branch
- [ ] Include composer.lock
- [ ] PHPUnit problems using docker

What would be the benefits from maintaining a PHP 8 branch?

herzmut:
> What would be the benefits from maintaining a PHP 8 branch?

PHP is met with suspicion among Chaos admins, especially when they have to rely on Debian / Ubuntu kind of timelines when it comes to updates. Part of how to encounter that, and to win the edge to an inevitable point (the discontinuation of PHP 7.x), a soft upgrade strategy for Engelsystem and its technology stack, which eventually only evolves on PHP 8, maintained in a separate branch and candidating for the next major version of Engelsystem, would address all these issues.

Unlike to other developers, using new language features and cutting edge technology is - while always being interesting on its own - my least concern for a software which so much people depend on and really no good reason to make this decision for Engelsystem. So if my first reason, pointed out in the paragraph above, is to your opinion not good enough for a reason, I would hardly know any other good reason to do it now, as we also have to consider some disadvantages coming with such a strategy, one pointed out already by @MyIgel: the peoplepower it requires.

MyIgel:
Hmm, I think this PR is pretty outdated as the Engelsystem works fine on PHP 8 so I would suggest closing it?
(and am still working on 8.1 support which requires a new release of `thephpleague/oauth2-client` which still not happened..

I would suggest changing the required version to 8 after the coming rc3 2021 release and using the `ppa:ondrej/php` ppa for the newest php version on ubuntu.

herzmut:
I'd agree, except on relying on ondrej's package. Not every Debian / Ubuntu has the preinstalled requirements to handle launchpads out of the box and one of the advantages of Engelsystem was, as far as I recall, that it won't take much prerequisites to run. I don't see any problem to bump to 7.4; former experience with other software has shown no obstacles on a path to upgrade between 7.1 and 7.4. In fact I would recommend it, as 7.3 ended its lifecycle recently and should not be used on a hacker congress anymore...

Issue #773, opened 2020-11-15
Message "Shift ends"


xuwhite:
The Message is grammaticly incorrect and should be replaced with "Shift ended" 

Thanks Xu

xuwhite:
![engelsystem](https://user-images.githubusercontent.com/73076134/99195643-29c86800-2787-11eb-8adf-aa9235d166df.png)

MyIgel:
Eventually reloading the counter when it reaches 0 is an option here

PR #799, opened 2021-03-30
Issue 773: Message "Shift ends"

herzmut:
solves issue #773

### DIFF
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -316,8 +316,10 @@ function User_shift_state_render($user)
             . '</span>';
     }
 
+    $hasShiftEndPassed = $nextShift['end'] < time();
+
     return '<span class="text-danger moment-countdown" data-timestamp="' . $nextShift['end'] . '">'
-        . __('Shift ends %c')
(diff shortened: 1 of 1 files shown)


MyIgel:
Does it? -> Needs further testing as i'm unshure :D

herzmut:
Should I write a unit test to cover that case?

MyIgel:
What I mean is that the part that shows the time in the frontend gets updated by js which itself has to get updated, not the backend code.
The backend itself normally shows either "No shifts" or directly the next shift so i wonder how to even get to this line?

MyIgel:
Even more interestingly, my ci run does not require the default values patch? hmm

herzmut:
That's strange. However, if the JavaScript is not in another repository and composed with this one on a build, it does not seem render these time values on its own but gets it as a template fragment from the backend. Is there any state of the application where you could view a shift which ended, like a detail view of done shifts? The only other option would be to leave the tab open over the end of the shift, so the one-time fetched moment.js template keeps on updating the placeholder without changing the sentence. True, this PR does not solve that yet.

msquare:
The change does not solve the problem.

Issue #806, opened 2021-04-14
Allow admins to configure the initial password of the admin user


neumantm:
Especially when using the docker container in semi-automated cloud deployments, it would be helpful if the the initial admin password could be configured (e.g through an environment variable).

PR #809, opened 2021-04-21
Allow to configure initial admin pw

neumantm:
Resolves: #806


MyIgel:
Thats imho the "easier" approach (and integrates better)

neumantm:
Thanks.
Will check it out.

PR #816, opened 2021-06-03
Removed postcss-preset-env for good as its job is already done by autoprefixer, package upgrades, fix CI builds

MyIgel:
This removes the postcss-preset-env postcss plugin as it requires postcss 7 which has security vulnerabilities.

As all work seems to also be done by autoprefixer itself that should be enough.

Also upgraded all other packages to their newest version.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -62,9 +62,9 @@ validate-yarn:
 #
 
 .docker_template: &docker_definition
-  image: docker:18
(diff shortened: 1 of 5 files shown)


MyIgel:
@weeman1337 was there a specific reason to use that old plugin?

weeman1337:
Jep - changing as less as possible because migrating the frontend is already a huge task.

PR #818, opened 2021-06-21
add nextcloud as an oauth provider

kuhball:
This PR adds nextcloud as an oauth provider. Using the generic provider is not possible since the nextcloud oauth implementation has some special touches to it. 

For using the plugin define the provider like this:
```
 'oauth'                   => [
        // the name needs to include 'nextcloud'
        'nextcloud' => [
            'name' => 'Some Cloud',
            'client_id' => 'abc',
            'client_secret' => 'abc',
            // Nextcloud URL
            'url_nextcloud' => 'https://cloud.domain.xyz'
        ],
    ],
```

The information about users provided by nextcloud leaves much to be desired. 

Some parts of this are a bit messy, please feel free to comment and optimize :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,8 @@
         "symfony/mime": "^5.2",
         "symfony/psr-http-message-bridge": "^2.1",
         "twig/twig": "^3.3",
-        "vlucas/phpdotenv": "^5.3"
+        "vlucas/phpdotenv": "^5.3",
+        "bahuma/oauth2-nextcloud": "^1.1"
     },
     "require-dev": {
         "dms/phpunit-arraysubset-asserts": "^0.3.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,8 +4,49 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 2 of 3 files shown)


MyIgel:
Hi and thanks for your first contribution!
The code looks good besides missing some unit tests and that it would be easier to add an option instead of interpreting the provider name ;)
The downside of adding an additional package is that its a less used one and requires special handling for an integration which was not required for a bigger event so far. As far as i can see and tested its just the accessing of nested info values which could be implemented in a more generic way. I implemented an approach that allows to support to access nested info attributes and would propose to close this PR in favor of #852 which can be configured with
```php
     'nextcloud' => [
            'name' => 'Nextcloud',
            'client_id' => '[client id]',
            'client_secret' => '[client secret',
            'url_auth' => 'https://cloud.example.com/apps/oauth2/authorize',
            'url_token' => 'https://cloud.example.com/apps/oauth2/api/v1/token',
            'url_info' => 'https://cloud.example.com/ocs/v2.php/cloud/user?format=json',
            'nested_info' => true,
            'id' => 'ocs.data.id',
            'username' => 'ocs.data.id',
            'email' => 'ocs.data.email',
        ],
```
to achieve the same result in a more maintainable way.

PR #842, opened 2021-10-12
email option updated

### LINKED ISSUE #669: Make e-mail options more clear on registration
> dieses Jahr hatten wir relativ viele Engel mit genug Stunden für nen
> Voucher, zu denen aber keine Mailadresse vorliegt (vmtl zu großen
> Teilen, weil sie nicht den Optin-Haken gesetzt haben?).
> 
> Ich erinner mich zwar so, als stünde beim signup ziemlich deutlichd
> abei, wofür der Haken ist, aber anscheinend hams ja viele verpasst:
> Falls du eh noch mal drüber guckst, kannst du noch mal gucken, ob man
> das noch deutlicher machen kann.

xuwhite:
fixes #669 

other option if preferred: Voucher mails (allow orga and heaven angels to contact you by mail)

### DIFF
--- a/db/migrations/2021_12_29_000000_users_settings_add_email_goody.php
+++ b/db/migrations/2021_12_29_000000_users_settings_add_email_goody.php
@@ -1,0 +1,42 @@
+<?php
+
+namespace Engelsystem\Migrations;
+
+use Engelsystem\Database\Migration\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+class UsersSettingsAddEmailGoody extends Migration
+{
+    use Reference;
+
+    /**
+     * Run the migration
+     */
+    public function up()
+    {
+        $this->schema->table(
+            'users_settings',
+            function (Blueprint $table) {
+                $table->boolean('email_goody')->default(false)->after('email_human');
+            }
+        );
+
+        $connection = $this->schema->getConnection();
+        $connection
+            ->table('users_settings')
+            ->update(['email_goody' => $connection->raw('email_human')]);
+    }
+
+    /**
+     * Reverse the migration
+     */
+    public function down()
+    {
+        $this->schema->table(
+            'users_settings',
+            function (Blueprint $table) {
+                $table->dropColumn('email_goody');
+            }
+        );
+    }
+}
--- a/includes/pages/guest_login.php
+++ b/includes/pages/guest_login.php
@@ -365,7 +365,7 @@ function guest_register()
                             ),
                             form_checkbox(
                                 'email_by_human_allowed',
-                                __('Allow orga and heaven angels to contact you by mail (for example this is necessary for distributing vouchers)'),
+                                __('To receive vouchers, agree that nick, email address, worked hours and shirt size will be stored until the next similar event. To withdraw your approval, send an email to <a href="mailto:%s">%1$s</a>.', [config('privacy_email')]),
                                 $email_by_human_allowed
                             ),
                         ])
--- a/includes/pages/guest_login.php
+++ b/includes/pages/guest_login.php
@@ -365,7 +365,7 @@ function guest_register()
                             ),
                             form_checkbox(
                                 'email_by_human_allowed',
-                                __('Humans are allowed to send me an email (e.g. for ticket vouchers)'),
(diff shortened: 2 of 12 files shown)


MyIgel:
@msquare do you have any wording opinions on this?

MyIgel:
This option should also be moved to a opt in config as not all events have vouchers

MyIgel (changes requested):
The email definitely should not be hardcoded and thus loaded from the config using `config(...)` by using the second parameter of `__(...)` to replace `%s` placeholders in the translated string.

Additionally the whole config should be made optional as its pretty ccc event specific.

MyIgel:
Shouldn't it be more `I agree` / `stimme Ich zu`?

MyIgel (approved):

MyIgel:
Eventually @weeman1337 could review this one? :)

db/migrations/2021_12_29_000000_users_settings_add_email_goody.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Run the migration
+     */
+    public function up()
+    {
+        $this->schema->table(
+            'users_settings',
+            function (Blueprint $table) {
+                $table->boolean('email_goody')->default(false)->after('email_human');
```
weeman1337, line 20: Migrations work on DB level. Booleans are stored as numbers: `false` should be `0`.
May work by accident because it is going to be casted somewhere :shrug: 

Typed default values should be set in the models
https://laravel.com/docs/8.x/eloquent#default-attribute-values

includes/pages/guest_login.php
```diff
@@ -118,6 +119,10 @@ function guest_register()
             $email_news = true;
         }
 
+        if ($request->has('email_goody')) {
```
weeman1337, line 122: Shouldn't the value be checked here? What if it is sent but the value is `false` or `0`?

resources/lang/de_DE/default.po
```diff
@@ -1617,8 +1617,14 @@ msgid "Notify me of new news"
...
-msgid "Humans are allowed to send me an email (e.g. for ticket vouchers)"
-msgstr "Menschen dürfen mir eine E-Mail senden (z.B. für Ticket Gutscheine)"
+msgid "Allow heaven angels to contact you by e-mail."
+msgstr "Erlaube Himmel-Engeln dich per Mail zu kontaktieren."
+
+msgid "To receive vouchers, agree that nick, email address, worked hours and shirt size will be stored until the next similar event."
+msgstr "Um Voucher zu erhalten, stimme zu, dass Nick, E-Mail-Adresse, geleistete Arbeit und Shirtgröße bis zum nächsten gleichartigen Event gespeichert werden."
+
+msgid "To withdraw your approval, send an email to <a href=\"mailto:%s\">%1$s</a>."
+msgstr "Dies kann jederzeit durch eine Email an <a href=\"mailto:%s\">%1$s</a> widerrufen werden."
```
weeman1337, line 1627:
```suggestion
msgstr "Dies kann jederzeit durch eine E-Mail an <a href=\"mailto:%s\">%1$s</a> widerrufen werden."
```

weeman1337 (changes requested):

db/migrations/2021_12_29_000000_users_settings_add_email_goody.php
MyIgel: It definitely gets converted totally fine and thus is imho more clear here?

And yeah, that should be eventually done in all models at some point (defaults but also explicit casting definitions)

includes/pages/guest_login.php
MyIgel: As it's a checkbox, browsers either send a value or don't send it at all so it's an ok shortcut imho

db/migrations/2021_12_29_000000_users_settings_add_email_goody.php
weeman1337: > And yeah, that should be eventually done in all models at some point (defaults but also explicit casting definitions)

Jep. Because default values only defined per DB will only apply after saving the model.

MyIgel: Its now also explicitly converted in the model :tada:

Issue #695, opened 2019-12-10
Provide event specific time zone via the shifts-json-export API


johnjohndoe:
Not sure if you noticed my comment initially posted on #689. I now move it here so you can handle it as a separate issue.

---

This is a follow-up of adding the `timezone` field to the response of the shifts-json-export API in b540ac93856065ce6e145930b6611e0a0946d646.

I created a couple of test shifts in the staging system both for the Camp 2019 time range (08/2019) and the 36C3 (12/2019). Here are two examples - first a Camp shift at 22.08.2019:

``` json
{
    "Comment": "",
    "Name": "Baikonur",
    "RID": 14,
    "SID": 585,
    "TID": 10,
    "UID": 32,
    "URL": null,
    "created_at_timestamp": 1565979719,
    "created_by_user_id": 32,
    "description": "Baikonur (kasachisc...",
    "edited_at_timestamp": 1567367727,
    "edited_by_user_id": 32,
    "end": 1566435600,
    "freeloaded": 0,
    "id": 42,
    "map_url": "https://openstreetmap.org",
    "name": "Sticker sammeln",
    "shifttype_id": 6,
    "start": 1566434700,
    "timezone": "+01:00",
    "title": "Tag 2: Notebook bekleben bis 03:00"
}
```

Then one for 36C3 at 27.12.2019:

``` json
{
    "Comment": "",
    "Name": "Kourou",
    "RID": 13,
    "SID": 627,
    "TID": 6,
    "UID": 32,
    "URL": null,
    "created_at_timestamp": 1575838846,
    "created_by_user_id": 32,
    "description": null,
    "edited_at_timestamp": 1575838846,
    "edited_by_user_id": null,
    "end": 1577440800,
    "freeloaded": 0,
    "id": 54,
    "map_url": null,
    "name": "Sticker sammeln",
    "shifttype_id": 6,
    "start": 1577437200,
    "timezone": "+01:00",
    "title": "36C3 Sticker sammeln 10-11"
}
```

I noticed that for both shifts the `timezone` is set to `+01:00`. The Camp however was before 27.10.2019 when [Germany changed to winter time](https://www.timechange.info/germany-2019/), `+02:00`. It became apparent because I am currently testing with the Camp 2019 schedule and shifts in the same time range.
To my understand each event needs to "deliver" its original time zone offset aka. the time zone offset which was valid when the event took place. I suppose the `+01:00` is returned for Camp shifts because that is the current time zone offset on the server as of today.
[...]

PR #856, opened 2021-12-09
Just a bunch of fixes

MyIgel:
* Fixed maintenance page assets paths and return 503 status code
* Only show defined contact infos for Anbeltype
* Added output of timezone formatted start/end string to shifts json for #695
* Fixed timezone for iCal exports

### DIFF
--- a/includes/pages/user_ical.php
+++ b/includes/pages/user_ical.php
@@ -1,5 +1,6 @@
 <?php
 
+use Carbon\Carbon;
 use Engelsystem\Http\Exceptions\HttpForbidden;
 
 /**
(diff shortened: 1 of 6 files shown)


includes/pages/user_ical.php
```diff
@@ -61,8 +61,8 @@ function make_ical_entry_from_shift($shift)
     if (isset($shift['Comment'])) {
         $output .= 'DESCRIPTION:' . str_replace("\n", "\\n", $shift['Comment']) . "\r\n";
     }
-    $output .= 'DTSTART;TZID=Europe/Berlin:' . date("Ymd\THis", $shift['start']) . "\r\n";
-    $output .= 'DTEND;TZID=Europe/Berlin:' . date("Ymd\THis", $shift['end']) . "\r\n";
+    $output .= 'DTSTART;TZID=' . config('timezone') . ':' . date("Ymd\THis", $shift['start']) . "\r\n";
+    $output .= 'DTEND;TZID=' . config('timezone') . ':' . date("Ymd\THis", $shift['end']) . "\r\n";
```
msquare, line 65: Please try to convert the time to utc. TZID requires a VTIMEZONE element which simply wont happen.
Also while being at it: Please at DTSTAMP with the same value as the DTSTART. Thanks!

msquare (changes requested):

msquare (approved):

PR #871, opened 2021-12-30
markdown: let Parsedown escape content

tobiaswiese:
Letting Parsedown escape the content, instead of calling
htmlspecialchars provides more context to the escape process.
For example the ampersand character can now be used in markdown links as
part of the url without breaking.

### DIFF
--- a/db/migrations/2021_12_30_000000_remove_admin_news_html_privilege.php
+++ b/db/migrations/2021_12_30_000000_remove_admin_news_html_privilege.php
@@ -1,0 +1,41 @@
+<?php
+
+namespace Engelsystem\Migrations;
+
+use Engelsystem\Database\Migration\Migration;
+
+class RemoveAdminNewsHtmlPrivilege extends Migration
+{
+    /**
+     * Run the migration
+     */
+    public function up()
+    {
+        if (!$this->schema->hasTable('GroupPrivileges')) {
+            return;
+        }
+
+        $connection = $this->schema->getConnection();
+
+        // Delete unused privileges
+        $connection->delete('
+                DELETE FROM `Privileges`
+                WHERE `name` = \'admin_news_html\'
+            ');
+    }
+
+    /**
+     * Reverse the migration
+     */
+    public function down()
+    {
+        if (!$this->schema->hasTable('GroupPrivileges')) {
+            return;
+        }
+
+        $connection->insert('
+                INSERT INTO `Privileges` (`name`, `desc`)
+                VALUES (\'admin_news_html\', \'Use HTML in news\')
+            ');
+    }
+}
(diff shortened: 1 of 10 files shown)


MyIgel (changes requested):
Hi, thank you for this PR!
I have found some changes that should be fixed, as it took me literally longer to test the PR then implementing it i think i will fix them soon ^^

db/migrations/2021_12_30_000000_remove_admin_news_html_privilege.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Reverse the migration
+     */
+    public function down()
+    {
+        if (!$this->schema->hasTable('GroupPrivileges')) {
+            return;
+        }
+
+        $connection->insert('
```
MyIgel, line 36: Which `$connection` do you want to use here?

db/migrations/2021_12_30_000000_remove_admin_news_html_privilege.php
```diff
@@ -0,0 +1,41 @@
+<?php
```
MyIgel, line 1: This should also be added to `tests/Unit/HasDatabase.php` as its not needed to run on unit tests

db/migrations/2021_12_30_000000_remove_admin_news_html_privilege.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        if (!$this->schema->hasTable('GroupPrivileges')) {
+            return;
+        }
+
+        $connection->insert('
+                INSERT INTO `Privileges` (`name`, `desc`)
+                VALUES (\'admin_news_html\', \'Use HTML in news\')
+            ');
+    }
```
MyIgel, line 40: I'm missing the addition of the permission to the `News Admin` ;)

tests/Unit/Controllers/Admin/NewsControllerTest.php
```diff
@@ -23,40 +23,11 @@ class NewsControllerTest extends ControllerTest
...
-            'text'       => '<b>foo</b>',
-            'is_meeting' => false,
+            'text'       => '**foo**',
             'user_id'    => 1,
         ]
     ];
 
-    /**
-     * @covers \Engelsystem\Controllers\Admin\NewsController::edit
-     * @covers \Engelsystem\Controllers\Admin\NewsController::showEdit
```
MyIgel, line 34: That means, that `showEdit` will have no coverage

MyIgel (changes requested):

MyIgel (approved):
Now i'm happy with it ^^

PR #885, opened 2022-03-13
README: Add link to further documentation

cyroxx:
Even though the documentation is still in the build-up phase, it is useful to know that this documentation exists.
So let's link it here already.


MyIgel:
And hope that anyone creates PRs there :zany_face:

PR #903, opened 2022-04-29
Docker: Use single container for server

MyIgel:
Combine the nginx and php-fpm services in a single container to make handling them easier and in preperation for hashed js/css files.


weeman1337:
Tested the builds. They are working. LGTM 👍

PR #908, opened 2022-05-07

frischler:
Most of the code in _includes/pages/user_messages.php_ has been replaced by the new _MessagesController.php_. While doing so, the UI of the messaging page has been improved to match the approach of most other messaging applications, having a conversations overview, and a conversation page.

conversations_overview
conversation

Why not all of the code in _user_messages.php_ got replaced
I couldn't figure out how to get rid of the function _user_unread_messages()_ entirely because of the way how it's used in the _Legacy.php_ (as a TwigFunction). I at least excluded the count logic and placed it in the _MessagesController_, leaving only the creation of the HTML code.


MyIgel:
Hi @frischler, THX for this PR!
As you can see it needs a git rebase on the current main branch (which is way better than a merge ;)), it would be nice if you could do that :)

frischler:
Due to Git history pollution created another pull request: https://github.com/engelsystem/engelsystem/pull/914

PR #914, opened 2022-05-22

frischler:
Most of the code in includes/pages/user_messages.php has been replaced by the new MessagesController.php. While doing so, the UI of the messaging page has been improved to match the approach of most other messaging applications, having a conversations overview, and a conversation page.

conversation
conversations_overview

Why not all of the code in user_messages.php got replaced:
I couldn't figure out how to get rid of the function user_unread_messages() entirely because of the way how it's used in the Legacy.php (as a TwigFunction). I at least excluded the count logic and placed it in the MessagesController, leaving only the creation of the HTML code.


frischler:
Due to Git history pollution created another pull request: https://github.com/engelsystem/engelsystem/pull/915

PR #915, opened 2022-05-22

frischler:
Most of the code in includes/pages/user_messages.php has been replaced by the new MessagesController.php. While doing so, the UI of the messaging page has been improved to match the approach of most other messaging applications, having a conversations overview, and a conversation page.

conversation
conversations_overview

Why not all of the code in user_messages.php got replaced:
I couldn't figure out how to get rid of the function user_unread_messages() entirely because of the way how it's used in the Legacy.php (as a TwigFunction). I at least excluded the count logic and placed it in the MessagesController, leaving only the creation of the HTML code.


src/Controllers/MessagesController.php
```diff
@@ -0,0 +1,285 @@
...
+            ->mapWithKeys(function ($unread) {
+                return [ $unread->user_id => $unread->amount ];
+            });
+    }
+
+    protected function latestMessagePerConversation($current_user): Collection
+    {
+        $latest_message_ids = $this->message
+            ->select($this->raw('max(id) as last_id'))
+            ->where('user_id', "=", $current_user->id)
```
MyIgel, line 266: Eventually using just `'` instead of a mix with `"` if no substitution is required is a great idea

src/Controllers/MessagesController.php
```diff
@@ -0,0 +1,285 @@
...
+            ->orWhere('receiver_id', "=", $current_user->id)
+            ->groupBy($this->raw(
+                '(CASE WHEN user_id = ' . $current_user->id .
```
MyIgel, line 269: Eventually that can be created as a (prepared) statement?

tests/Unit/Controllers/MessagesControllerTest.php
```diff
@@ -0,0 +1,585 @@
...
+
+    /** @var Carbon */
+    protected $now;
+    /** @var Carbon */
+    protected $one_minute_ago;
+    /** @var Carbon */
+    protected $two_minutes_ago;
+
+    /**
+     * @testdox index: underNormalConditions -> returnsCorrectViewAndData
```
MyIgel, line 46: `@covers` annotations are missing to only cover the relevant code for coverage

tests/Unit/Controllers/MessagesControllerTest.php
```diff
@@ -0,0 +1,585 @@
...
+        $this->expectException(ModelNotFoundException::class);
+        $this->controller->delete($this->request);
+    }
+
+    /**
+     * @testdox delete: tryingToDeleteSomeonesMessage -> throwsException
+     */
+    public function testDeleteTryingToDeleteSomeonesMessageThrowsException()
+    {
+        $this->expectException(HttpForbidden::class);
```
MyIgel, line 480: I would add the expected exceptions at the end (before the call) but thats optional

MyIgel (changes requested):
Hi, thx for the rewrite of the messages page, it looks great! <3

All changes are smaller improvements which would be nice if you could add them.

From the messages list page it would be nice to also be able to klick on the angel name to go to the message as its eventually harder to klick on a small text than on a much longer angel name (and also thats what i would expect when clicking on the name there).
On the message page by itself it would be helpful, esp. for shicos, to be able to klick the receiving angels nick to directly go to their profile page.

If you have further questions what i mean please ask me :D

includes/pages/user_messages.php
```diff
@@ -1,174 +1,14 @@
...
-function messages_title()
-{
-    return __('Messages');
-}
+use Engelsystem\Controllers\MessagesController;
 
 /**
  * @return string
  */
 function user_unread_messages()
```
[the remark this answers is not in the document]
MyIgel: Or even better directly moving it to the menu as its pretty trivial

MyIgel (approved):

PR #916, opened 2022-05-23
Database: Increase shift_size column size

xesxen:
This resolves the following error message we're seeing:

```
 Exception: SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'shirt_size' at row 1 (SQL: insert into `users_personal_data` (`first_name`, `last_name`, `pronoun`, `shirt_size`, `planned_arrival_date`, `user_id`) values (, , , 3XL-G, 2022-07-21 00:00:00, 101))
```

Users are running into this on signup, and since the user_personal_data is being saved before their `Angel` role has been assigned, these users are not able to do anything due to lack of rights.

### DIFF
--- a/db/migrations/2022_05_23_000000_increase_tshirt_field_width.php
+++ b/db/migrations/2022_05_23_000000_increase_tshirt_field_width.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Engelsystem\Migrations;
+
+use Engelsystem\Database\Migration\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+/**
+ * To allow for larger key names such as "2XL-G"
+ */
+class IncreaseTshirtFieldWidth extends Migration
+{
+    /**
+     * Run the migration
+     */
+    public function up()
+    {
+        $this->schema->table('users_personal_data', function (Blueprint $table) {
+            $table->string('shirt_size', 10)->change();
+        });
+    }
+
+    /**
+     * Reverse the migration
+     */
+    public function down()
+    {
+        $this->schema->table('users_personal_data', function (Blueprint $table) {
+            $table->string('shirt_size', 4)->change();
+        });
+    }
+}


MyIgel:
Thats a valid workaround, what values are you using for the shirt sizes?

xesxen:
Our current list consists of this: 
```php
[
        'S'    => 'Small Straight-Cut',
        'S-G'  => 'Small Fitted-Cut',
        'M'    => 'Medium Straight-Cut',
        'M-G'  => 'Medium Fitted-Cut',
        'L'    => 'Large Straight-Cut',
        'L-G'  => 'Large Fitted-Cut',
        'XL'   => 'XLarge Straight-Cut',
        'XL-G' => 'XLarge Fitted-Cut',
        '2XL-G' => '2XLarge Fitted-Cut',
        '3XL-G' => '3XLarge Fitted-Cut',
        '2XL'  => '2XLarge Straight-Cut',
        '3XL'  => '3XLarge Straight-Cut',
        '4XL'  => '4XLarge Straight-Cut',
        '5XL'  => '5XLarge Straight-Cut',
]
```

For this, an increase to varchar(5) would suffice, but an increase to 6 would make a bit more sense so someone is able to add an 2-character extension onto the type (eg. 5XL-GR for 5XL in GReen, or even 4XL-SG for "5XL Straight, color Grey)

db/migrations/2022_05_23_000000_increase_tshift_field_width.php
```diff
@@ -0,0 +1,42 @@
...
+ * To allow for larger key names such as "2XL-G"
+ */
+class IncreaseTshiftFieldWidth extends Migration
+{
+    /**
+     * Run the migration
+     */
+    public function up()
+    {
+        if (!$this->schema->hasTable('GroupPrivileges')) {
```
MyIgel, line 17: That check should be irrelevant as only a new table is  changed (and if used the later changed table should be referenced).

db/migrations/2022_05_23_000000_increase_tshift_field_width.php
```diff
@@ -0,0 +1,42 @@
...
+            return;
+        }
+
+        $connection = $this->schema->getConnection();
+
+        $connection->statement(
+            'ALTER TABLE users_personal_data MODIFY shirt_size VARCHAR(6) ;'
```
MyIgel, line 24: To be portable, the creation of the statement should be done by the db abstraction and not manually as that forces it to be only MySQL compatible and breaks for SQLite which is used for tests.
A previous migration is for example [ChangeUsersContactDectFieldSize](https://github.com/engelsystem/engelsystem/blob/main/db/migrations/2018_12_21_000000_change_users_contact_dect_field_size.php#L43) which uses
```php
$table->string($column, $length)->change();
```

Also i think at least 10 would be eventually better here to have more room for others?

db/migrations/2022_05_23_000000_increase_tshift_field_width.php
```diff
@@ -0,0 +1,42 @@
...
+     */
+    public function down()
+    {
+        if (!$this->schema->hasTable('GroupPrivileges')) {
+            return;
+        }
+
+        $connection = $this->schema->getConnection();
+        $connection->statement(
+            'ALTER TABLE users_personal_data MODIFY shirt_size VARCHAR(4) ;'
```
MyIgel, line 39: Same here -> Direct SQL is not a great option

db/migrations/2022_05_23_000000_increase_tshift_field_width.php
```diff
@@ -0,0 +1,42 @@
...
+            'ALTER TABLE users_personal_data MODIFY shirt_size VARCHAR(6) ;'
+        );
+    }
+
+    /**
+     * Reverse the migration
+     */
+    public function down()
+    {
+        if (!$this->schema->hasTable('GroupPrivileges')) {
```
MyIgel, line 33: Should not be needed

MyIgel (changes requested):
I found some changes that should be accommodated for ^^

(and also the commit message should be fixed too ;) )

db/migrations/2022_05_23_000000_increase_tshift_field_width.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Engelsystem\Migrations;
+
+use Engelsystem\Database\Migration\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+/**
+ * To allow for larger key names such as "2XL-G"
+ */
+class IncreaseTshiftFieldWidth extends Migration
```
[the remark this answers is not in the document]
MyIgel: (sorry that i missed that last review)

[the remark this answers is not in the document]
xesxen: Whoops, I wonder how I didn't even spot the typo... fixed...

MyIgel:
Thx for your contribution!

Issue #647, opened 2019-08-24
Shift Groups (backend facing)


ofeldt:
Every shift should have a **ShiftGroupID**. If you create bulk shifts, all shifts belonging to the bulk creation should have the same **ShiftGroupID**.

Such an ID allows us to create the following _quality of life_ improvements in the future:

- Delete shifts en bloc.
- Move shifts by amount of time en bloc.
- Rename shift titles en bloc.

Currently this are all labor and time intensive repetitive tasks, which should be avoided on the user layer.

MyIgel:
* [x] Delete shifts en bloc, implemented in #921
* [ ] Move shifts by amount of time en bloc -> needs a new UI and a definition what "moving" means here
* [ ] Rename shift titles en bloc -> what is the handling of already renamed ones? should they be overwritten? or ignored? how to know what the "original" one was?

PR #921, opened 2022-06-03
Add transaction id to shifts for bulk deletion

MyIgel:
See also https://github.com/engelsystem/engelsystem/issues/647

### DIFF
--- a/includes/model/Shifts_model.php
+++ b/includes/model/Shifts_model.php
@@ -557,12 +557,21 @@ function Shift_update($shift)
 }
 
 /**
+ * Get the next free shifts transaction id
+ */
+function Shift_get_next_transaction_id(): int
+{
+    return Db::selectOne('SELECT MAX(transaction_id) + 1 AS transaction_id FROM Shifts')['transaction_id'] ?? 1;
+}
+
+/**
  * Create a new shift.
  *
  * @param array $shift
+ * @param int   $transactionId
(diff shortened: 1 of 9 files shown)


resources/lang/de_DE/default.po
```diff
@@ -531,6 +531,18 @@ msgstr "Benötigte Engel"
 msgid "Shift deleted."
 msgstr "Schicht gelöscht."
 
+msgid "Shifts history"
+msgstr "Schichten historie"
```
weeman1337, line 535:
```suggestion
msgstr "Schichten Historie"
```

includes/model/Shifts_model.php
```diff
@@ -556,13 +556,22 @@ function Shift_update($shift)
     );
 }
 
+/**
+ * Get the next free shifts transaction id
+ */
+function Shift_get_next_transaction_id(): int
+{
+    return Db::selectOne('SELECT MAX(transaction_id) + 1 AS transaction_id FROM Shifts')['transaction_id'] ?? 1;
```
weeman1337, line 564: As discussed I would suggest one of these options:
- Lock the shifts table during shift creation
- Use a uuid as shift transaction id
- Create a shifttransaction table and reference to it as shift transaction id

MyIgel: UUIDs it is

weeman1337 (approved):

PR #923, opened 2022-06-08
Settings Themes rewrite

frischler:
- The theme settings are no longer handled in _includes/pages/user_settings.php_ and _includes/view/User_view.php_ but in the _SettingsController.php_.
- The theme settings are now have a separate section, just as passwords.
- The german wording "Aussehen" has been changed to also be "Theme".

themes

### DIFF
--- a/resources/views/pages/settings/theme.twig
+++ b/resources/views/pages/settings/theme.twig
@@ -1,0 +1,19 @@
+{% extends 'pages/settings/settings.twig' %}
+{% import 'macros/form.twig' as f %}
+{% import 'macros/base.twig' as m %}
+
+{% block title %}{{ __('settings.theme') }}{% endblock %}
+
+{% block row_content %}
+    <form action="" enctype="multipart/form-data" method="post">
+        {{ csrf() }}
+
+        <div class="row">
+            <div class="col-md-12">
+                {{ m.info(__('settings.theme.info')) }}
+                {{ f.select('select_theme', themes, __('settings.theme'), current_theme) }}
+                {{ f.submit() }}
+            </div>
+        </div>
+    </form>
+{% endblock %}
--- a/src/Controllers/SettingsController.php
+++ b/src/Controllers/SettingsController.php
@@ -99,6 +99,49 @@ class SettingsController extends BaseController
     /**
      * @return Response
      */
+    public function theme(): Response
+    {
+        $themes = array_map(function ($theme) {
+            return $theme['name'];
+        }, config('themes'));
+
+        $currentTheme = $this->auth->user()->settings->theme;
+
+        return $this->response->withView(
+            'pages/settings/theme',
+            [
+                'settings_menu' => $this->settingsMenu(),
+                'themes'        => $themes,
+                'current_theme' => $currentTheme
+            ] + $this->getNotifications()
+        );
+    }
+
+    /**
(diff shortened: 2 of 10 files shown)


src/Controllers/SettingsController.php
```diff
@@ -96,6 +96,49 @@ public function savePassword(Request $request): Response
...
+        }, config('themes'));
+
+        $currentTheme = $this->auth->user()->settings->theme;
+
+        return $this->response->withView(
+            'pages/settings/theme',
+            [
+                'settings_menu' => $this->settingsMenu(),
+                'themes' => $themes,
+                'current_theme' => $currentTheme
```
MyIgel, line 115: I would have alligned the `=>` and added a `,` after the last one but thats only a minor stylistic change

resources/views/pages/settings/theme.twig
```diff
@@ -0,0 +1,19 @@
+{% extends 'pages/settings/settings.twig' %}
+{% import 'macros/form.twig' as f %}
+{% import 'macros/base.twig' as m %}
+
+{% block title %}{{ __('settings.theme') }}{% endblock %}
+
+{% block row_content %}
+        <form action="" enctype="multipart/form-data" method="post">
```
MyIgel, line 8: Why 8 spaces? 0o

tests/Unit/Controllers/SettingsControllerTest.php
```diff
@@ -329,7 +418,20 @@ public function setUp(): void
         parent::setUp();
         $this->initDatabase();
 
-        $this->config = new Config(['min_password_length' => 6]);
+        $this->themes = [
+            0 => [
```
MyIgel, line 422: The explicit numerical naming is not really needed here but ok as its explicitly used as a reference

tests/Unit/Controllers/SettingsControllerTest.php
```diff
@@ -329,7 +418,20 @@ public function setUp(): void
...
+                'name' => 'Engelsystem light',
+                'type' => 'light',
+                'navbar_classes' => 'navbar-light bg-light',
```
MyIgel, line 425: to keep things short i would personally not add more data then needed ^^

src/Controllers/SettingsController.php
```diff
@@ -96,6 +96,49 @@ public function savePassword(Request $request): Response
...
+     * @param Request $request
+     * @return Response
+     */
+    public function saveTheme(Request $request): Response
+    {
+        $user = $this->auth->user();
+        $data = $this->validate($request, ['select_theme' => 'int']);
+        $selectTheme = $data['select_theme'];
+
+        if (!array_key_exists($selectTheme, config('themes'))) {
```
MyIgel, line 130: a `if (isset(config('themes')[$selectTheme]))` would be also feasible but the explicit form is also good here

tests/Unit/Controllers/SettingsControllerTest.php
```diff
@@ -329,7 +418,20 @@ public function setUp(): void
         parent::setUp();
         $this->initDatabase();
 
-        $this->config = new Config(['min_password_length' => 6]);
+        $this->themes = [
```
MyIgel, line 421: Why is this saved as a class var if it has no further usage as one?

PR #944, opened 2022-07-13
Mobile required Option

PowerPan:
Add Option to make mobile number required for Registration

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -258,6 +258,9 @@ return [
 
     // Whether the DECT field should be enabled
     'enable_dect'             => (bool)env('ENABLE_DECT', true),
+    
+    // Whether the Mobile field should be required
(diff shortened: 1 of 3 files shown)


PowerPan:

MyIgel (changes requested):
At the moment this PR has merge conflicts so a rebase would be required here.
Also to implement this option properly the same checks should be added to the users settings page (which will interfere with #972 (Settings profile modernization) so I would recommend eventually waiting for that one to be merged first)

weeman1337 (changes requested):
Hey @PowerPan :wave: Can you fix the issues mentioned by previous reviewers and solve the conflicts in your branch? If we don't hear from you for another two weeks, we may unfortunately close this pull request.

weeman1337:

MyIgel:
Hi @PowerPan, for what event do you need that change and are you able to update this PR to also cover the user page input? If yes, feel free to reopen this PR :)

PR #948, opened 2022-07-15
Custom translations

PowerPan:
Add Possibilty to add Custom Tranlations in folder:

"config/lang/__LOCALE__/custom.po

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ _vimrc_local.vim
 
 # Project files
 /config/config.php
+/config/lang/
 /test/coverage
 /public/coverage
 /coverage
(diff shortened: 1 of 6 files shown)


PowerPan:

MyIgel (changes requested):
Hi, thanks for the PR!
* Imho local would  be a better name? But custom should be fine too.
* If you run `phpcs` it will show you some issues with the formatting and imports -> could you fix them too please?
* And when splitting up code the tests have to be updated accordingly to also count the lines to the new coverage.

weeman1337 (changes requested):
Hey @PowerPan :wave:  Can you fix the issues mentioned by previous reviewers and solve the conflicts in your branch? If we don't hear from you for another two weeks, we may unfortunately close this pull request.

PR #950, opened 2022-07-15
Shift table view

PowerPan:
Add Table View For Shifts
Bildschirmfoto 2022-07-16 um 00 30 56


PowerPan:

weeman1337 (changes requested):
Hey @PowerPan :wave: Can you fix the issues mentioned by previous reviewers and solve the conflicts in your branch? If we don't hear from you for another two weeks, we may unfortunately close this pull request.

weeman1337:

MyIgel:
On the other side as noted above:
:exclamation:  This is a feature that needs to be maintained and heavily relies on legacy code.

PR #952, opened 2022-07-18
Fix barchart division by 0 error

weeman1337:
:ping_pong: @MyIgel


MyIgel:
Thx for the fix! :heart_eyes:

PR #956, opened 2022-07-21
Fix dropdown overflow → make dropdown scrollable

weeman1337:
:ping_pong: @MyIgel


MyIgel:
At least my firefox hides one room there at the end :(

weeman1337:
> At least my firefox hides one room there at the end :(

Check again.

PR #959, opened 2022-07-23
Add direct link to the angeltyes page from user view

weeman1337:
grafik

:ping_pong:  @MyIgel

### DIFF
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -1,6 +1,7 @@
 <?php
 
 use Carbon\Carbon;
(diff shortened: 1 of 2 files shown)


weeman1337:
Not sure if we should open the link in a new tab to prevent form input loss.

MyIgel:
Eventually a text like "You can use the Angeltypes Page to change your Angeltypes" would be more fitting?
And a new tab sounds eventually like a good idea

MyIgel:
Hmm, on the other side: normally you don't change anything there and might only use it to navigate there -> hmm

PR #961, opened 2022-07-24
Remove jquery-ui :axe: 

weeman1337:
:ping_pong: @MyIgel


MyIgel:
Something package.json something?

weeman1337:
Something something @MyIgel if you're back from the loo, have another look :laughing:

PR #969, opened 2022-10-01
Extend contributing guidelines

MyIgel:
Update the contributing guidelines to make it clearer and easier to maintain.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,10 +4,22 @@
   You may use `composer run phpcs` and [Editorconfig-Checker](https://editorconfig-checker.github.io) to verify that.
 * Use `use` statements wherever possible instead of writing the fully qualified name.
 * Order the composer/npm dependencies alphabetically.
-* Do not use code from the [includes](./includes) directory anywhere else.
-* Please cover your code by unit tests. Code under `includes` does not require tests.
+* Do not use code from the [includes](includes) directory anywhere else.
+* Please cover your code by unit tests, our goal is to stay at 100% line coverage.
+  Code under `includes` does not require tests as it's mostly not testable and needs to be rewritten.
 * Do not use vendor prefixes like `-webkit` in styles.
-  This is done by PostCSS + Autoprefixer according to the [`.browserslistrc`](./.browserslistrc).
+  This is done by PostCSS + Autoprefixer according to the [`.browserslistrc`](.browserslistrc).
+* Translations must be abbreviated, for example `form.save`.
+  The `default.po` files contain translations that can be auto-detected using Poedit, `additional.po` contains generated messages like validations.
 
 ## Pull requests
-Please create single pull requests for every feature instead of creating one big monster of pull request containing a complete rewrite.
+* The PR should contain a short overview of the changes.
+* Before implementing bigger changes, please open an issue to discuss the feature and possible implementation options.
+* Please create single pull requests for every feature instead of creating one big monster of pull request containing a complete rewrite.
+* Squash similar commits to make the review easier.
+* For visual changes, include both before and after screenshots to easily compare and discuss changes.
+
+## Commits
+* The commit message must be meaningful. It should serve as a short overview of the changes.
+  If needed, an additional description can be provided.
+* A commit should be self-contained and result in a working Engelsystem.


MyIgel:
@msquare, @weeman1337 & @marudor do you have any other suggestions?

CONTRIBUTING.md
```diff
@@ -5,9 +5,17 @@
...
+* Please cover your code by unit tests, our goal is 100% line coverage. Code under `includes` does not require tests as it's mostly not testable.
 * Do not use vendor prefixes like `-webkit` in styles.
   This is done by PostCSS + Autoprefixer according to the [`.browserslistrc`](./.browserslistrc).
 
 ## Pull requests
-Please create single pull requests for every feature instead of creating one big monster of pull request containing a complete rewrite.
+* The PR should contain a short overview of the changes.
+* Before implementing bigger changes, please open an issue to discuss the feature and possible implementation options.
+* Please create single pull requests for every feature instead of creating one big monster of pull request containing a complete rewrite.
+* Squash similar commits to make the review easier.
```
weeman1337, line 16: I would just enable squash commit PR merges for the repo and skip this point. In my opinion it's not worth the discussions. If the commits of a PR are not great: just squash it :hammer:

CONTRIBUTING.md
```diff
@@ -5,9 +5,17 @@
...
+
+## Commits
+* The commit message must be meaningful. It should serve as a short overview of the changes.
+  If needed, an additional description can be provided.
+* A commit should be self-contained and result in a working engelsystem.
```
weeman1337, line 21:
```suggestion
* A commit should be self-contained and result in a working Engelsystem.
```

weeman1337:
Nobody reads that anyway :laughing:

CONTRIBUTING.md
```diff
@@ -5,9 +5,17 @@
...
 ## Pull requests
-Please create single pull requests for every feature instead of creating one big monster of pull request containing a complete rewrite.
+* The PR should contain a short overview of the changes.
+* Before implementing bigger changes, please open an issue to discuss the feature and possible implementation options.
+* Please create single pull requests for every feature instead of creating one big monster of pull request containing a complete rewrite.
+* Squash similar commits to make the review easier.
+
+## Commits
+* The commit message must be meaningful. It should serve as a short overview of the changes.
+  If needed, an additional description can be provided.
```
weeman1337, line 20: I would prefer code comments to commit descriptions.

weeman1337:
I think for visual changes it would make sense to ask for some screenshots showing it **before** and **after** the changes.

CONTRIBUTING.md
MyIgel: That's something only @msquare can do because i sadly don't have the permissions to :(

MyIgel: But in general ppl should also just do useful commits and that normally also means squashing

MyIgel:
> some screenshots showing it **before** and **after** the changes.

Do you have some wording suggestions for that? ^^

CONTRIBUTING.md
msquare: All possible merge options are already enabled, can you please check?

CONTRIBUTING.md
```diff
@@ -5,9 +5,17 @@
 * Use `use` statements wherever possible instead of writing the fully qualified name.
 * Order the composer/npm dependencies alphabetically.
 * Do not use code from the [includes](./includes) directory anywhere else.
-* Please cover your code by unit tests. Code under `includes` does not require tests.
+* Please cover your code by unit tests, our goal is 100% line coverage. Code under `includes` does not require tests as it's mostly not testable.
```
msquare, line 8: Right now, it is not only our goal to have 100% line coverage in the new source folder, but it is required to get the PR merged. May we should make that more clear?

weeman1337:
> > some screenshots showing it **before** and **after** the changes.
> 
> Do you have some wording suggestions for that @weeman1337? ^^

:man_shrugging: „For visual changes, include both before and after screenshots to easily compare and discuss what's changing.“

CONTRIBUTING.md
MyIgel: It is & i guess i was dumb :upside_down_face:

MyIgel: Anyhow, i would prefer to have it there so that eventually ppl will just squash their commits as it might be useful to just have two or three instead of ten changing every file by its own

PR #971, opened 2022-10-16
add option to show mobile number of user in users view

### LINKED ISSUE #957: Mobile number not visible
Not every angel has a DECT phone. While the system allows you to add a mobile number to your profile, it does not show for other people.

xuwhite:
Now there is an option to show the mobile number
both opt-in at user_view and config.default
Fixes #957

### DIFF
--- a/includes/pages/guest_login.php
+++ b/includes/pages/guest_login.php
@@ -36,6 +36,7 @@ function guest_register()
     $min_password_length = config('min_password_length');
     $enable_password = config('enable_password');
     $enable_pronoun = config('enable_pronoun');
+    $enable_mobile_show = config('enable_mobile_show');
     $config = config();
     $request = request();
     $session = session();
@@ -49,6 +50,7 @@ function guest_register()
     $preName = '';
     $dect = '';
     $mobile = '';
+    $mobile_show = false;
     $email = '';
     $pronoun = '';
     $email_shiftinfo = false;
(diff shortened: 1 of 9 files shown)


includes/pages/guest_login.php
```diff
@@ -119,6 +121,10 @@ function guest_register()
             $msg .= error(__('Please enter a nickname.'), true);
         }
 
+        if ($request->has('mobile_show') && $enable_mobile_show) {
```
MyIgel, line 124: This implies that the consent will be revoked on save if the setting gets disabled temporarily and users update their consent setting

includes/view/User_view.php
```diff
@@ -82,6 +83,14 @@ function User_settings_view(
                             ) : '',
                             $enable_dect ? form_text('dect', __('DECT'), $user_source->contact->dect, false, 40) : '',
                             form_text('mobile', __('Mobile'), $user_source->contact->mobile, false, 40),
+                            $enable_mobile_show ? form_checkbox(
+                                'mobile_show',
+                                __(
+                                    'Show mobile number to other users to contact me',
+                                    [config('app_name')]
```
MyIgel, line 90: Thats useless here and seems to be a c&p artifact

includes/pages/user_settings.php
```diff
@@ -91,7 +91,10 @@ function user_settings_main($user_source, $enable_tshirt_size, $tshirt_sizes)
         }
     }
     $user_source->contact->mobile = strip_request_item('mobile', $user_source->contact->mobile);
-
+    if (config('enable_mobile_show')) {
+        $user_source->settings->mobile_show = $request->has('mobile_show');
+    }
```
MyIgel, line 96: There are useless spaces

PR #972, opened 2022-10-17
Settings profile modernization

frischler:
The users' profile settings page (former _/user-settings_) has been modernized and is now reachable under _/settings/profile_. The file `user_settings.php` and the `User_settings_view` method in `User_view.php` has been removed. The UI of the profile settings page nearly stayed the same, besides of some spacing differences to clarify the sections.

Localizations used in _/user-settings_ hasn't yet been removed, since there are still used in other pages, like the `guest_login.php`.


tests/Unit/Http/Validation/ValidatorTest.php
```diff
@@ -60,6 +60,27 @@ public function testValidateChaining()
         ));
     }
 
+    /**
+     * @covers \Engelsystem\Http\Validation\Validator::validate
+     */
+    public function testValidateMultipleParameters()
```
frischler, line 66: There was not yet a test for multiple parameters in a validation. Since I was now using one in the SettingsController, I added the test here, too.

resources/views/macros/form.twig
```diff
@@ -55,14 +63,18 @@
     </div>
 {%- endmacro %}
 
-{% macro checkbox(name, label, checked, value, disabled) %}
+{% macro checkbox(name, label, checked, value, disabled, raw_label) %}
```
frischler, line 66: raw needed for the privacy_email in the label.

frischler:
> Hi @frischler, could you rebase the branch on the current main? At the moment it can't be merged :(

Good notice. Done as requested.

MyIgel (commented):
Some wip comments:

src/Controllers/SettingsController.php
```diff
@@ -52,6 +53,103 @@ public function __construct(
...
+            $user->personalData->pronoun = $data['pronoun'];
+        }
+
+        if (config('enable_user_name')) {
+            $user->personalData->first_name = $data['first_name'];
+            $user->personalData->last_name = $data['last_name'];
+        }
+
+        if (config('enable_planned_arrival')) {
+            if (!$this->isArrivalDateValid($data['planned_arrival_date'], $data['planned_departure_date'])) {
```
MyIgel, line 107: This case is not covered by unit tests ;)

src/Controllers/SettingsController.php
```diff
@@ -52,6 +53,103 @@ public function __construct(
...
+                $this->addNotification('settings.profile.planned_arrival_date.invalid', 'errors');
+                return $this->redirect->to('/settings/profile');
+            } elseif (!$this->isDepartureDateValid($data['planned_arrival_date'], $data['planned_departure_date'])) {
+                $this->addNotification('settings.profile.planned_departure_date.invalid', 'errors');
```
MyIgel, line 111: And that one too

src/Controllers/ChecksArrivalsAndDepartures.php
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        if (!is_null($departure_carbon) && $arrival_carbon->greaterThan($departure_carbon)) {
+            return false;
+        }
+
+        return !$this->isBeforeBuildup($arrival_carbon) && !$this->isAfterTeardown($arrival_carbon);
+    }
+
+    protected function isDepartureDateValid(?string $arrival, ?string $departure): bool
```
MyIgel, line 26: I personally prefer having doc blocks for visual seperation but thats just a note from my side

src/Controllers/ChecksArrivalsAndDepartures.php
```diff
@@ -0,0 +1,56 @@
...
+    private function toCarbon(?string $date_string): ?Carbon
+    {
+        $dateTime = DateTime::createFromFormat('Y-m-d', $date_string ?: '');
+        return $dateTime ? new Carbon($dateTime) : null;
+    }
+
+    private function isBeforeBuildup(Carbon $date): bool
+    {
+        $buildup = config('buildup_start');
+        return !empty($buildup) && $date->lessThan($buildup->setTime(0, 0));
```
MyIgel, line 48: There is a `->startOfDay()` method that can be used here (which also resets seconds and microseconds)

src/Controllers/ChecksArrivalsAndDepartures.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private function isAfterTeardown(Carbon $date): bool
+    {
+        $teardown = config('teardown_end');
+        return !empty($teardown) && $date->greaterThanOrEqualTo($teardown->addDay()->setTime(0, 0));
```
MyIgel, line 54: Same here

resources/views/pages/settings/profile.twig
```diff
@@ -0,0 +1,151 @@
...
+                    'shirt_size',
+                    config('tshirt_sizes'),
+                    __('settings.profile.shirt_size'),
+                    user.personalData.shirt_size
+                ) }}
+            </div>
+
+            <div class="col-md-12">
+                {{ m.info(__('settings.profile.angeltypes.info', [url('/angeltypes')]), true) }}
+                {{ f.submit() }}
```
MyIgel, line 147: Some separation between the button and the text would be nice here, at the moment its in the same line as the info span

resources/views/pages/settings/profile.twig
```diff
@@ -0,0 +1,151 @@
...
+                            'value': user.personalData.planned_arrival_date.format('Y-m-d'),
+                            'required': true,
+                            'entry_required_icon': true,
+                            'min': config('buildup_start') ? config('buildup_start').format('Y-m-d') : '',
+                            'max': config('teardown_end') ? config('teardown_end').format('Y-m-d') : '',
+                        }
+                    ) }}
+                    {% set planned_departure_date = user.personalData.planned_departure_date %}
+                    {{ f.input(
+                        'planned_departure_date',
```
MyIgel, line 62: Removing the date does not validate correctly as its not that optional

resources/views/pages/settings/profile.twig
```diff
@@ -0,0 +1,151 @@
...
+                        user.settings.email_goody,
+                        user.settings.email_goody,
+                        false,
+                        true
+                    ) }}
+                {% endif %}
+            </div>
+
+            <div class="col-md-12">
+                {{ f.select(
```
MyIgel, line 137: I think we should mark this as required too as we don't have the option to set it as empty

resources/views/pages/settings/profile.twig
```diff
@@ -0,0 +1,151 @@
...
+                        email_goody_label,
+                        user.settings.email_goody,
+                        user.settings.email_goody,
+                        false,
+                        true
+                    ) }}
+                {% endif %}
+            </div>
+
+            <div class="col-md-12">
```
MyIgel, line 136: We have a `enable_tshirt_size` option which should be checked here

src/Controllers/SettingsController.php
```diff
@@ -52,6 +53,103 @@ public function __construct(
...
+        $user->email = $data['email'];
+        $user->settings->email_shiftinfo = $data['email_shiftinfo'] ?: false;
+        $user->settings->email_news = $data['email_news'] ?: false;
+        $user->settings->email_human = $data['email_human'] ?: false;
+
+        if (config('enable_goody')) {
+            $user->settings->email_goody = $data['email_goody'] ?: false;
+        }
+
+        if (isset(config('tshirt_sizes')[$data['shirt_size']])) {
```
MyIgel, line 138: The `enable_tshirt_size` should be checked too

MyIgel (changes requested):
I found some more things:

resources/views/macros/base.twig
```diff
@@ -36,8 +36,19 @@
...
+        {%- if raw|default(false) -%}
+            {{ text|raw }}
+        {%- else -%}
+            {{ text }}
+        {%- endif -%}
+    </span>
+{%- endmacro %}
+
+{% macro entry_required(text) %}
+    <h4 class="help-block">{{ _self.icon('exclamation-triangle', 'info') }}{{ text }}</h4>
```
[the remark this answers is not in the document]
frischler: So, I shall do the following things?
- change the _entry_required_ macro to only be an icon wrapper.
- the macro to also contain the text shall be defined inside the page in which it is used.
- change the h4 to be a paragraph. (I was using h4 because that's how it was in the old page, too. Still, you are right.)

src/Controllers/SettingsController.php
```diff
@@ -52,6 +53,97 @@ public function __construct(
...
+            $user->personalData->shirt_size = $data['shirt_size'];
+            $user->personalData->save();
```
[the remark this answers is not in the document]
frischler: Really? Where?
Or do you mean the _UserShirtController_? Because the only duplication I see there are these 4 lines.
In the UserShirtController, I could make these lines a public function and let the SettingsController call it. I didn't do it because I didn't see the huge benefit in it in this case. But I'm open for suggestions.

src/Controllers/SettingsController.php
```diff
@@ -52,6 +53,97 @@ public function __construct(
...
+    /**
+     * @param Request $request
+     * @return Response
+     */
+    public function saveProfile(Request $request): Response
+    {
+        $user = $this->auth->user();
+
+        config('buildup_start');
+        config('teardown_end');
```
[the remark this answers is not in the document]
frischler: They ... are. Oops.

src/Controllers/SettingsController.php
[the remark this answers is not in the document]
frischler: Or you probably meant just the $user->personalData->save();

src/Controllers/SettingsController.php
MyIgel: @frischler it still isn't (same as the next row) ^^

resources/views/pages/settings/profile.twig
MyIgel: @frischler that's still an issue if i see that right

MyIgel:
Or i'm dumb *dumdidum

MyIgel (changes requested):
Just a missing "this is required" flag on the tshirts select then its fine

PR #975, opened 2022-10-21
Migrations: Removed references to models to prevent further migration issues

MyIgel:
Fixes the issue that models have values that are not yet added to the database, for example in user settings.


MyIgel:
@weeman1337 :eyes:

weeman1337 (approved):
looks sane :+1:

Issue #740, opened 2020-05-01
RFC/Draft: New permissions concept


MyIgel:
# Objective
* Cleanup/group the permissions
* Make the Engelsystem more "widely usable" by using better wording (at least in the backend)
* Permissions hierarchy: `foo.bar.edit`, `foo.bar.view`
* Make permissions more maintainable and extendable (`lorem` also grants `lorem.ipsum`)

# Main ideas
## Renaming
| before | after | example |
| -------- | -------- | -------- |
| engel | user | admin, foofbar123 |
| engeltype | team | Angel, Infodesk, LOC Coordinator, Camera Angel |
| group | role | Guest, User, Bureaucrat, Admin |
| right/privilege | permission | `startpage.view`, `faq.view`, `shifts.edit` |

### Users
user (formerly angel) have teams (formerly angeltypes)

## Inheritance
Users have teams which have optional roles with optional permissions
```
   Team
   |  |
User  Role
       |
     Permission
```

MyIgel:
Eventually split the team into more angeltype equivalent subparts which then have roles

DarkGigaByte:
Wie besprochen lasse ich dir diesen Link mal hier https://symfony.com/doc/current/security.html#hierarchical-roles
und für weitere Recherche dieser Suchbegriff: RBAC (Role-based access-control)
OWASP phprbac: http://phprbac.net/
NIST Paper: https://csrc.nist.gov/projects/role-based-access-control

teal-bauer:
For more inspiration, rescheduled has RBAC: https://git.cccv.de/c3lingo/rescheduled/-/commit/874d68a308f7ee6b91c6c04ba0ce14092a8214aa

PR #980, opened 2022-10-24
Permissions refactoring

MyIgel:
Replaced Privileges/Groups with corresponding models (the rename and restructuring according to #740 (RFC/Draft: New permissions concept) has to be done at a later stage ;)


MyIgel:
@weeman1337 / @msquare could you have a look / review / test this one and point me to things that i missed? :D

MyIgel:
Another open question would be how to proceed with the "sorting" of groups, eventually the user edit page should be changed to only allow assigning groups to others that an angel has itself?

MyIgel:
(And those commits should definitely be squashed on merge as they leave a broken engelsystem on their own)

weeman1337 (approved):
Looks good

PR #983, opened 2022-11-08
Refactor ShiftTypes

MyIgel:
* Removed referenced angeltype from shift type as it got never properly used (besides being displayed)
* Replaced ShiftTypes with shift_types / ShiftType model


db/factories/Shifts/ShiftTypeFactory.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Database\Factories\Engelsystem\Models\Shifts;
+
+use Engelsystem\Models\Shifts\ShiftType;
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+class ShiftTypeFactory extends Factory
+{
+    /** @var string */
+    protected $model = ShiftType::class;
```
weeman1337, line 11: PHP prop types

```suggestion
    protected string $model = ShiftType::class;
```

includes/view/ShiftTypes_view.php
```diff
@@ -57,56 +59,56 @@ function ShiftType_edit_view($name, $description, $shifttype_id)
...
         ]),
         heading(__('Description'), 2),
-        $parsedown->parse((string)$shifttype['description'])
+        $parsedown->parse($shifttype->description)
     ], true);
 }
 
 /**
- * @param array[] $shifttypes
+ * @param ShiftType[]|array[]|\Illuminate\Support\Collection $shifttypes
```
weeman1337, line 90: use `use`

weeman1337 (approved):

db/factories/Shifts/ShiftTypeFactory.php
MyIgel: Sadly not possible here as it inherits the base class which has no type definition here

PR #993, opened 2022-11-26
ESLint

weeman1337:
:ping_pong: @MyIgel


MyIgel (changes requested):
At the moment the container build  fails:

```
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
```

PR #997, opened 2022-11-27
Added AngelTypes model

MyIgel:


MyIgel:
@weeman1337 :wave:

includes/controller/angeltypes_controller.php
```diff
@@ -340,22 +351,51 @@ function angeltypes_list_controller()
...
+function AngelTypes_with_user($userId): Collection
+{
+    return AngelType::query()
+        ->select([
+            'angel_types.*',
+            'UserAngelTypes.id AS user_angeltype_id',
+            'UserAngelTypes.confirm_user_id',
+            'UserAngelTypes.supporter',
+        ])
+        ->leftJoin('UserAngelTypes', function ($join) use ($userId) {
```
weeman1337, line 394:
```suggestion
        ->leftJoin('UserAngelTypes', function (JoinClause $join) use ($userId) {
```

\+ remove `/** @var JoinClause $join */`

src/Models/AngelType.php
```diff
@@ -0,0 +1,73 @@
...
+ * @method static QueryBuilder|AngelType[] whereContactName($value)
+ * @method static QueryBuilder|AngelType[] whereContactDect($value)
+ * @method static QueryBuilder|AngelType[] whereContactEmail($value)
+ * @method static QueryBuilder|AngelType[] whereRestricted($value)
+ * @method static QueryBuilder|AngelType[] whereRequiresDriverLicense($value)
+ * @method static QueryBuilder|AngelType[] whereNoSelfSignup($value)
+ * @method static QueryBuilder|AngelType[] whereShowOnDashboard($value)
+ * @method static QueryBuilder|AngelType[] whereHideRegister($value)
+ */
+class AngelType extends BaseModel
```
weeman1337, line 35: If angel types should be ordered by name by default we should probably set it on the model instead of every query

Something like

```
protected static function boot(): void
{
    parent::boot();
    static::addGlobalScope('order', fn (Builder $builder) => $builder->orderBy('name');
}
```

src/Models/AngelType.php
```diff
@@ -0,0 +1,73 @@
...
+{
+    use HasFactory;
+
+    /** @var array */
```
weeman1337, line 39:
```suggestion
    /** @var string[] */
```

src/Models/AngelType.php
```diff
@@ -0,0 +1,73 @@
...
+        'contact_email',
+
+        'restricted',
+        'requires_driver_license',
+        'no_self_signup',
+        'show_on_dashboard',
+        'hide_register',
+    ];
+
+    /** @var string[] */
```
weeman1337, line 55:
```suggestion
    /** @var array<string, string> */
```

src/Models/AngelType.php
```diff
@@ -0,0 +1,73 @@
...
+    protected $casts = [
+        'restricted'              => 'boolean',
+        'requires_driver_license' => 'boolean',
+        'no_self_signup'          => 'boolean',
+        'show_on_dashboard'       => 'boolean',
+        'hide_register'           => 'boolean',
+    ];
+
+    /**
+     * @return bool
```
weeman1337, line 65: Redundant doc block, can be removed

tests/Unit/Models/AngelTypeTest.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Engelsystem\Test\Unit\Models;
+
+use Engelsystem\Models\AngelType;
+
+class AngelTypeTest extends ModelTest
+{
+    /**
+     * @return array[]
```
weeman1337, line 10:
```suggestion
     * @return array<array{boolean, string, string, string}>
```

weeman1337 (commented):
Tested the migration and function in general. Seems to be fine :+1: 

Some smaller comments :wave:

db/migrations/2022_11_08_000000_create_angel_types_table.php
```diff
@@ -0,0 +1,125 @@
...
+            $table->boolean('no_self_signup')->default(false);
+            $table->boolean('show_on_dashboard')->default(true);
+            $table->boolean('hide_register')->default(false);
+        });
+
+        if (!$this->schema->hasTable('AngelTypes')) {
+            return;
+        }
+
+        /** @var stdClass[] $records */
```
weeman1337, line 41: I think this returns an Illuminate Collection or something like that.

weeman1337 (approved):
Looks good :+1:

db/migrations/2022_11_08_000000_create_angel_types_table.php
MyIgel: It does (and that contains the stdClass instances)

PR #1000, opened 2022-11-28
codestyle

ximex:
unify indentation in config.default.php
unify javascript codestyle in sticky-headers.js

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -280,7 +280,7 @@ return [
     'enable_goody'            => (bool)env('ENABLE_GOODY', false),
 
     // Enables the food voucher in the user profile
-    'enable_voucher'            => (bool)env('ENABLE_VOUCHER', true),
+    'enable_voucher'          => (bool)env('ENABLE_VOUCHER', true),
 
     // Number of shifts to freeload until angel is locked for shift signup.
     'max_freeloadable_shifts' => env('MAX_FREELOADABLE_SHIFTS', 2),
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,6 +3,7 @@
   "extends": [ "plugin:editorconfig/all" ],
   "plugins": [ "editorconfig" ],
   "rules": {
+    "function-paren-newline": "error",
     "prefer-arrow-callback": "error",
     "no-var": "error",
     "quotes": [
--- a/resources/assets/js/dashboard.js
+++ b/resources/assets/js/dashboard.js
@@ -25,9 +25,8 @@ ready(() => {
   document.getElementById('dashboard-fullscreen')
     ?.addEventListener('click', (event) => {
       event.preventDefault();
-      document.querySelectorAll(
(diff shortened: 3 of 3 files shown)


config/config.default.php
```diff
@@ -278,7 +278,7 @@
     'enable_goody'            => (bool)env('ENABLE_GOODY', false),
 
     // Enables the food voucher in the user profile
-    'enable_voucher'            => (bool)env('ENABLE_VOUCHER', true),
+    'enable_voucher'          => (bool)env('ENABLE_VOUCHER', true),
```
weeman1337, line 281: Can you try to add a rule that covers this style to the [phpcs rule set](https://github.com/engelsystem/engelsystem/blob/main/.phpcs.xml)?

resources/assets/js/sticky-headers.js
```diff
@@ -14,19 +14,18 @@ $(function () {
...
-        timeLanes.css({
-          'left': Math.max(0, $(window).scrollLeft() - left) + 'px'
-        });
-        headers.css({
-          'top': Math.max(0, $(window).scrollTop() - top
-                        - 13
-                        + topReference.offset().top)
-                        + 'px'
-        });
+    $(window).scroll(function () {
```
weeman1337, line 17: Can you try to add a rule that covers this style to the [ESLint rule set](https://github.com/engelsystem/engelsystem/blob/main/.eslintrc.json)?

weeman1337 (commented):
I would say we should not format code if it passes the code style linters. Otherwise people will format code back and forth.

resources/assets/js/sticky-headers.js
ximex: done

config/config.default.php
ximex: never worked with phpcs. i only aligned it as all the other indentations

weeman1337: Never mind. I think this only works for direct siblings.

weeman1337 (changes requested):
Can you rebase your PR and re-request review?

weeman1337 (approved):

weeman1337 (changes requested):
Can you fix the current ESLint errors and then re-request review?

```
~/git/chaos/engelsystem (codestyle)$ yarn run lint
yarn run v1.22.18
$ eslint .

/home/weeman/git/chaos/engelsystem/resources/assets/js/dashboard.js
  28:32  error  Unexpected newline after '('   function-paren-newline
  30:7   error  Unexpected newline before ')'  function-paren-newline

✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

weeman1337 (approved):
Looks good now. Thanks :+1:

PR #1003, opened 2022-11-29
use template literals instead of string concatenation

ximex:


weeman1337 (changes requested):
Thank you for your contribution :+1: 

Can you apply the following changes?:
- Please add the `prefer-template` ESLint rule
- We don't want tu call functions in template literals. Can you extract them to a variable before?
- Finally this rule should be mentioned in `CONTRIBUTING.md`

If you have questions just ask. Re-request review once these points are done.

ximex:
which rule? to use template rules or/and to don't put function calls into them?

weeman1337 (changes requested):
> which rule? to use template rules or/and to don't put function calls into them?

We cannot enforce the „no function call in templates“ by tooling → needs to go to the [CONTRIBUTING.md](https://github.com/engelsystem/engelsystem/blob/main/CONTRIBUTING.md) for visibility.

Can you add it please? Looks good otherwise.

weeman1337 (approved):
Looks good now, thanks :+1:

PR #1004, opened 2022-11-29
fix border radius on time input in shift filter

ximex:
before
grafik
after
grafik


weeman1337 (approved):
Looks sane :+1: 

Thank you for your contribution @ximex

PR #1007, opened 2022-11-29
Checkbox UI

ximex:
removed unused function `form_multi_checkboxes`
style checkboxes with bootstrap classes

before:
grafik
after:
grafik

### DIFF
--- a/includes/pages/user_shifts.php
+++ b/includes/pages/user_shifts.php
@@ -372,10 +372,11 @@ function make_select($items, $selected, $name, $title = null, $additionalButtons
     $html .= buttons($buttons);
 
     foreach ($items as $i) {
-        $htmlItems[] = '<div class="checkbox">'
-            . '<label><input type="checkbox" name="' . $name . '[]" value="' . $i['id'] . '" '
+        $id = $name . '_' . $i['id'];
+        $htmlItems[] = '<div class="form-check">'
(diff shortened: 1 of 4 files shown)


includes/pages/user_shifts.php
```diff
@@ -372,10 +372,11 @@ function make_select($items, $selected, $name, $title = null, $additionalButtons
...
 
     foreach ($items as $i) {
-        $htmlItems[] = '<div class="checkbox">'
-            . '<label><input type="checkbox" name="' . $name . '[]" value="' . $i['id'] . '" '
+        $id = $name . '_' . $i['id'];
+        $htmlItems[] = '<div class="form-check">'
+            . '<input class="form-check-input" type="checkbox" id="' . $id . '" name="' . $name . '[]" value="' . $i['id'] . '" '
             . (in_array($i['id'], $selected) ? ' checked="checked"' : '')
-            . ' > ' . $i['name'] . '</label>'
+            . ' /><label class="form-check-label" for="' . $id . '">' . $i['name'] . '</label>'
```
MyIgel, line 379: xhtml is not needed hare?

MyIgel:
```suggestion
            . ' ><label class="form-check-label" for="' . $id . '">' . $i['name'] . '</label>'
```

ximex:
I'm a fan of the self-closing tags. also things like `disabled="disabled"`. but could change it if not wanted

MyIgel:
That would be great :)

PR #1013, opened 2022-12-01
fix select_array. needed for a correct format for select2

ximex:
if `$data` is a `Collection` than the return format is `[{"key":["value"]}]` and not `[{"key":"value"}]`


MyIgel:
Thank you for the fix!

PR #1019, opened 2022-12-05
Improve spinner

ximex:
used better id's
spinner input is now from type `number`
use the javascript api to decrease/increase the value on button click

### DIFF
--- a/includes/sys_form.php
+++ b/includes/sys_form.php
@@ -20,13 +20,12 @@ function form_hidden($name, $value)
  *
  * @param string $name
  * @param string $label
- * @param string $value
+ * @param float  $value
  * @return string
  */
-function form_spinner($name, $label, $value)
+function form_spinner(string $name, string $label, float $value)
 {
     $id = 'spinner-' . $name;
-    $value = htmlspecialchars((string)$value);
 
     return form_element($label, '
         <div class="input-group">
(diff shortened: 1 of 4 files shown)


includes/sys_form.php
```diff
@@ -20,34 +20,32 @@ function form_hidden($name, $value)
  *
  * @param string $name
  * @param string $label
- * @param string $value
+ * @param float  $value
```
weeman1337, line 23: You use a `number` field with a stepping of 1 here. So we should use `int` instead of `float`.

```suggestion
 * @param int  $value
```

resources/assets/themes/base.scss
```diff
@@ -372,3 +372,14 @@ code {
   max-width: 75%;
   display: inline-block;
 }
+
+/* Chrome, Safari, Edge, Opera */
```
weeman1337, line 376: Can you add a quick note about what this is doing?

weeman1337 (changes requested):
Thanks for your contribution :+1: 

Can you solve the conflicts and then re-request review?

resources/assets/themes/base.scss
ximex: added

includes/sys_form.php
ximex: changed. thought about if we use this function somewhere else where the step size is a float that this is already prepared. but then we also need to change the function signature to provide the step size

PR #1020, opened 2022-12-06
Secure tokens

ximex:
`md5()` isn't secure any more. replaced it with `random_bytes()`
32 chars is a bit less for a api key. increased it to 64 chars

`password_resets`->`token` is type `text` in database. should we change that to `varchar(32)`?
when do i have to add a migration in the `HasDatabase.php` migration list?

### DIFF
--- a/db/factories/User/PasswordResetFactory.php
+++ b/db/factories/User/PasswordResetFactory.php
@@ -16,7 +16,7 @@ class PasswordResetFactory extends Factory
     public function definition()
     {
         return [
-            'token' => md5($this->faker->unique()->password()),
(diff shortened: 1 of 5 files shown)


ximex:
@MyIgel 
> password_resets->token is type text in database. should we change that to varchar(32)?
when do i have to add a migration in the HasDatabase.php migration list?

MyIgel:
@ximex we could use some varchar here but i don't think that it makes such a big difference (as its mostly never used). Feel free if you want to ^^

The `HasDatabase.php` is used to skip migrtations that only change legacy tables and thus can be skipped when migrating the in-memory-sqlite db.

PR #1021, opened 2022-12-08
fix tag nesting

ximex:
changed
`<a href...><b>...</a>:</b>`
to
`<b><a href...>...</a>:</b>`


MyIgel:
That PR would be better fitted with more smaller fixes as it's kind of not worth the review? :D

PR #1023, opened 2022-12-09
Route parameter improvements

ximex:


weeman1337 (changes requested):
@ximex thanks for your contribution :+1: 

Some testes require an update to match the new code. You can run them locally by calling `./vendor/bin/phpunit` or if you use docker `docker-compose exec  es_workspace vendor/bin/phpunit` from the `docker/dev` directory.

Re-request review if the tests are fixed.

Looks good otherwise.

weeman1337 (approved):
Looks good now :+1:

PR #1024, opened 2022-12-10
code improvements

ximex:
unify lowercase variant of form method post
remove duplicate class selector
improve comments


weeman1337:
Blocked by #1031

ximex:
sry for the regression. but why is this blocking this PR?

weeman1337:
This one needs testing. And that does make less fun with a broken main.

PR #1025, opened 2022-12-10
change form submit button text to "save"

ximex:
changed the submit buttion in the user settings pages from `Absenden` to `Speichern`


weeman1337 (approved):
LGTM :+1:

PR #1026, opened 2022-12-10
use check and cross character instead of icon in tables

ximex:


weeman1337 (commented):
Thanks for your contribution @ximex 

To be honest I am not sure if we should change these places to use chars instead of the icons. Using icons makes the look all over the place more consistent.

Before

icon_before

After

icon_after

What is your opinion on this @MyIgel ?

MyIgel:
@weeman1337, that was actutally also my first thought about that. Using characters here makes it kind of inconsistent.

weeman1337:
Going to close this then.

Sorry @ximex :hugs:  If you have a completely different opinion on this feel free to leave a comment here.

ximex:
The idea is following: the real characters improve the accessibility. the bootstrap icons are only a visual thing but not "visible" in the DOM. also if you mark/select and copy the table and paste it somewhere else you don't have anything for this cells. this is why i only did it for the tables. as for the buttons you still have the text next to the icon

weeman1337:
Hat least for accessibility we can use ARIA features.

PR #1027, opened 2022-12-10
Shifts filter

ximex:
improve shifts filter html structure
replace inline javascript with event handler for set-date and set-time

### DIFF
--- a/resources/assets/js/forms.js
+++ b/resources/assets/js/forms.js
@@ -17,12 +17,12 @@ global.checkAll = (id, checked) => {
 /**
  * Sets the checkboxes according to the given type
  *
- * @param {string} id The elements ID
+ * @param {string} id The Id of the element containing all the checkboxes
  * @param {int[]} shiftsList A list of numbers
  */
 global.checkOwnTypes = (id, shiftsList) => {
   document.querySelectorAll(`#${id} input[type="checkbox"]`).forEach((element) => {
-    const value = parseInt(element.value, 10);
+    const value = Number(element.value);
     element.checked = shiftsList.includes(value);
   });
 };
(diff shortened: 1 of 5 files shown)


resources/assets/js/forms.js
```diff
@@ -25,77 +17,99 @@ global.checkAll = (id, checked) => {
...
+    to.setHours(to.getHours() + hours);
+
+    if (to < from) {
+      setInput(to, from);
+    } else{
+      setInput(from, to);
+    }
+  };
+
+  document.querySelectorAll('.set-date').forEach((element) => {
```
weeman1337, line 106: If we already have the data attributes, why not using them for selection as well? Something like this? So that we don't need to use the CSS class as functional marker.

```suggestion
  document.querySelectorAll('[data-set-days]').forEach((element) => {
```

resources/assets/js/forms.js
```diff
@@ -25,77 +17,99 @@ global.checkAll = (id, checked) => {
 /**
  * Sets the checkboxes according to the given type
  *
- * @param {string} id The elements ID
+ * @param {string} Id of the element containing all the checkboxes
```
weeman1337, line 20:
```suggestion
 * @param {string} id of the element containing all the checkboxes
```

weeman1337 (commented):
Thank you for your contribution @ximex :+1: 

Tested the changes, looks good. Some minor comments in the code.

resources/assets/js/forms.js
ximex: The plan was that the class choose what to set (time/date) and the data attribute the different units. so you could also easy use `data-minutes` etc in the future. should i still change it?

resources/assets/js/forms.js
ximex: fixed

ximex:
fixed

PR #1028, opened 2022-12-10
Sys functions improvements

ximex:
i hope i found all wrong types

### DIFF
--- a/includes/pages/admin_free.php
+++ b/includes/pages/admin_free.php
@@ -112,7 +112,7 @@ function admin_free()
             div('row', [
                 div('col-md-12 form-inline', [
                     div('row', [
-                        form_text('search', __('Search'), $search, null, null, null, 'col'),
+                        form_text('search', __('Search'), $search, false, null, null, 'col'),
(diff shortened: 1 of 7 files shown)


MyIgel:
I am somehow pretty sure that its a bad idea to use `declare(strict_types=1);` in any legacy code from the `includes` dir.
Same goes for introducing types there -> just letting the old code sit there and don't touching it is normally a wise choice (and replacing it with newly written implementations).

ximex:
I don't know anything against it. also if we know that the includes code would get completely replaced in the future. And the next usage is far in the future if we overlooked a case that would cause a crash. so much testing time

MyIgel:
The point here is that the old code relies on interesting type interpretations and changes which strict types could eventually break.
 I highly personally prefer to keep the changes inside that dir to a bare minimum and that includes not adding that many types and definitely not adding strict typing  there.

weeman1337:
My opinion on this: Don't mess too much around with `includes`.

msquare:
Sorry, we agreed to not mess with the legacy code too much.

PR #1029, opened 2022-12-10
improve countdown logic

ximex:
use `Intl.RelativeTimeFormat` to support different l10n
add week as possible duration

### DIFF
--- a/resources/assets/js/countdown.js
+++ b/resources/assets/js/countdown.js
@@ -38,10 +38,10 @@ ready(() => {
 
   document.querySelectorAll('[data-countdown-ts]').forEach((element) => {
     const timestamp = Number(element.dataset.countdownTs);
-    const template = element.innerHTML;
-    element.innerHTML = template.replace('%c', formatFromNow(timestamp));
+    const template = element.textContent;
+    element.textContent = template.replace('%c', formatFromNow(timestamp));
     setInterval(() => {
-      element.innerHTML = template.replace('%c', formatFromNow(timestamp));
+      element.textContent = template.replace('%c', formatFromNow(timestamp));
     }, 1000);
   });
 });
--- a/resources/assets/js/countdown.js
+++ b/resources/assets/js/countdown.js
@@ -1,85 +1,46 @@
 import { ready } from './ready';
 
-const lang = document.documentElement.getAttribute('lang');
-
-const templateFuture = 'in %value %unit';
-const templatePast = lang === 'en'
-  ? '%value %unit ago'
-  : 'vor %value %unit';
-
-const yearUnits = lang === 'en'
-  ? ['year', 'years']
-  : ['Jahr', 'Jahren'];
-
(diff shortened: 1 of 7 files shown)


resources/assets/js/countdown.js
```diff
@@ -1,81 +1,46 @@
...
+
+    return rtf.format(0, 'second');
+  }
+
   document.querySelectorAll('[data-countdown-ts]').forEach((element) => {
-    const timestamp = element.dataset.countdownTs;
-    const template = element.innerHTML;
-    element.innerHTML = template.replace('%c', formatFromNow(timestamp));
+    const timestamp = Number(element.dataset.countdownTs);
+    element.innerHTML = formatFromNow(timestamp);
```
weeman1337, line 41: Can you please roll back to keep the rest of the text and only replace `%c`? This works as kind of a template variable in the template.

weeman1337 (changes requested):
Thank you for your contribution @ximex 

Can you check my comment?

weeman1337:
@MyIgel changes look good from my side though they change the countdown display slightly. What do you think?

Before

cd_before

After

cd_after

ximex:
if the `übermorgen` etc naming is not wanted it's easy to remove by removing this option: `{ numeric: 'auto' }`. We only need to create another object with this option for `return rtf.format(0, 'second');` to get here `jetzt`.
The shift of one unit (minute, hours, etc): could check to change that the shown value is always smaller than the real time (real 3min20sec -> shown 3min or 4min)

resources/assets/js/countdown.js
ximex: what do you mean? the template variable is completely removed. the complete text is generated by the browser API

MyIgel:
I think having something like "tomorrow" is fine here but eventually we should also add a `title` attribute with the real datetime value

resources/assets/js/countdown.js
weeman1337: Look for example at the design page https://github.com/engelsystem/engelsystem/blob/main/resources/views/pages/design.twig#L325-L338 There can be more text in the element. Just `%c` should be replaced with the countdown.

weeman1337:
Picking up MyIgels comment: Can you add a title? I would recommend the full text „in 2 Tagen 23 Stunden und 42 Minuten“ (just an example; something like that; longest description you can get from the Browser API).

And if we are setting the title we can also add the date + time.

```
Nächste Schicht in 2 Tagen, 23 Stunden und 42 Minuten
13.12.2022 18:23
```

resources/assets/js/countdown.js
ximex: now i got it. fixed

ximex:
@weeman1337 thats not possible with this API

MyIgel:
In that case just having a date and  time in the title attribute is enough imho

ximex:
@MyIgel that way?

PR #1035, opened 2022-12-10
set autocomplete attribute for password inputs

ximex:


weeman1337 (approved):
Looks good, thanks :+1:

PR #1037, opened 2022-12-11
Improve migration

ximex:


db/migrations/2021_05_23_000000_set_admin_password.php
```diff
@@ -43,4 +45,11 @@ public function up()
 
         $this->auth->setPassword($admin, $setupPassword);
     }
+
+    /**
+     * Reverse the migration
+     */
+    public function down(): void
+    {
```
ximex, line 53: is here some migration needed?

MyIgel: nope there isn't as it is only used to set the admin password on initial install as a shortcut for automated installs (for example when deploying to k8s)

src/Database/Migration/IMigration.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Engelsystem\Database\Migration;
+
+interface IMigration {
```
MyIgel, line 7: At the moment i don't see the usecase for an additional interface here?

db/migrations/2021_05_23_000000_set_admin_password.php
MyIgel: And also adding useless method stubs is kind of pointless?

ximex:
I added the interface as one time the down method was missing. and i added it as on some other migrations there is also an empty down method

MyIgel:
-> Then its better to just remove the empty methods i guess? ^^

ximex:
Could also do that. Thought the other way is correct/wanted

ximex:
that way ok?

MyIgel:
Hi @ximex, this branch should be rebased on the current main branch as the other (broader) cleanup already fixed some of the changes.

ximex:
done

MyIgel (changes requested):
It would be great to have two commits (based on a rebase on main): one that formats the files (and which id can be added to the [.git-blame-ignore-revs](https://github.com/engelsystem/engelsystem/blob/main/.git-blame-ignore-revs) and one that adds the strict types and private functions.

I would do that by myself but am not allowed to (would be nice if you could check the "maintainers are allowed to change the PR code" checkbox next time :) )

weeman1337:
> It would be great to have two commits (based on a rebase on main): one that formats the files 

I would then open a PR next week and „Further improve migration“ with another formatting :upside_down_face: Can we spec the code style changes in PHPCS, so that this does not happen?

ximex:
waiting until this PR gets merged: https://github.com/engelsystem/engelsystem/pull/1055

MyIgel:
@ximex the PR is merged :)

ximex:
rebased the PR. what changes should i do now?

db/migrations/2020_04_07_000000_change_mysql_database_encoding_to_utf8mb4.php
```diff
@@ -27,12 +27,4 @@ public function up(): void
             );
         }
     }
-
-    /**
-     * Reverse the migration
-     */
-    public function down(): void
-    {
-        // As utf8mb4 is a superset of utf8, there is nothing to do here
```
MyIgel, line 36: Imho that comment is relevant and should stay even after removing the method.

MyIgel (changes requested):
Besides one small change this PR imho can be squashed and merged

db/migrations/2020_04_07_000000_change_mysql_database_encoding_to_utf8mb4.php
ximex: changed

PR #1042, opened 2022-12-11
PHP 8.2 compatibility, bump min. version to 8.1

### LINKED ISSUE #1022: Drop PHP 7.4 Support
PHP 7.4 isn't supported any more: https://www.php.net/supported-versions
Instead we should check if PHP 8.2 (today released) works

MyIgel:
Make the Engelsystem PHP 8.2 compatible and updates packages

Closes #1022


MyIgel:
At the moment it requires https://github.com/FakerPHP/Faker/pull/570 to be merged and a new package version released for the CI to run without warnings which get converted to errors.

ximex:
why not allow 8.0? It's still supported
also update node?
`FROM node:14-alpine as themes` -> `FROM node:18-alpine as themes`

MyIgel:
Updating node is way out of scope here..
We could move the 8.1 update to a later date but it would be useful to already have support for some 8.1 features like new in initializers and intersection types but that can be discussed i guess. There might also be some package dependencies which i had to check then again.

MyIgel:
(and also the current stable / lts versions of debian and ubuntu package 8.1 so thats no problem either)

MyIgel:
Faker got updated so this can be tested and merged :tada:

weeman1337 (approved):

PR #1043, opened 2022-12-13
Add UserAngelTypes model

MyIgel:
This adds the UserAngelType Model and its refactorings.
Hopefully everything still works as expected _\*keeps fingers crossed_.

I can also move the OAuth2 reformatting to its own PR but it seemed fitting here.


weeman1337 (approved):
Go

PR #1044, opened 2022-12-13
Ignore the timezone configured in php.ini, better cleanup after feature tests

MyIgel:
* Feature tests currently reset it to the configured one which is unexpected
* Servers can be misconfigured / have default configs which we don't want


MyIgel:
@weeman1337 :wave:

weeman1337 (approved):

PR #1045, opened 2022-12-15
Use native types where possible + ctor prop promotion 💣

weeman1337:
- Native prop types
- Native return types
- Native param types
- Drop useless comments
- Ctor prop promotion

Also contains a new `.git-blame-ignore-revs` that can be used to ignore this changes for `git blame`.

[pipeline](https://chaos.expert/engelsystem/engelsystem/-/pipelines/36942)


ximex:
I'm not a fan of the new syntax where you could remove the class property definitions if they got directly defined in the constructor.
what is with all the `// phpcs:ignore`? why not set the type?

weeman1337:
> I'm not a fan of the new syntax where you could remove the class property definitions if they got directly defined in the constructor. 

Oh you mean it is great to have to write all the boilerplate

```php
private string $something;

public function __construct(string $somthing)
{
    $this->something = $something;
}
```

instead of just

```php
public function __construct(private string $something) {}
```

:wink: 

> what is with all the `// phpcs:ignore`? why not set the type?

Params that are not typed in the base classes of a lib (yet).

ximex:
i know it's much boilerplate code. i also like it if there is a way with less typing. In this case i like it more to have all properties (except that one from parent classes) in one place to see what is there. just my 2 cent

src/Controllers/Admin/NewsController.php
```diff
@@ -15,52 +15,23 @@ class NewsController extends BaseController
...
+        protected LoggerInterface $log,
+        protected News $news,
         Redirector $redirector,
-        Response $response
+        protected Response $response
     ) {
-        $this->auth = $auth;
-        $this->log = $log;
-        $this->news = $news;
         $this->redirect = $redirector;
```
MyIgel, line 32: Also here (and following)

MyIgel (changes requested):

src/Helpers/Schedule/Conference.php
```diff
@@ -9,106 +9,58 @@ class Conference
...
-        ?string $timeslotDuration = null,
-        ?string $baseUrl = null
+        protected ?string $start = null,
+        protected ?string $end = null,
+        protected ?int $days = null,
+        protected ?string $timeslotDuration = null,
+        protected ?string $baseUrl = null
     ) {
         $this->title = $title;
         $this->acronym = $acronym;
```
MyIgel, line 31: Thats the same but why not replaced?

src/Helpers/Schedule/XmlParser.php
```diff
@@ -82,7 +76,6 @@ protected function parseXml(): void
 
     /**
      * @param SimpleXMLElement[] $eventElements
-     * @param Room               $room
      * @return array
```
MyIgel, line 79: Is also redundant

MyIgel (changes requested):

src/Controllers/NewsController.php
```diff
@@ -216,9 +154,7 @@ protected function showOverview(bool $onlyMeetings = false): Response
     }
 
     /**
-     * @param string $page
      * @param array $data
```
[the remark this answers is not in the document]
weeman1337: Commented on old commit

Issue #1041, opened 2022-12-11
Shift-creation improvements


ximex:
- [x] Default Radio-Button Selection (https://github.com/engelsystem/engelsystem/pull/1033)
- [x] Disable Input fields if the corresponding radio-button is disabled (https://github.com/engelsystem/engelsystem/pull/1046)
- [ ] Show the room default-angle-counts

MyIgel:
The "Use room angeltypes" setting is normally only used on shifts which get imported automatically from a schedule and normally not when creating normal shifts as their required angeltypes normally differ too much.
Also on bigger events you normally create more "multiple shifts" than single tasks (at least that are my previous experiences).

Graying out input fields sounds like a good idea indeed and showing the room angel counts (only if they are not 0) might be an improvement worth trying.

PR #1046, opened 2022-12-16
enable/disable all html-tags corresponding to a radio button checked state

ximex:
fixes parts of https://github.com/engelsystem/engelsystem/issues/1041

missing:
set the `data-radio-name="xxx"` and `data-radio-value="yyy"` attributes on the html tags that should update its disabled state
possible html tags: 'button', 'command', 'fieldset', 'input', 'keygen', 'optgroup', 'option', 'select', 'textarea'

already tested on the shift-creation page with adding the tags for testing only

### DIFF
--- a/includes/pages/admin_shifts.php
+++ b/includes/pages/admin_shifts.php
@@ -422,7 +422,11 @@ function admin_shifts()
             . form_spinner(
                 'angeltype_count_' . $type->id,
                 $type->name,
-                $needed_angel_types[$type->id]
+                $needed_angel_types[$type->id],
+                [
+                    'radio-name'  => 'angelmode',
+                    'radio-value' => 'manually'
+                ]
             )
             . '</div>';
     }
@@ -465,7 +469,15 @@ function admin_shifts()
                             __('Length'),
                             $request->has('length')
                                 ? $request->input('length')
-                                : '120'
+                                : '120',
+                            false,
+                            null,
+                            null,
+                            '',
+                            [
+                                'radio-name'  => 'mode',
+                                'radio-value' => 'multi'
+                            ]
                         ),
                         form_radio(
                             'mode',
@@ -478,7 +490,15 @@ function admin_shifts()
                             __('Shift change hours'),
                             $request->has('change_hours')
                                 ? $request->input('change_hours')
-                                : '00, 04, 08, 10, 12, 14, 16, 18, 20, 22'
+                                : '00, 04, 08, 10, 12, 14, 16, 18, 20, 22',
+                            false,
+                            null,
+                            null,
+                            '',
+                            [
+                                'radio-name'  => 'mode',
+                                'radio-value' => 'variable'
+                            ]
                         ),
                         form_checkbox(
                             'shift_over_midnight',
--- a/includes/sys_form.php
+++ b/includes/sys_form.php
@@ -21,19 +21,24 @@ function form_hidden($name, $value)
  * @param string $name
  * @param string $label
  * @param int    $value
+ * @param array  $data_attributes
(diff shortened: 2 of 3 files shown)


ximex:
the shift-creation page is the only one that needs this. added the data attributes for the 2 radio button groups

includes/pages/admin_shifts.php
```diff
@@ -422,7 +422,8 @@ function admin_shifts()
             . form_spinner(
                 'angeltype_count_' . $type->id,
                 $type->name,
-                $needed_angel_types[$type->id]
+                $needed_angel_types[$type->id],
+                ['data-radio-name="angelmode"', 'data-radio-value="manually"']
```
frischler, line 426: I would prefer to have an object here instead of the list, similar to how the "opt" parameter is used in all the macros in form.twig.
So, it would look like this:
{'data-radio-name': 'angelmode', 'data-radio-value': 'manually'}

ximex: changed

MyIgel: Eventually (as this are only data attributes) explicitly save define them as that? So that its more `['radio-name' => 'angelmode']`

ximex: good idea. changed

MyIgel:
Looks good so far to me, @weeman1337 do you have any additional requests?

resources/assets/js/forms.js
```diff
@@ -111,6 +111,35 @@ ready(() => {
   });
 });
 
+/**
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled
```
weeman1337, line 115:
```suggestion
 * {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled#overview}
```

resources/assets/js/forms.js
```diff
@@ -111,6 +111,35 @@ ready(() => {
...
+ */
+const DISABLE_ELEMENTS = [
+  'button', 'command', 'fieldset', 'input', 'keygen', 'optgroup', 'option', 'select', 'textarea'
+];
+ready(() => {
+  document.querySelectorAll('input[type="radio"]').forEach((radioElement) => {
```
weeman1337, line 121: Can you please add some description about what this entire block does? It is really hard to follow…

includes/sys_form.php
```diff
@@ -21,19 +21,24 @@ function form_hidden($name, $value)
  * @param string $name
  * @param string $label
  * @param int    $value
+ * @param array  $data_attributes
```
weeman1337, line 24: Even we are in `includes` this is new code. Can you specify the array contents here please?
https://phpstan.org/writing-php-code/phpdoc-types

weeman1337 (changes requested):
Can you have a look at the comments @ximex . Still needs a round of testing in the application by me.

resources/assets/js/forms.js
ximex: changed

resources/assets/js/forms.js
ximex: added more comments

includes/sys_form.php
ximex: what do you mean? instead of `array` use `string[]` or `array<string, string>`?

ximex:
also positiv with this solution: if you have disabled javascript it still works

includes/sys_form.php
MyIgel: More the later one i think ^^

weeman1337 (approved):

PR #1053, opened 2022-12-23
Replace select 2 with choices.js and kill jQuery

### LINKED ISSUE #1001: Replace `jquery` with plain javascript
Replace `jquery` code with plain javascript for custom javascript code
`bootstrap` doesn't required jquery any more
`select2` still requires jquery. maybe we could also replace this to drop jquery completely

weeman1337:
closes #1001

### DIFF
--- a/resources/assets/js/forms.js
+++ b/resources/assets/js/forms.js
@@ -1,4 +1,4 @@
-import 'select2';
+import Choices from 'choices.js';
 import { formatDay, formatTime } from './date';
 import { ready } from './ready';
 
@@ -37,6 +37,22 @@ ready(() => {
   };
 
   /**
+   * Sets a select value and triggers a change.
+   * If the select has a Choices.js instances, it uses this instead to set the value.
+   *
+   * @param {HTMLSelectElement} element
+   * @param {*} value
+   */
+  const setSelectValue = (element, value) => {
+    if (element.choices) {
+      element.choices.setChoiceByValue(value);
+    }
+
+    element.value = value;
+    triggerChange(element);
+  };
+
+  /**
    * Sets the values of the input fields with the IDs to from/to:
    * - date portion of from → start_day
    * - time portion of from → start_time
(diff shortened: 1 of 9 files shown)


MyIgel (changes requested):
Looks like only the first select of a page gets used? On the shifts page only the from date is selectable, the to is the old select. Same for the room select on the shifts creation page
image

On the messages recepient select the text is too long:
Screenshot_20221224_212418

At least the Engelsystem dark theme needs some contrast improvements:
Screenshot_20221224_212554

ximex:
see: https://github.com/Choices-js/Choices#setup
> Note: If you pass a selector which targets multiple elements, the first matching element will be used. Versions prior to 8.x.x would return multiple Choices instances.

weeman1337:
> Looks like only the first select of a page gets used? 

Done

> On the messages recepient select the text is too long:

Done

resources/assets/js/forms.js
```diff
@@ -200,10 +200,15 @@ ready(() => {
 });
 
 ready(() => {
-  $('select').select2({
-    theme: 'bootstrap-5',
-    width: '100%',
-  });
+  document.querySelectorAll('select').forEach(
+    (element) =>
```
ximex, line 204: i would put the `(element) =>` into the previous line

weeman1337: Code style in JS/CSS is handled by [prettier](https://prettier.io/).

weeman1337:
> At least the Engelsystem dark theme needs some contrast improvements:

Dark theme contract is bad anyway :stuck_out_tongue: 

image

ximex:
this PR should get rebased on main to have this commit: https://github.com/engelsystem/engelsystem/commit/30f50dab6d758977b3a2e763a49f511b60986fd9
this is related to the today/yesterday/tomorrow buttons

MyIgel:
@ximex That literally doesn't help in any way with any of the above problems so yes it could be rebased but that's literally useless.

weeman1337:
> Directly using full white might be questionable but works at least.

select_after

tabs

I do only use the same colours for selected elements, like for instance tabs. If you want other colours here, please send me them and I will change them.

> More problematic is the point that the "Today" / "Yesterday" / "Tomorrow" buttons stopped working.

Done

> Also (at least in my setup with many shifts) there seems to be some weird caching issue or so when searching -> we should have a look at this too.

Can you explain the issue?

MyIgel:
Looks good so far, the issue is that the select seems to only show the first four results when searching. When i search for 2022 i expect all 2022 dates to show up but only get the 1st January until the 4th January and not more.

weeman1337:
> only show the first four results when searching. 

done

Design now provides a 50 options list for testing.

weeman1337:
:ping_pong: @MyIgel can you re-test the search?

PR #1054, opened 2022-12-23
Fix passing wrong type to date_default_timezone_set

weeman1337:
Raised an error when setting up a new environment.
Expects a `string`; gets a `CarbonTimeZone`.


weeman1337:
Not sure why this happened…

MyIgel:
So this one is irelevant?

weeman1337:
Yes. Was my weird setup.

PR #1056, opened 2022-12-25
Enable space after cast sniff

weeman1337:
Stay close to [the PSR-12 example](https://www.php-fig.org/psr/psr-12/#61-unary-operators).

[CI](https://chaos.expert/engelsystem/engelsystem/-/pipelines/37085)

Removes `.git-blame-ignore-revs`. Great experiment. But already outdated and hard to keep up to date.


MyIgel:
I would not say that it's aoutdated but it needs some additional maintenance, thats right..

PR #1058, opened 2022-12-29
replace sticky-headers javascript impl with css

ximex:
use css instead of js for sticky headers in shift table. prevents flickering on scrolling.
add background color for first column heading to prevent overlapping text


MyIgel (changes requested):
This only makes the top bar sticky but ignores the times at the side which need to be sticky when scrolling sidewise (happens when too much shifts (~6) are in parallel)

ximex:
fixed

MyIgel:
Hmm at some point the bar sits below the menu and sometimes it still scrolls outside the viewport:

Screenshot_20230122_204305

ximex:
rebased
@MyIgel pls test again. with the merged PR of the navbar this problem should be gone

PR #1059, opened 2022-12-29
improve disabling buttons on form submit

ximex:
use `disabled` attribute instead of `readonly` attribute and `disabled` class

### DIFF
--- a/resources/assets/js/forms.js
+++ b/resources/assets/js/forms.js
@@ -100,14 +100,14 @@ global.setHours = (hours) => {
 ready(() => {
   /**
    * Disable every submit button after clicking (to prevent double-clicking)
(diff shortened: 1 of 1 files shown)


weeman1337:
Thank you for your contribution @ximex .

At least creating shift types doesn't work with the changes:

- Go to admin → shift types
- Try to create a shift type
- Page reloads but doesn't do anything

weeman1337 (changes requested):
Thank you for your contribution @ximex .

At least creating shift types doesn't work with the changes:

- Go to admin → shift types
- Try to create a shift type
- Page reloads but doesn't do anything

ximex:
found the problem. didn't know that disabled prevents the input element to send it's value. have to think about it. maybe i close this PR

PR #1060, opened 2022-12-29
invert input time icon color for chrome based browsers in dark themes

ximex:
before:
image
after:
image

should we also hide the date-picker icon?
```css
input[type='time']::-webkit-calendar-picker-indicator,
input[type='datetime-local']::-webkit-calendar-picker-indicator {
  display: none;
  -webkit-appearance: none;
}
```

### DIFF
--- a/resources/assets/themes/dark.scss
+++ b/resources/assets/themes/dark.scss
@@ -6,3 +6,10 @@ $secondary: #222;
 $table-striped-bg: rgba(#fff, 0.05);
 
 $es-choices-highlight-color: #000;
+
+$invert-color-value: 1 !default;
+
+/* Invert the clock icon color for Chrome */
+input[type='time']::-webkit-calendar-picker-indicator {
+  filter: invert($invert-color-value);
+}
--- a/resources/views/pages/user-shifts.html
+++ b/resources/views/pages/user-shifts.html
@@ -13,7 +13,7 @@
                                 pattern="^\d{1,2}:\d{2}$" placeholder="HH:MM" maxlength="5" value="%start_time%"
                             >
                             <button class="btn btn-secondary" title="Now" type="button">
-                                <span class="bi bi-clock"></span>
+                                <span class="bi bi-arrow-counterclockwise"></span>
                             </button>
                         </div>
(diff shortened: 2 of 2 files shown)


weeman1337:
Thanks for the screenshots :+1: 

Isn't the time picker icon not was chrome-based browser users expect to see? @MyIgel any opinion on that?

ximex:
the problem is also that there is a bug that it's not possible to change the icon color

weeman1337:
Need to do some research on that…

MyIgel:
For the sake of a consistent look i would also prevent hiding the chrome icon so that it can be styled independently.

weeman1337 (changes requested):
If we remove the icon, we loose the time picker function of the browser.

It is possible to invert the icon for dark modes:

image

```
input[type='time']::-webkit-calendar-picker-indicator {
     filter: invert(1);
}
```

We should then use [this icon](https://icons.getbootstrap.com/icons/arrow-counterclockwise/) for the „reset to now“ button.

What do you think? Can you apply these changes?

ximex:
`invert(1)` is a bit to much white. maybe `0.5`?

weeman1337:
It should have the same colour as the button. I would suggest adding this to the dark style file and also use a variable with the option for a default value for it. Then we can tweak it for specific themes, if required.

resources/assets/themes/dark.scss
```diff
@@ -6,3 +6,8 @@ $secondary: #222;
 $table-striped-bg: rgba(#fff, 0.05);
 
 $es-choices-highlight-color: #000;
+
+/* Hide the clock icon for Chrome */
```
MyIgel, line 10: It does not hide it but inverts it.
Also the reset button should be changed as weeman mentioned.

MyIgel (changes requested):

ximex:
which reset button do you mean? the "now" button?

MyIgel:
@ximex yes that one -> when having two identical icons its confusing esp. when they do different  things.

ximex:
done

PR #1062, opened 2022-12-31
Improve alerts and conversation

ximex:
- add `role="alert"` attribute for alerts
- improve conversation styling
  - remove vertical space between messages if there are about <4 messages
  - remove `message` class
  - space between messages and message input field
  - improve style for different window sizes
- use cards instead of alerts for messages


weeman1337:
Thank you for your contribution @ximex :+1: 

For all PRs: Can you please always provide a description / motivation of what you have done there? This would make reviews much easier.

PR #1064, opened 2022-12-31
Shifts

MyIgel:
Created a migration to migrate the `Shifts` table to a new `shifts` table, added the `Shifts` model and refactored the code to use it.
Some parts might have fallen off and shrapnel can fly in any direction here i guess?

:exclamation: The commits must be squashed on merge! :exclamation:

### DIFF
--- a/db/factories/Shifts/ShiftFactory.php
+++ b/db/factories/Shifts/ShiftFactory.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Database\Factories\Engelsystem\Models\Shifts;
+
+use Engelsystem\Models\Room;
+use Engelsystem\Models\Shifts\Shift;
+use Engelsystem\Models\Shifts\ShiftType;
+use Engelsystem\Models\User\User;
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+class ShiftFactory extends Factory
+{
+    /** @var string */
+    protected $model = Shift::class; // phpcs:ignore
+
+    public function definition(): array
+    {
+        $start = $this->faker->dateTimeThisMonth('2 weeks');
+        return [
+            'title'          => $this->faker->unique()->text(15),
+            'description'    => $this->faker->text(),
+            'url'            => $this->faker->url(),
+            'start'          => $start,
+            'end'            => $this->faker->dateTimeInInterval($start, '+3 hours'),
+            'shift_type_id'  => ShiftType::factory(),
+            'room_id'        => Room::factory(),
+            'transaction_id' => $this->faker->optional()->uuid(),
+            'created_by'     => User::factory(),
+            'updated_by'     => $this->faker->optional(.3)->boolean() ? User::factory() : null,
+        ];
+    }
+}
--- a/includes/controller/shifts_controller.php
+++ b/includes/controller/shifts_controller.php
@@ -1,44 +1,46 @@
 <?php
 
+use Carbon\CarbonTimeZone;
 use Engelsystem\Helpers\Carbon;
 use Engelsystem\Http\Exceptions\HttpForbidden;
 use Engelsystem\Models\AngelType;
-use Engelsystem\Models\Room;
 use Engelsystem\Models\Shifts\ScheduleShift;
+use Engelsystem\Models\Shifts\Shift;
 use Engelsystem\Models\Shifts\ShiftType;
 use Engelsystem\Models\User\User;
 use Engelsystem\ShiftSignupState;
+use Illuminate\Support\Collection;
 
 /**
- * @param array $shift
+ * @param array|Shift $shift
  * @return string
  */
 function shift_link($shift)
 {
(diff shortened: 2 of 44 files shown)


db/factories/Shifts/ShiftFactory.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
weeman1337, line 1: I would say all new files should `declare(strict_types=1);`.

includes/controller/shifts_controller.php
```diff
@@ -149,14 +151,22 @@ function shift_edit_controller()
...
-            Shift_update($shift);
+            $shift->shift_type_id = $shifttype_id;
+            $shift->title = $title;
+            $shift->description = $description;
+            $shift->room_id = $rid;
+            $shift->start = $start;
+            $shift->end = $end;
+            $shift->updatedBy()->associate(auth()->user());
+
+            mail_shift_change(Shift($shift->id), $shift);
```
weeman1337, line 162: Shouldn't this not better go after `save()`. Otherwise, if save fails, there will be an e-mail.

includes/controller/shifts_controller.php
```diff
@@ -390,19 +398,74 @@ function shifts_json_export_controller()
...
+            // Talk URL
+            'URL' => $shift->url,
+
+            // Room name
+            'Name' => $shift->room->name,
+            // Location map url
+            'map_url' => $shift->room->map_url,
+
+            // Start timestamp
+            /** @deprecated  */
```
weeman1337, line 432: If this is deprecated, what should be used instead?
Can we add a comment?

includes/pages/admin_shifts.php
```diff
@@ -199,17 +200,17 @@ function admin_shifts()
...
+                    'title'         => $title,
+                    'shift_type_id' => $shifttype_id,
+                    'description'   => $description,
                 ];
             } elseif ($mode == 'multi') {
-                $shift_start = (int) $start;
+                $shift_start =  $start;
                 do {
-                    $shift_end = $shift_start + (int) $length * 60;
+                    $shift_end = (clone $shift_start)->addSeconds((int) $length * 60);
```
weeman1337, line 213: `CarbonImmutable` 4tw :hammer:

src/Controllers/Metrics/Stats.php
```diff
@@ -309,13 +310,9 @@ public function rooms(): int
...
     }
 
-    /**
-     * @codeCoverageIgnore
-     */
     public function shifts(): int
     {
-        return $this
-            ->getQuery('Shifts')
+        return Shift::query()
```
weeman1337, line 315: This can be `return Shift::count();`

src/Models/Shifts/Shift.php
```diff
@@ -0,0 +1,108 @@
...
+ * @property string                      $title
+ * @property string                      $description
+ * @property string                      $url
+ * @property Carbon                      $start
+ * @property Carbon                      $end
+ * @property int                         $shift_type_id
+ * @property int                         $room_id
+ * @property string                      $transaction_id
+ * @property int                         $created_by
+ * @property int                         $updated_by
```
weeman1337, line 25: This can be `null`.

src/Models/Shifts/Shift.php
```diff
@@ -0,0 +1,108 @@
...
+ * @method static QueryBuilder|Shift[] whereUpdatedAt($value)
+ */
+class Shift extends BaseModel
+{
+    use HasFactory;
+
+    /** @var bool enable timestamps */
+    public $timestamps = true; // phpcs:ignore
+
+    /** @var array<string> */
```
weeman1337, line 56:
```suggestion
    /** @var array<string, string> */
```

weeman1337 (changes requested):
Some smaller comments and found two issues:

- Create a shift
- Delete single shift from the shift view

image

→ Explosion

```
 TypeError
Engelsystem\Events\Listener\Shift::deletedEntryCreateWorklog(): Argument #2 ($start) must be of type Engelsystem\Helpers\Carbon, Illuminate\Support\Carbon given, called in /var/www/vendor/illuminate/container/BoundMethod.php on line 36
```

---

- Edit a shift entry
- Clear start field
- Save

→

```
 Carbon \ Exceptions \ InvalidFormatException
Not enough data available to satisfy format
```

Looks good otherwise :+1:

includes/pages/admin_shifts.php
MyIgel: Something about no big changes ;)

db/factories/Shifts/ShiftFactory.php
MyIgel: Then we should create a check that enforces it ;)

includes/controller/shifts_controller.php
MyIgel: Fair point

PR #1068, opened 2023-01-17
update libs and move bootstrap-icons to dependencies

ximex:
there is no button to allow edits:
grafik


MyIgel:
Thats interesting, hmm.
Normally that can happen if you use a fork of a fork but that does not seem to be the case. Do you have some interesting security settings active? Or is you account somehow managed by an org?

ximex:
i think i found the problem. not possible from an organization: https://github.com/orgs/community/discussions/5634

ximex:
@MyIgel fixed

weeman1337 (approved):
Did a quick test. Looks good :+1:

PR #1069, opened 2023-01-20
NeededAngelTypes

MyIgel:
* Added NeededAngelTypes model
* Update the code to only save entries where count > 0 as no usecase was found where having 0 is needed


weeman1337 (approved):
:+1:

Issue #1016, opened 2022-12-04
Pre commit hook


weeman1337:
Provide a pre-commit hook to check the code style / run quick tests

- PHP based, e.g. https://github.com/captainhookphp/captainhook
- Python based, e.g. https://pre-commit.com/
- Plain and simple shell skript

ximex:
or this: https://www.npmjs.com/package/husky

PR #1071, opened 2023-01-22
Add pre-commit hook

weeman1337:
- Installs editorconfig-checker as npm package
- Includes editorconfig-checker in `yarn lint`
- Removes `check-editorconfig` CI step
- Finally adds a pre-commit :hook: 

Also removes note about `.git-blame-ignore-revs`. It was a nice experiment, but it didn't work.

closes https://github.com/engelsystem/engelsystem/issues/1016


package.json
```diff
@@ -9,9 +9,10 @@
     "build": "NODE_ENV=production yarn build:webpack",
     "build:webpack": "webpack",
     "build:watch": "webpack --watch",
-    "lint": "yarn lint:eslint && yarn lint:prettier",
+    "lint": "yarn lint:eslint && yarn lint:prettier && yarn lint:ec",
     "lint:eslint": "eslint .",
     "lint:prettier": "prettier --check resources/assets",
+    "lint:ec": "ec --exclude \"node_modules|vendor|public/assets|.git\"",
```
MyIgel, line 15: `public/coverage` /for phpunit coverage reporting) and eventually public `*.xml` files should be excluded too (to be able to import mock fahrplan data from localhost).
Other than that it looks good and can be merged :heart_eyes:

MyIgel (changes requested):

package.json
weeman1337: to be continued…

PR #1073, opened 2023-01-24
Add phpunit:coverage command

weeman1337:
- `phpunit` command
- `phpunit:coverage` command with Igel config™
- editorconfig-checker ignore for that

Pipeline run https://chaos.expert/engelsystem/engelsystem/-/pipelines/37524
Locally the translation test failed :man_shrugging:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,9 @@
             "composer phpcs:setup",
             "phpcbf -p"
         ],
-        "phpstan": "phpstan"
+        "phpstan": "phpstan",
+        "phpunit": "phpunit",
+        "phpunit:coverage": "phpunit --coverage-text --coverage-html ./public/coverage/"
     },
     "require": {
         "php": ">=8.1.0",
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -137,6 +137,15 @@ at least `phpdbg -qrr`(which has problems with switch case statements) as using 
 php -d pcov.enabled=1 -d pcov.directory=. vendor/bin/phpunit --testsuite Unit --coverage-text
 ```
 
+If using the Docker-based development environment
+you can run the following script to retrieve a coverage report.
+```sh
+docker compose exec es_workspace composer phpunit:coverage
+```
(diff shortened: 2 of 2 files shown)


MyIgel:
That variation of the editorconfig checker now also matches the `.idea` folder content (and thus also will probably a .code (or was it .vscode?) dir) which for example a run of `docker run -it --rm -v "$PWD":/check mstruebing/editorconfig-checker ec -v --exclude 'node_modules|vendor|public/(assets|coverage|.*\.xml)|.git'` will not do

weeman1337:
Then I would say switch to a white list? You never know what files will appear for someone :man_shrugging:

weeman1337:
:thinking: https://github.com/editorconfig-checker/editorconfig-checker/issues/99
Will check that in the afternoon

MyIgel:
I rebased it to include the better fix for the ec exclude ;)

composer.json
```diff
@@ -23,7 +23,9 @@
             "composer phpcs:setup",
             "phpcbf -p"
         ],
-        "phpstan": "phpstan"
+        "phpstan": "phpstan",
+        "phpunit": "phpunit",
+        "phpunit:coverage": "phpunit --coverage-text --coverage-html public/coverage"
```
MyIgel, line 28: At the moment that will only work if you have a local php installation with some existent coverage provider like `pcov` configured and enabled (like in the `.gitlab-ci.yml`) which is not the case in the workspace container at the moment.

weeman1337: :upside_down_face:  https://github.com/engelsystem/engelsystem/blob/main/docker/dev/Dockerfile#L5-L7

weeman1337:
Now also with documentation + [pcov is already enabled in the dev container](https://github.com/engelsystem/engelsystem/blob/main/docker/dev/Dockerfile#L5-L7)

composer.json
MyIgel: At least it was not *looksconfused

I mean: i explicitly tested it...

PR #1074, opened 2023-01-24
Code improvements

ximex:
i know it's includes but this should be straight forward changes and the types are already in the comments

### DIFF
--- a/resources/views/pages/design.twig
+++ b/resources/views/pages/design.twig
@@ -427,7 +427,7 @@
         {{ bar_chart | raw }}
     </div>
     </div>
-    <script>
+    <script type="text/javascript">
         document.getElementById('select-all-checkboxes').addEventListener('click', () => {
             checkAll('checkboxes', true);
         });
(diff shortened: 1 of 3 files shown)


MyIgel:
This PR changes code in the `includes` folder which should, in the best case, not be touched if not really needed and thus needs additional testing

resources/views/pages/design.twig
```diff
@@ -427,7 +427,7 @@
         {{ bar_chart | raw }}
     </div>
     </div>
-    <script>
+    <script type="text/javascript">
```
weeman1337, line 430: From [MDN recommendations](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type):

> Authors are encouraged to omit the attribute if the script refers to JavaScript code rather than specify a MIME type.

I would discard this change

weeman1337 (changes requested):
As :hedgehog:  already said. We should not mess around in `includes`. The `script` `type` attribute should also not be added. → close this PR @MyIgel ?

MyIgel:
Closing that PR is an option too, yes :+1:

weeman1337:
Sorry @ximex for discarding your work here :neutral_face: 

As concluded above, we don't want to change things under `/includes`.

PR #1075, opened 2023-01-24
Dockerfile

ximex:
uppercase the keywords in dockerfile
update `node:14-alpine` to `node:18-alpine` image

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,7 +11,7 @@ COPY resources/lang/ /data
 RUN find /data -type f -name '*.po' -exec sh -c 'file="{}"; msgfmt "${file%.*}.po" -o "${file%.*}.mo"' \;
 
 # Build the themes
-FROM node:14-alpine AS themes
+FROM node:20-alpine AS themes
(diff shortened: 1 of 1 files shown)


MyIgel:
I would personally wait 'till node 20.x which will get updates until 2026-04-30 (and will be released ~2023-04-18) as we have no need to update it but :man_shrugging:

weeman1337 (changes requested):
Same opinion as Igel. We do only use this for builds and not in production. So there is no need to rush for newer versions here, unless it would block us from using newer/other tooling. :heavy_plus_sign: for waiting for the next LTS

MyIgel:
The next LTS is planned for `2023-04-18`

weeman1337:
> The next LTS is planned for `2023-04-18`

That was yesterday @ximex , go ! :laughing:

ximex:
docker image not available yet for node v20: https://hub.docker.com/_/node

ximex:
here it is ;-)

ximex:
now the images is available

PR #1076, opened 2023-01-24
rename `docker-compose` to `docker compose`

ximex:
`docker compose` is the newer way

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -49,39 +49,39 @@ Run this once
 
 ```bash
 cd docker/dev
-docker-compose up
+docker compose up
 ```
 
 Run these commands once initially and then as required after changes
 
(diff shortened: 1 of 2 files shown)


weeman1337:
Reading [this](https://github.com/docker/compose/tree/v2#about-update-and-backward-compatibility) we should really do some testing before merging this. Eventually also provide a link tot he instructions.

ximex:
i worked all the time with `docker compose` for now and had no issue so far

MyIgel:
I don't think that it's that much of a problem as we are not using many "advanced" features (and i had no problems here too) so moving on should be fine but this branch needs a rebase (or rebuild as its easier ;)) which only @ximex can do as its a PR from an org.

ximex:
done

PR #1079, opened 2023-01-27
Improve admin shifts

ximex:
- improve size of start/end date
- use enum's instead of string values
  - use strict checks there
  - better variable naming
- better `shift_over_midnight` check


ximex:
@MyIgel @weeman1337 ready for review/merge

MyIgel:
:exclamation: This is another change of legacy code that needs extensive testing at some point

xuwhite:
Since there is no new feature and the code has to be rewritten for the first complete rewrite I see no point in making the effort of extensive testing for these changes.

PR #1080, opened 2023-01-27
improve profile layout

ximex:
put names and dates next to each other if window width is big enough

### DIFF
--- a/resources/views/pages/settings/profile.twig
+++ b/resources/views/pages/settings/profile.twig
@@ -9,21 +9,24 @@
         {{ csrf() }}
 
         <div class="row g-4">
-            <div class="col-md-12">
+            <div class="col-12">
                 <strong class="help-block">
                     {{ f.entry_required() }} = {{ __('settings.profile.entry_required') }}
                 </strong>
             </div>
-
-            <div class="col-md-12">
+            <div class="col-12">
                 {{ m.info(__('settings.profile.user_details.info')) }}
+            </div>
+            <div class="col-lg-6">
                 {{ f.input(
                     'nick',
                     __('settings.profile.nick'),
                     'text',
                     {'value': user.name, 'disabled': true}
                 ) }}
-                {% if config('enable_pronoun') %}
+            </div>
+            {% if config('enable_pronoun') %}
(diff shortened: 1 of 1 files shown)


ximex:
@MyIgel @weeman1337 ready for review/merge

MyIgel (changes requested):
When changing the layout imho the email and notification checkboxes should also be moved together. Same for dect and mobile phone numbers as they are related too.
I would have done that by myself but am not able to due to the repo being a org repo so here is a  diff instead:

```diff
diff --git a/resources/views/pages/settings/profile.twig b/resources/views/pages/settings/profile.twig
index f0d7cfde..233c4fd5 100644
--- a/resources/views/pages/settings/profile.twig
+++ b/resources/views/pages/settings/profile.twig
@@ -85,15 +85,17 @@
                 </div>
             {% endif %}
 
-            <div class="col-12">
-                {% if config('enable_dect') %}
+            {% if config('enable_dect') %}
+                <div class="col-md-6">
                     {{ f.input(
                         'dect',
                         __('settings.profile.dect'),
                         'text',
                         {'value': user.contact.dect, 'max_length': 40}
                     ) }}
+                </div>
                 {% endif %}
+                <div class="col-md-6">
                 {{ f.input(
                     'mobile',
                     __('settings.profile.mobile'),
@@ -107,6 +109,8 @@
                         user.settings.mobile_show
                     ) }}
                 {% endif %}
+                </div>
+            <div class="col-md-6">
                 {{ f.input(
                     'email',
                     __('settings.profile.email'),
@@ -115,7 +119,7 @@
                 ) }}
             </div>
 
-            <div class="col-12">
+            <div class="col-md-6">
                 {{ f.checkbox(
                     'email_shiftinfo',
                     __('settings.profile.email_shiftinfo', [config('app_name')]),
```

ximex:
have you thought of all variants of the config if it looks good in all cases? This was the reason why i changed only this inputs

MyIgel:
You could eventually put a row around the email & notification options but the rest looks fine imho

ximex:
@MyIgel rebased and applied your patch

PR #1083, opened 2023-01-28
change port for local development to 5080

xuwhite:
because of apple!!!

### DIFF
--- a/docker/dev/docker-compose.yml
+++ b/docker/dev/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       APP_NAME: Engelsystem DEV
     env_file: deployment.env
     ports:
-      - "5000:80"
+      - "5080:80"
     networks:
       - database
       - internet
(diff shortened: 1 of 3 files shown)


MyIgel:
Any opinion on that @marudor / @weeman1337 ?
Eventually having something with 80 (which is not 8080) would be more explicit.

docker/dev/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       APP_NAME: Engelsystem DEV
     env_file: deployment.env
     ports:
-      - "5000:80"
+      - "5001:80"
```
MyIgel, line 23: To be consistent that should be changed in `docker/docker-compose.yml` too (and added to the dev documentation as it is missing there ^^).

MyIgel (changes requested):

docker/dev/docker-compose.yml
xuwhite: done

weeman1337:
`5080`?

PR #1055, opened 2022-12-23
Set strict types everywhere (except includes)

weeman1337:
Also adds a PHPCS rule for that

- Did set up a new system and clicked through each view
- CI https://chaos.expert/engelsystem/engelsystem/-/pipelines/37066

depends on https://github.com/engelsystem/engelsystem/pull/1056

PR #939, opened 2022-07-10
Show real name

PowerPan:
Add A Config Option To Show the Full Name (First and Last Name) instead of the username

PR #1086, opened 2023-02-03
Unified user notifications

MyIgel:
Changed all user notifications to be saved in the session using the same structure and output globally

### DIFF
--- a/src/Controllers/NotificationType.php
+++ b/src/Controllers/NotificationType.php
@@ -1,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Engelsystem\Controllers;
+
+enum NotificationType: string
+{
+    case ERROR = 'error';
+    case WARNING = 'warning';
+    case INFORMATION = 'information';
+    case MESSAGE = 'message';
+}
--- a/src/Renderer/Twig/Extensions/Notification.php
+++ b/src/Renderer/Twig/Extensions/Notification.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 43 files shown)


src/Controllers/NotificationType.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
```
ximex, line 2: add `declare(strict_types=1);` in new files

src/Renderer/Twig/Extensions/Notification.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
```
ximex, line 2: add `declare(strict_types=1);` in new files

tests/Unit/Renderer/Twig/Extensions/NotificationTest.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
```
ximex, line 2: add `declare(strict_types=1);` in new files

ximex (changes requested):

src/Controllers/NotificationType.php
MyIgel: https://github.com/engelsystem/engelsystem/pull/1055 is not merged so :shrug:

weeman1337 (approved):
Looks good :+1: 

Update: Looks not good :D

weeman1337 (changes requested):
Breaks at least add user to shift notification

image

MyIgel:
Eventually having #939 merged first and then adding the real user name to notifications might be useful here?

weeman1337 (approved):
LGTM :+1:

PR #1090, opened 2023-02-06
add phpcs rule "SlevomatCodingStandard.Arrays.TrailingArrayComma"

ximex:
fix all except includes/ directory

don't know if this rule is wanted but i like it as it's easier to look at diffs if an array is changed
should i also run the fixer over the includes directory?


ximex:
done in https://github.com/engelsystem/engelsystem/commit/ad360a70fbea6c225b390cba905ebab58e0a5b07

PR #1091, opened 2023-02-06
phpstan fixes (level 3)

ximex:
fixed some reported errors by running phpstan on level 3


MyIgel:
THis commit could habve been in the https://github.com/engelsystem/engelsystem/pull/1089 PR

MyIgel (changes requested):
This also breaks phpcs linter and even unit tests -> please run them before submitting a PR...

MyIgel:
Also just fixing some parts inst4ad of all is pretty useless -> we should fix all of them or keep it as is so that we don't have to run multiple rounds of "just fixed something" every couple of days

ximex:
this was my first try to work with phpstan. i don't understand why it shouldn't make sense to fix _some_ of the issues. as i didn't understand most of the other problems how to solve them i fixed only them. so you could choose between fixing none of them or some of them.
yes i forgot to run the unit-tests and broke something. this is now reverted

src/Helpers/Schedule/Event.php
```diff
@@ -21,7 +21,7 @@ class Event
      * @param string      $slug globally unique
      * @param string[]    $persons id => name
      * @param string|null $language two letter code
-     * @param string|null $recording license (and opt out in XML, null if not recorded, empty if no license defined)/
+     * @param string      $recording license (and opt out in XML, null if not recorded, empty if no license defined)/
```
MyIgel, line 24: The method type should be fixed here instead

tests/Unit/Models/Stub/BaseModelImplementation.php
```diff
@@ -7,9 +7,6 @@
 use Engelsystem\Models\BaseModel;
 use Illuminate\Database\Eloquent\Builder as QueryBuilder;
 
-/**
- * @property string foo
- */
```
MyIgel, line 12: That whole file can be removed if i see that correctly

MyIgel:
I would fix them by myself just to have that PR merged but i can't as this comes from an org repo

tests/Unit/Models/Stub/BaseModelImplementation.php
ximex: deleted file. also couldn't find any references

src/Helpers/Schedule/Event.php
ximex: don't understand what you mean

MyIgel: @ximex what i meant here is that the `protected string $recording = '',` should be `protected ?string $recording = '',` instead as it is supposed to be nullable by spec.

ximex: done

MyIgel (changes requested):
```
> phpcs -p --cache
............................................................  60 / 543 (11%)
............................................................ 120 / 543 (22%)
...................................................E........ 180 / 543 (33%)
............................................................ 240 / 543 (44%)
............................................................ 300 / 543 (55%)
...........................................E................ 360 / 543 (66%)
............................................................ 420 / 543 (77%)
............................................................ 480 / 543 (88%)
............................................................ 540 / 543 (99%)
...                                                          543 / 543 (100%)



FILE: /var/www/src/Controllers/Metrics/Stats.php
-----------------------------------------------------------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
-----------------------------------------------------------------------------------------------------------------------------------
 178 | ERROR | [x] Method \Engelsystem\Controllers\Metrics\Stats::licenses() has useless @param annotation for parameter $vehicle.
 180 | ERROR | [x] Method \Engelsystem\Controllers\Metrics\Stats::licenses() does not need documentation comment.
-----------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-----------------------------------------------------------------------------------------------------------------------------------


FILE: /var/www/tests/Unit/Controllers/SettingsControllerTest.php
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
FOUND 2 ERRORS AFFECTING 2 LINES
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 378 | ERROR | [x] Method \Engelsystem\Test\Unit\Controllers\SettingsControllerTest::testSavePasswordValidation() has useless @param annotation for parameter $newPassword.
 379 | ERROR | [x] Method \Engelsystem\Test\Unit\Controllers\SettingsControllerTest::testSavePasswordValidation() has useless @param annotation for parameter $newPassword2.
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Time: 3.69 secs; Memory: 26MB
```

Bitte immer alle checks aus führen vorm committen, siehe https://github.com/engelsystem/engelsystem/blob/main/DEVELOPMENT.md#pre-commit-hooks

ximex:
hope this is now ok. run phpcs, phpunit, phpstan

PR #1092, opened 2023-02-06
strict (not-)equal checks

ximex:
use strict (not-)equal checks in db/, src/, tests/
unit tests still works


MyIgel:
To be honest i don't see a good reason her (besides having to type more characters and remember to do so)

ximex:
sorry but typing one more char is no real argument for me to reject this change. the remembering argument is solved by the phpcs rule.
Using only `==` or `!=` instead of `===` or `!==` could lead to problems if you compare things like `1` with `'1'` or want to differentiate `''` and `null` and `[]`. In every software project i worked with everyone also thought so that this is a good change.
see: https://www.php.net/manual/en/types.comparisons.php

xuwhite:
@MyIgel only because you need to test this extensively @ximex is still right in refactoring this in new code... Please test this PR and also apply the strict checks to the new code since this PR was opened.

MyIgel:
The strict equal does not work for example for Carbon instances which currently breaks tests so at least that must be handled differently.

PR #1093, opened 2023-02-06
improve webpack.config.js

ximex:
remove .jsx support, trailing spaces, other small codestyle changes


weeman1337 (changes requested):
If we update the code style, we should also add it to the prettier checks

ximex:
done

weeman1337:
Thanks :+1:

PR #1094, opened 2023-02-06
update libs

ximex:
most of the major version update breaking changes is the new min node version of v14.15. so no problem here.


weeman1337:
Thanks for the PR @ximex .

We should only update libs if one of the following applies:
- We need a bug fix from one of the libs
- There is a critical security issue in one of the libs
- We need a new feature from one of the libs

weeman1337 (approved):

PR #1096, opened 2023-02-09
Important news

### LINKED ISSUE #501: Show important news on the public dashboard
In our latest Heaven mumble, the idea came up, to show news headlines prominently on the public dashboard. At the moment, I don't really have an idea how that would look the best, so maybe this needs some brainstorming.

As a first start: How about adding a flag to news like "show on dashboard" (like "meeting") that can be set by bureaucrats? The headline of the latest news post (of if it looks good space-wise maybe the last 3?) is somehow displayed on the public dashboard. This would allow us to publish critical information at an obvious place very easily. When looking into the angel system, angels then could view the full news post and see additional details in the text. In addition, by doing it this way, it will also reach the angels who do not look at the physical dashboard in Heaven but only into the angel system news (this is the main advantage of the proposal over, e.g. a flipchart in Heaven).

Maybe this would look cool in the form of a news ticker like in TV using a fixed div at the bottom of the screen where the headline scrolls through in all uppercase like "+++ IMPORTANT: WE NEED MORE ANGELS! +++"? If no news post with the dashboard flag is present, the div would simply be hidden.

What do you think?

xuwhite:
fixes #501

### DIFF
--- a/db/factories/NewsFactory.php
+++ b/db/factories/NewsFactory.php
@@ -16,11 +16,12 @@ class NewsFactory extends Factory
     public function definition(): array
     {
         return [
-            'title'      => $this->faker->text(50),
-            'text'       => $this->faker->realText(),
-            'is_meeting' => $this->faker->boolean(),
-            'is_pinned'  => $this->faker->boolean(.1),
-            'user_id'    => User::factory(),
+            'title'        => $this->faker->text(50),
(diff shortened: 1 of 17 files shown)


xuwhite:
@MyIgel pls fix merge conflicts

xuwhite:
maybe i fixed them

PR #1097, opened 2023-02-13
Reimplemented about angeltypes page

MyIgel:


weeman1337 (approved):
Tested, looks good :+1:

PR #1098, opened 2023-02-19
refactor goodie config options

xuwhite:
/UserShirtControllerTest.php: testSaveShirt is failing, running it multiple times gives different outputs and I don't know why *shrug*


xuwhite:
ready now

PR #1100, opened 2023-02-25
refactor goodie config option as enum

xuwhite:


config/config.default.php
```diff
@@ -275,14 +277,11 @@
...
-    // When true changes everything from shirts to goodies and disables shirt size
-    'other_goodie'            => (bool) env('OTHER_GOODIE', false),
-
-    // Enables the goody configuration on signup and profile
-    'enable_goody'            => (bool) env('ENABLE_GOODY', false),
+    // Resembles the Goodie Type. There are three options:
+    // 'none' => no goodie at all
+    // 'goodie' => a goodie which has no sizing options
+    // 'tshirt' => goodie that is called tshirt and has sizing options
+    'goodie_type' => GoodieType::from(env('GOODIE_TYPE', 'goodie')),
```
MyIgel, line 284: The only allowed function in config files is env() as it just reads values and static castst. Everything else will confuse users.

The `GoodieType::from` should be used in the code itself instead.

xuwhite: where else?

xuwhite: done

config/routes.php
```diff
@@ -174,7 +174,7 @@ function (RouteCollector $route): void {
             function (RouteCollector $route): void {
                 // Shirts
                 $route->addGroup(
-                    '/shirt',
+                    '/goodie',
```
xuwhite, line 177: und für die kleinen changes ein ganzer commit ???
grummel grummel

xuwhite:
git stash hilft gummmel grummel

config/routes.php
MyIgel: Just to not confuse you...

PR #1117, opened 2023-03-29
hide language/theme

ximex:
hide language/theme menu if not more than 1 selection is available

ready for review/merge

### DIFF
--- a/resources/views/layouts/parts/navbar.twig
+++ b/resources/views/layouts/parts/navbar.twig
@@ -92,14 +92,16 @@
                     </li>
                 {% endif %}
 
-                <li class="nav-item dropdown">
-                    <a class="nav-link dropdown-toggle" href="#" role="button" data-bs-toggle="dropdown" aria-expanded="false">
-                        {{ m.icon('translate') }}
-                    </a>
-                    <ul class="dropdown-menu dropdown-menu-end">
-                        {{ menuLanguages()|join(" ")|raw }}
-                    </ul>
-                </li>
+                {% if config('locales')|length > 1 %}
+                    <li class="nav-item dropdown">
+                        <a class="nav-link dropdown-toggle" href="#" role="button" data-bs-toggle="dropdown" aria-expanded="false">
+                            {{ m.icon('translate') }}
+                        </a>
+                        <ul class="dropdown-menu dropdown-menu-end">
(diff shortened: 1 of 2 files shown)


resources/views/layouts/parts/navbar.twig
```diff
@@ -92,14 +92,16 @@
...
 
-                <li class="nav-item dropdown">
-                    <a class="nav-link dropdown-toggle" href="#" role="button" data-bs-toggle="dropdown" aria-expanded="false">
-                        {{ m.icon('translate') }}
-                    </a>
-                    <ul class="dropdown-menu dropdown-menu-end">
-                        {{ menuLanguages()|join(" ")|raw }}
-                    </ul>
-                </li>
+                {% if showLanguagesMenu() %}
```
MyIgel, line 95: As this function is only used once i would just directly use `config('locales') | length > 1`

src/Controllers/SettingsController.php
```diff
@@ -243,10 +243,16 @@ public function settingsMenu(): array
         $menu = [
             url('/settings/profile')  => 'settings.profile',
             url('/settings/password') => 'settings.password',
-            url('/settings/language') => 'settings.language',
-            url('/settings/theme')    => 'settings.theme',
         ];
 
+        if (count(config('locales')) >= 2) {
```
MyIgel, line 248: Imho `> 1` would be easier to read/understand here

MyIgel (changes requested):

src/Controllers/SettingsController.php
```diff
@@ -243,10 +243,16 @@ public function settingsMenu(): array
...
             url('/settings/password') => 'settings.password',
-            url('/settings/language') => 'settings.language',
-            url('/settings/theme')    => 'settings.theme',
         ];
 
+        if (count(config('locales')) > 1) {
+            $menu[url('/settings/language')] = 'settings.language';
+        }
+
+        if (count(config('themes')) >= 2) {
```
MyIgel, line 252: Here too (as its even more confusing that way) ;)

MyIgel (changes requested):

src/Controllers/SettingsController.php
ximex: fixed

PR #1120, opened 2023-04-01
Fix migrations, routing, layout and added CI caching

MyIgel:


config/config.default.php
```diff
@@ -13,8 +13,8 @@
         'password' => env('MYSQL_PASSWORD', ''),
     ],
 
-    // For accessing stats
-    'api_key'                 => '',
+    // For accessing /metrics
```
ximex, line 16: It's for `/stats` and `/metrics`

ximex (changes requested):

weeman1337 (approved):
fetzt :+1: 

Merge it in the way you prefer @MyIgel

PR #1125, opened 2023-04-06
Add Username validation

weeman1337:
- New `Username`
- Replaces `User_validate_Nick`

The validation no longer has side effects. Before it trimmed the string.


weeman1337:
Pipeline: https://chaos.expert/engelsystem/engelsystem/-/pipelines/38481

tests/Unit/Http/Validation/Rules/UsernameTest.php
```diff
@@ -0,0 +1,80 @@
...
+            'empty string' => ['', false],
+            'max length exceeded' => [str_repeat('1', 25), false],
+            'invalid chars' => ['abc!!!', false],
+            'valid Greek letters' => ['λουκάνικο', true],
+            'min length valid' => ['a', true],
+            'valid with all chars' => ['abc123_-. jkl', true],
+            'valid with accents' => ['café', true],
+            'max length valid' => [str_repeat('a', 24), true],
+        ];
+    }
```
ximex, line 42: is the same as `provideValidateWithCustomConfigTestData`. only `'valid Greek letters'` and `'min length valid'` are swapped

src/Http/Validation/Rules/Username.php
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Username validation.
+ * Usernames must have 1-24 chars and only contain alphanumeric chars or "-", "_", "." or "<space>".
+ */
+class Username extends AbstractRule
+{
+    public function validate(mixed $input): bool
+    {
+        $regex = config('username_regex', '/([^\p{L}\p{N}\-_. ]+)/ui');
```
MyIgel, line 18: I would remove that regex from here as it is confusing (and differs from the config which overwrites it)

tests/Unit/Http/Validation/Rules/UsernameTest.php
weeman1337: Will be gone when moving the default to the config.

src/Http/Validation/Rules/Username.php
weeman1337: Du machst mich feddisch… 

> (and differs from the config which overwrites it)

Until now the default config did not contain the regular expression.

weeman1337:
[Recent pipeline](https://chaos.expert/engelsystem/engelsystem/-/pipelines/38581)

weeman1337:
https://chaos.expert/engelsystem/engelsystem/-/pipelines/38605

PR #1128, opened 2023-04-19
Tweak navbar collapse as guest + fix language dropdown position

weeman1337:
- If not logged in, the navbar does not need to collapse so early
- It is common sense that the upper right item is a primary action / the user menu. Currently this is annoyingly the language switcher. Swapping this with the Login/Register buttons as guest and the user menu as logged in user.

---

## Before

Login page small(er) screen

login_before_1

Login page with expanded navbar

login_before_2

Logged in view

logged_in_before

---

## After

login_after

logged_in_after


weeman1337:
[pipeline](https://chaos.expert/engelsystem/engelsystem/-/pipelines/38640)

Issue #1124, opened 2023-04-06
Register page overhaul


weeman1337:
The sign-up page is the second page people will see after the home page. It does currently look chaotic and crowded. There is a plan to redesign the page roughly like this:

![image](https://user-images.githubusercontent.com/6216686/230389451-f2f39c6a-a241-4151-9494-b75864527fa5.png)

PR #1138, opened 2023-05-03
Add ShirtSize validation

weeman1337:
Towards https://github.com/engelsystem/engelsystem/issues/1124

[Pipeline](https://chaos.expert/engelsystem/engelsystem/-/pipelines/38815)

### DIFF
--- a/src/Http/Validation/Rules/ShirtSize.php
+++ b/src/Http/Validation/Rules/ShirtSize.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Engelsystem\Http\Validation\Rules;
+
+use Respect\Validation\Rules\In;
+
+/**
+ * Validates whether the value is contained in the keys of the config item "tshirt_sizes".
+ */
+class ShirtSize extends In
+{
+    public function __construct()
+    {
+        parent::__construct(array_keys(config('tshirt_sizes')));
+    }
+}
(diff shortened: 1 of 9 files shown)


tests/Unit/Http/Validation/Rules/ShirtSizeTest.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        self::assertSame($expectedValid, $this->subject->validate($value));
+    }
+
+    /**
+     * @return array<string>
+     */
+    private function loadDefaultShirtSizes(): array
+    {
+        $defaultConfig = include __DIR__ . '/../../../../../config/config.default.php';
```
MyIgel, line 54: Using the default config is not the best idea as users might change it (esp. for tshirts). It woudl be better to instantiate a new config and set it in the container. (and the other places that use the config file should be removed later)

src/Http/Validation/Rules/ShirtSize.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Engelsystem\Http\Validation\Rules;
+
+use Respect\Validation\Rules\In;
+
+/**
+ * Validates whether the value is contained in the keys of the config item "tshirt_sizes".
+ */
+class ShirtSize extends In
+{
+    public function __construct()
```
MyIgel, line 14: Having validation messages here (exists & wrong) might be helpful here as users might have none configured (change from disabled to enabled in config) or the wrong one (the organizer removed / changed some sizes after users have registered)?

weeman1337: Field is required and there is an empty option for all shirt selects.

MyIgel (changes requested):

PR #1151, opened 2023-06-29
Add .devcontainer for easy IDE integration

poelzi:
Allows to easily start and connect to the docker compose dev environment.

Uses the already provided compose file

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,17 @@
+{
+    "dockerComposeFile": "docker/dev/docker-compose.yml",
+	"service": "es_workspace",
+    "runServices": [
+        "es_database",
+        "es_server"
+    ],
+	"customizations": {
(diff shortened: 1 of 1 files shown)


MyIgel:
Hmm, at the moment none of us is using vscode (or derived editors that implement the spec) so maybe adding it to the docs would be more helpful?

poelzi:
It is one file which references the already existing and maintained docker file.
What is the advantage of increasing the docs so people do copy paste from documentation into a file to start the ide again instead of having it as a proper file in the first place ?
It can not even be tested when it is in the documation

MyIgel:
The point here is that configs for various IDEs are at the end cluttering the project root and there are too many IDEs with even more users that have their own preferences and ideas what plugins/extensions should be loaded and which not.
Supporting them all is not possible so having possible options documented is a great starting point but forcing people to use "the great one" will not work.

PR #1156, opened 2023-07-14
Smaller bugfixes & changes

MyIgel:
* Fix 500 error on angeltypes edit page when logged out
* Fix error when OAuth provider sends id as int
* Sign-Up button not always displayed

### DIFF
--- a/includes/controller/angeltypes_controller.php
+++ b/includes/controller/angeltypes_controller.php
@@ -92,7 +92,7 @@ function angeltype_edit_controller()
         // Edit existing angeltype
         $angeltype = AngelType::findOrFail($request->input('angeltype_id'));
 
-        if (!auth()->user()->isAngelTypeSupporter($angeltype) && !auth()->can('admin_user_angeltypes')) {
+        if (!auth()->user()?->isAngelTypeSupporter($angeltype) && !auth()->can('admin_user_angeltypes')) {
             throw_redirect(page_link_to('angeltypes'));
         }
     } else {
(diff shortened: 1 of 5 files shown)


includes/controller/angeltypes_controller.php
```diff
@@ -92,7 +92,10 @@ function angeltype_edit_controller()
         // Edit existing angeltype
         $angeltype = AngelType::findOrFail($request->input('angeltype_id'));
 
-        if (!auth()->user()->isAngelTypeSupporter($angeltype) && !auth()->can('admin_user_angeltypes')) {
+        if (
+            !auth()->user()
+            || (!auth()->user()->isAngelTypeSupporter($angeltype) && !auth()->can('admin_user_angeltypes'))
```
weeman1337, line 97: The future is now! :rocket: 

This should work:

```suggestion
            !auth()->user()?->isAngelTypeSupporter($angeltype) && !auth()->can('admin_user_angeltypes')
```

src/Controllers/OAuthController.php
```diff
@@ -150,7 +150,7 @@ public function index(Request $request): Response
         if (!$oauth) {
             return $this->redirectRegister(
                 $providerName,
-                $resourceId,
+                (string) $resourceId,
```
weeman1337, line 153: Can you modify one of the tests to provide a non-string value here? Then it would be covered by a test.

weeman1337 (commented):
Looks good. Some minor comments.

weeman1337 (approved):

PR #1158, opened 2023-07-25
A small step for an API (v0.0.1-beta)

MyIgel:
This PR adds a minimal, read-only, API to get general information about shifts.

* Base path: `/api/v0-beta`
* `/angeltypes` for a list of all angeltypes
* `/angeltypes/{id}/shifts` all shifts that need the angeltype
* `/news` to show all news
* `/locations` an overview of all locations
* `/locations/{id}/shifts` get all shifts in a location
* `/users/{id}` show user information
* `/users/{id}/angeltypes` get all angeltypes of a user
* `/users/{id}/shifts` the users shifts
* `/info` API & Application state information
* `/openapi` OpenAPI specification file

where the user ID can be replaced with `self`

The `openapi.yml` can be conveniently viewed in [editor.swagger.io](https://editor.swagger.io).
This PR should be merged after the CERT was able to verify that their needed endpoints where implemented successfully ;)


config/routes.php
```diff
@@ -101,7 +101,32 @@ function (RouteCollector $route): void {
...
+            function (RouteCollector $route): void {
+                $route->addRoute(['OPTIONS'], '[/{resource:.+}]', 'Api\IndexController@options');
+                $route->get('', 'Api\IndexController@indexV0');
+
+                $route->get('/angeltypes', 'Api\AngelTypeController@index');
+                $route->get('/news', 'Api\NewsController@index');
+                $route->get('/rooms', 'Api\RoomsController@index');
+                $route->get('/rooms/{room_id:\d+}/shifts', 'Api\ShiftsController@entriesByRoom');
+
+                $route->addRoute(
```
weeman1337, line 120: I would say this should return a 404 for unknown resources.
And for the resources from above the response should be 405 for any method except `GET`.

501 should be used if the server implements some kind of protocol.

Reference in MDN: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501
- . The only methods that servers are required to support (and therefore that must not return 501) are [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD).
- If the server does recognize the method, but intentionally does not support it, the appropriate response is [405 Method Not Allowed](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405).

resources/api/openapi.yml
```diff
@@ -0,0 +1,382 @@
...
+          type: array
+          items:
+            $ref: '#/components/schemas/User'
+        type:
+          $ref: '#/components/schemas/AngelType'
+        needs:
+          type: integer
+          description: |
+            Number of users needed for the shift.
+            Can be more than users count when not full, less when overbooked or 0 when additional users where added.
```
weeman1337, line 218:
```suggestion
            Can be more than users count when not full, less when overbooked, or 0 when additional users have been added.
```

resources/api/openapi.yml
```diff
@@ -0,0 +1,382 @@
...
+          example: 42
+        name:
+          type: string
+          example: Angel
+        description:
+          type: string
+          example: Meta-Group of all registered Angels
+        url:
+          type: string
+          example: https://example.com/news/42
```
weeman1337, line 71: This example URL does not look correct. Can you add a description to the URL fields to explain where they are leading to? Such as „Browsable URL of the angel type in the Engelsystem“.

src/Controllers/Api/IndexController.php
```diff
@@ -0,0 +1,63 @@
...
+                'description' => $info->description,
+                'paths' => $paths,
+            ]));
+    }
+
+    public function options(): Response
+    {
+        // Respond to browser preflight options requests
+        return $this->response
+            ->setStatusCode(204)
```
weeman1337, line 52: The API should probably better use 200 here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS#status_code Even though it mentions „browsers“ there, it could also be that some clients suffer from the same problems :man_shrugging: .

src/Controllers/Api/ShiftsController.php
```diff
@@ -0,0 +1,140 @@
...
+use Engelsystem\Models\Shifts\NeededAngelType;
+use Engelsystem\Models\Shifts\Shift;
+use Illuminate\Database\Eloquent\Collection;
+
+class ShiftsController extends ApiController
+{
+    public function entriesByRoom(Request $request): Response
+    {
+        $roomId = (int) $request->getAttribute('room_id');
+        /** @var Room $room */
```
weeman1337, line 19: Can you add `findOrFail` and `shifts` to [the `Room` model doc](https://github.com/engelsystem/engelsystem/blob/main/src/Models/Room.php#L15)?
Then it will work everywhere and inline type hints should not be required.

tests/Unit/Controllers/Api/AngelTypeControllerTest.php
```diff
@@ -0,0 +1,33 @@
...
+
+        $response = $controller->index();
+        $this->validateApiResponse('/angeltypes', 'get', $response);
+
+        $this->assertEquals(['application/json'], $response->getHeader('content-type'));
+        $this->assertJson($response->getContent());
+
+        $data = json_decode($response->getContent(), true);
+        $this->assertArrayHasKey('data', $data);
+        $this->assertCount(3, $data['data']);
```
weeman1337, line 31: Should the test not also assert, that the correct data is returned (not only syntactically with API schema validation)? Applies to the other tests as well.

src/Middleware/ApiRouteHandler.php
```diff
@@ -0,0 +1,78 @@
...
+        $isApi = $this->apiPrefix && (Str::startsWith($path, $this->apiPrefix . '/') || $path == $this->apiPrefix);
+        $request = $request->withAttribute('route-api', $isApi);
+
+        return $isApi ? $this->processApi($request, $handler) : $handler->handle($request);
+    }
+
+    /**
+     * Process the API request by ensuring that JSON is returned
+     */
+    protected function processAPi(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```
weeman1337, line 48:
```suggestion
    protected function processApi(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
```

weeman1337 (commented):
Looks good in general :+1:

Some minor comments. The main point of discussion would be the HTTP status codes.

src/Controllers/Api/ShiftsController.php
MyIgel: `findOrFail` is part of the `vendor/illuminate/database/Eloquent/Builder.php` and thus should not be needed to be addressed explicitly and shifts is already there?

src/Controllers/Api/IndexController.php
MyIgel: Not sure which "browsers" or "clients" that should be but whatever

KampfCaspar:
I try to test the API for CERT - but know next to nothing about the inner workings of Engelsystem ;(
Are there OAuth2 endpoints to get permission? Basic Auth seems not to work. An open session helps, but emulating a user client seems a bit cumbersome ;) Any hints appreciated...

MyIgel:
@KampfCaspar i can give a short introduction to that in the meeting next week :)

MyIgel:
Additional requirement: All angeltypes of a given user

xuwhite:
Looks good

db/migrations/2023_05_21_000000_create_api_permissions.php
```diff
@@ -0,0 +1,45 @@
...
+            ['name' => 'api', 'description' => 'Use the API'],
+        ]);
+        $db->table('groups')->insert([
+            ['id' => 40, 'name' => 'API'],
+        ]);
+
+        $shiftCoordinatorGroup = 60;
+        $apiId = $db->table('privileges')->where('name', 'api')->first()->id;
+        $db->table('group_privileges')->insert([
+            ['group_id' => $shiftCoordinatorGroup, 'privilege_id' => $apiId],
```
xuwhite, line 27: I think it would be better to put this privilege into the buerocrat group for now

resources/api/openapi.yml
```diff
@@ -0,0 +1,455 @@
...
+        title:
+          type: string
+          example: Cleanup the venue
+        description:
+          type: string
+          example: You clean up the venue after the event, its fun, we promise!
+          description: |
+            Shift description, should be added to the shift type description as it might be empty.
+            Normally contains additional information for a specific shift / task.
+        start:
```
johnjohndoe, line 195: I made good experience with using the `_at` suffix for date/time fields in general. It clarifies the expected type already
Hence, I recommend naming `start` -> `starts_at` and `end` - `ends_at`.

johnjohndoe:
Is the `Comment` resp. `user_comment` still part of the API? Will it be Markdown text like in `description` and `shifttype_description`?

resources/api/openapi.yml
MyIgel: Thats a good idea, i'll change that it

MyIgel:
> Is the `Comment` resp. `user_comment` still part of the API? Will it be Markdown text like in `description` and `shifttype_description`?

The user comment will be part of the users own shifts export which i will implement soon(TM).
At the moment the API implements the required endpoints for the CERT backend and is "hidden" (permission wise) from "normal" users to be tested at the congress and then extended and released at some point in the future ^^
(and to the comment: at the moment its not markdown, just plain text)

johnjohndoe:
Thank you. Is there a description of the use case of the CERT somewhere?

MyIgel:
At the moment not, basically they have an internal dispatcher system which needs to know the state of their shifts (who is doing what and how qualified is that person) and a list of news (so basically what the API is currently implementing).

I will add additional endpoints soon so if you have ideas/requirements/input you can add them here :)

MyIgel:
The PR works for the CERT so its ready for review & merge :tada:

PR #1161, opened 2023-07-31
Add release script

poelzi:
Add script to create release archives

Had to create releases from a custom branch regularly. Should be usefull for others as well

### DIFF
--- a/bin/create_release.sh
+++ b/bin/create_release.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -x
+
+AUTOVERSION=$(git describe --tags)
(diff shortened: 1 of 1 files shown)


poelzi:
The release is still a bit fat, I think lot's of files could be removed after some build finished, but I haven'ts investigeted which parts.

MyIgel:
I don't know how useful this script would be for anyone else as normally it is not needed to package the Engelsystem in that particular way (and just using GitLab for it is normally enough as the turnaround time is around ~10 min. (Although i would be interested to know more about your use case here)

xuwhite:
closed due to inactivity, feel free to reopen / discuss this topic in an issue

PR #1166, opened 2023-08-03
Add registration link when password is disabled (when using OAuth)

poelzi:
Some small improvements to make oauth more smooth and less complains by users.

Allows you to set a link for the register button if register is disabled.
Does not show login form by default if enable_password is false. There is a small link to show the form or by passing password in the get parameter.

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -256,6 +256,9 @@ return [
     // Users are able to sign up
     'registration_enabled'    => (bool) env('REGISTRATION_ENABLED', true),
 
+    // URL to external registration page, used on login page
+    'external_registration_url'   => env('EXTERNAL_REGISTRATION_URL'),
+
     // Required user fields
     'required_user_fields' => [
         'pronoun'            => (bool) env('PRONOUN_REQUIRED', false),
@@ -300,9 +303,8 @@ return [
     // The minimum length for passwords
     'min_password_length'     => env('PASSWORD_MINIMUM_LENGTH', 8),
 
-    // Whether the Password field should be enabled on registration.
-    // This is useful when using oauth, disabling it also disables normal
-    // registration without oauth.
+    // Whether the login and registration via password should be enabled (login will be hidden)
+    // This is useful when using oauth, disabling it also disables normal registration without oauth
     'enable_password'         => (bool) env('ENABLE_PASSWORD', true),
 
     // Whether the DECT field should be enabled
--- a/resources/views/pages/login.twig
+++ b/resources/views/pages/login.twig
@@ -30,7 +30,8 @@
 
         <div class="row mb-5">
             <div class="col-md-6 offset-md-3 col-lg-4 offset-lg-4">
-                <div class="card {{ m.type_bg_class() }}">
+                <div class="card {{ m.type_bg_class() }} registration">
+                    {% set login_hidden = not config('enable_password') %}
                     <div class="card-body">
                         {% include 'layouts/parts/messages.twig' %}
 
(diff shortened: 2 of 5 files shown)


config/config.default.php
```diff
@@ -229,6 +229,9 @@
     // Users are able to sign up
     'registration_enabled'    => (bool) env('REGISTRATION_ENABLED', true),
 
+    // URL to external register page
```
MyIgel, line 232: Should mention that this functionality is only available when `enable_password` is `false`. Ans naming it `external_registration_url` would imho fit better with the current naming scheme

resources/views/pages/login.twig
```diff
@@ -40,6 +40,7 @@
 
                         <form action="" enctype="multipart/form-data" method="post">
                             {{ csrf() }}
+                            {% if config('enable_password') or request.get('password') %}
```
MyIgel, line 43: I don't like this additional "clunky" parameter, see comment below:

resources/views/pages/login.twig
```diff
@@ -73,6 +74,11 @@
                                     {{ __('login.login') }}
                                 </button>
                             </div>
+                            {% else %}
+                            <div class="mb-3 text-center">
+                                <a href="{{ url(request.url(), {'password':1}) }}">…</a>
```
MyIgel, line 79: That is a not-so-pretty workaround (as its only presumably used by admins when the OAuth is not working) and imho we should use the existing feature of "showing" stuff when something gets clicked which is used to show additional OAuth buttons in https://github.com/engelsystem/engelsystem/blob/main/resources/assets/js/forms.js#L273

resources/views/pages/login.twig
```diff
@@ -89,11 +95,13 @@
                                 </div>
                             {% endif %}
 
+                            {% if config('enable_password') or request.get('password') %}
```
MyIgel, line 98: (same here of course)

resources/views/pages/login.twig
```diff
@@ -107,11 +115,17 @@
                     {% if config('enable_password') %}
                         <p>{{ __('Please sign up, if you want to help us!') }}</p>
                         <a href="{{ url('register') }}" class="btn btn-primary">{{ __('Register') }} &raquo;</a>
+                    {% elseif config('external_register_url') %}
+                        <a href="{{ config('external_register_url') }}" class="btn btn-primary">{{ __('Register') }} &raquo;</a>
```
MyIgel, line 119: At the moment this "doubles" the Register title above so we should either change the text or remove the title?
Also having multiple "buttons" in the code makes it imho not that greatly readable, so just changing the URL via short syntax would be imho preferrable here.

resources/views/pages/login.twig
```diff
@@ -107,11 +115,17 @@
...
                     {% else %}
                         <p>{{ __('Registration is only available via external login.') }}</p>
                     {% endif %}
                 {% else %}
-                    {{ m.alert(__('Registration is disabled.'), 'danger') }}
+                    {% if config('external_register_url') %}
+                        <a href="{{ config('external_register_url') }}" class="btn btn-primary">{{ __('Register') }} &raquo;</a>
```
MyIgel, line 125: (same here obviously)

MyIgel (changes requested):
Hi, thank you for those improvements, it shows that we eventually should have a look around the general workflow here a bit more in general at some point :D

I added some smaller comments to some parts:

MyIgel:
Hi @poelzi, do you have some time to have a look here? (it also has merge conflicts so needs a rebase)

PR #1185, opened 2023-08-17
Fix msgfmt error: keyword "settings" unknown

jplitza:
When running
```
docker compose exec es_workspace find /var/www/resources/lang -type f -name '*.po' -exec sh -c 'file="{}"; msgfmt "${file%.*}.po" -o "${file%.*}.mo"' \;
```
(as instructed in DEVELOPMENT.md) I got this:

```
/var/www/resources/lang/en_US/default.po:311: keyword "settings" unknown
/var/www/resources/lang/en_US/default.po:311:15: syntax error
msgfmt: found 2 fatal errors
```

### DIFF
--- a/resources/lang/en_US/default.po
+++ b/resources/lang/en_US/default.po
@@ -308,7 +308,7 @@ msgstr "Here you can change your theme."
 msgid "settings.theme.success"
 msgstr "Theme was changed successfully."
 
-msgid settings.certificates
+msgid "settings.certificates"
 msgstr "Certificates"
 
 msgid "settings.certificates.info"


MyIgel:
Eventually we should add a linter that checks the format here :D

jplitza:
I was thinking more about the lines of executing the setup commands in CI

MyIgel:
@jplitza it already gets executed, see https://github.com/engelsystem/engelsystem/blob/c63a671dc460dd34027f8dcdd5c1e15cb49bcf97/docker/Dockerfile#L11
But maybe its another version there or something?

PR #1196, opened 2023-08-28
Degender shirt sizes shortcuts

skorpy2009:


config/config.default.php
```diff
@@ -338,13 +338,13 @@
     // Available T-Shirt sizes, set value to null if not available
     'tshirt_sizes'            => [
```
MyIgel, line 339: Those changes should be accompanied with a migration that changes the values of the `users_personal_data.shirt_size` column.

MyIgel (changes requested):

PR #1201, opened 2023-08-30
update models with defaults

### LINKED ISSUE #1199: Update models with defaults
Some of the models are missing default values / castings etc.
All models should be compared with their tables.

xuwhite:
fixes #1199

### DIFF
--- a/src/Models/Room.php
+++ b/src/Models/Room.php
@@ -39,6 +39,13 @@ class Room extends BaseModel
     /** @var bool Enable timestamps */
     public $timestamps = true; // phpcs:ignore
 
+    /** @var array<string, null> default attributes */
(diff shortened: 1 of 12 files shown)


src/Models/Room.php
```diff
@@ -39,6 +39,13 @@ class Room extends BaseModel
     /** @var bool Enable timestamps */
     public $timestamps = true; // phpcs:ignore
 
+    /** @var array<string, null> default attributes */
+    protected $attributes = [ // phpcs:ignore
+        'map_url' => null,
```
MyIgel, line 44: Something indentation something? Consistent styling might be  a good idea.

MyIgel (changes requested):

PR #1207, opened 2023-09-08
Add dutch translation file

Bladedfast:
Dutch translation is added for broader user base (Like myself and my peers)

### DIFF
--- a/resources/lang/nl_BE/default.po
+++ b/resources/lang/nl_BE/default.po
@@ -1,0 +1,535 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: Engelsystem 2.0\n"
+"POT-Creation-Date: 2017-12-29 19:01+0100\n"
+"PO-Revision-Date: 2019-06-04 23:41+0200\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


xuwhite:
sadly that doesnt translate everything. we have very much legacy translation (hardcoded english text which is only translated in the german translation file) 
maybe you should have a look at the german translation file and translate everything to dutch not only parts

MyIgel:
& in general, when adding an additional translation, there must be someone to maintain them (looking at the, effectively "dead" translation `pt_BR`).
And what @xuwhite says: A half translated engelsystem is not that great to use.

I would personally suggest to wait with the Translation until the current bigger rewrite is more or less "done" to have less work with rewriting translations.

MyIgel:
@Bladedfast how do you see the chance, that the engelsystem gets a (mostly full) translation at some point?
Without at least some commitment i would rather not add a half one at that point and eventually wait for most of the engelsystem to be in a somewhat "stable" state?
(And after that having some discussion about being able to have a workflow for a sustainable further translation of it so that every release has a state that is mostly useable without having a mush of half translated Dutch and English parts mixed)

xuwhite:
Since there has been no update, I'll close it for now. Feel free to reopen the PR, if you have time.

PR #1213, opened 2023-09-17
Implement user session management

MyIgel:
This PR adds a new Page that lets a user view all its sessions and delete some or all of them (which required some logic while saving the session).
In addition we also clear all sessions on password reset and all not active ones when setting a new password.


weeman1337:
Ausgezeichnet!

PR #1215, opened 2023-09-18
rename is_important to is_highlighted

xuwhite:


db/migrations/2023_09_18_000000_permissions_rename_news_important_to_news_highlight.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
MyIgel, line 1: Imho those migrations should be combined into one to make migrating easier

src/Models/News.php
```diff
@@ -18,7 +18,7 @@
  * @property string                        $text
  * @property bool                          $is_meeting
  * @property bool                          $is_pinned
- * @property bool                          $is_important
+ * @property bool                          $is_highlighted
```
MyIgel, line 21: Normally those should be re-arranged to stay in alphabetic order but thats not the case here anyway

MyIgel (changes requested):

db/migrations/2023_09_18_000000_permissions_rename_news_important_to_news_highlight.php
xuwhite: what should be the name of the migration?

PR #1221, opened 2023-09-30
make required fields configurable

### LINKED ISSUE #897: Feature Request: Pflichtfelder im Config definieren
Pflichtfelder als Confligflag und nicht hardcoded.

xuwhite:
I purposely not implemented planned_arrival_date in the config because at the moment it is not possible to not pick an arrival date even if it is not required. I think it makes no sense implementing it without messing with the date picker which is out of scope for me.
Fixes #897

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -233,6 +233,17 @@ return [
     // Users are able to sign up
     'registration_enabled'    => (bool) env('REGISTRATION_ENABLED', true),
 
+    // Required fields on sign-up page
+    'signup_required_fields'   => [
+        'pronoun'              => (bool) env('PRONOUN_REQUIRED', false),
+        'firstname'            => (bool) env('FIRSTNAME_REQUIRED', false),
+        'lastname'             => (bool) env('LASTNAME_REQUIRED', false),
+        'planned_arrival_date' => (bool) env('PLANNED_ARRIVAL_DATE_REQUIRED', true),
+        'tshirt_size'          => (bool) env('TSHIRT_SIZE_REQUIRED', true),
+        'mobile'               => (bool) env('MOBILE_REQUIRED', false),
+        'dect'                 => (bool) env('DECT_REQUIRED', false),
+    ],
+
     // Only arrived angels can sign up for shifts
     'signup_requires_arrival' => (bool) env('SIGNUP_REQUIRES_ARRIVAL', false),
 
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -238,7 +238,6 @@ return [
         'pronoun'              => (bool) env('PRONOUN_REQUIRED', false),
         'firstname'            => (bool) env('FIRSTNAME_REQUIRED', false),
         'lastname'             => (bool) env('LASTNAME_REQUIRED', false),
-        'planned_arrival_date' => (bool) env('PLANNED_ARRIVAL_DATE_REQUIRED', true),
         'tshirt_size'          => (bool) env('TSHIRT_SIZE_REQUIRED', true),
         'mobile'               => (bool) env('MOBILE_REQUIRED', false),
         'dect'                 => (bool) env('DECT_REQUIRED', false),
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -234,13 +234,13 @@ return [
     'registration_enabled'    => (bool) env('REGISTRATION_ENABLED', true),
 
     // Required fields on sign-up page
-    'signup_required_fields'   => [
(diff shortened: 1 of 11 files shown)


src/Controllers/SignUpController.php
```diff
@@ -101,6 +102,12 @@ private function renderSignUpPage(): Response
                 'isPronounEnabled' => $this->config->get('enable_pronoun'),
                 'isFullNameEnabled' => $this->config->get('enable_user_name'),
                 'isPlannedArrivalDateEnabled' => $this->config->get('enable_planned_arrival'),
+                'isPronounRequired' => $requiredFields['pronoun'],
```
MyIgel, line 105: Using variables here imho does not make it better readable but as it done so before thats ok (and should definitely be refactored at a later stage)

src/Factories/User.php
```diff
@@ -67,6 +67,16 @@ public function determineBuildUpStartDate(): DateTimeInterface
         return $this->config->get('buildup_start') ?? CarbonImmutable::now();
     }
 
+    private function required(string $key): string
+    {
+        $requiredFields = $this->config->get('signup_required_fields');
+        $isRequired = 'optional';
```
MyIgel, line 73:
```suggestion
        return $requiredFields[$key] ? 'required' : 'optional';
```

tests/Utils/SignUpConfig.php
```diff
@@ -27,10 +34,18 @@ public static function setMaximumConfig(Config $config): void
         $config->set('enable_user_name', true);
         $config->set('enable_mobile_show', true);
         $config->set('enable_dect', true);
+        $config->set('signup_required_fields', $requiredFields);
     }
 
     public static function setMinimumConfig(Config $config): void
     {
+        $requiredFields = $config->get('signup_required_fields');
```
MyIgel, line 42: (same here)

MyIgel (changes requested):
I have some smaller suggestions to this ;)

config/config.default.php
```diff
@@ -233,6 +233,16 @@
     // Users are able to sign up
     'registration_enabled'    => (bool) env('REGISTRATION_ENABLED', true),
 
+    // Required fields on sign-up page
+    'signup_required_fields'   => [
+        'pronoun'              => (bool) env('PRONOUN_REQUIRED', false),
```
MyIgel, line 238: Spacing might be improved too ;)

MyIgel:
Another point is that
* a user currently can remove some data after sign up in their profile
* There is no hint implemented that notifies that some fields are missing when the config gets changed during an evnet

src/Controllers/SignUpController.php
xuwhite: I like it that way!
won't change

tests/Utils/SignUpConfig.php
```diff
@@ -11,6 +11,13 @@ final class SignUpConfig
 {
     public static function setMaximumConfig(Config $config): void
     {
+        $requiredFields = $config->get('signup_required_fields');
```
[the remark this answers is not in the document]
xuwhite: done

src/Factories/User.php
xuwhite: done

tests/Utils/SignUpConfig.php
xuwhite: what same?

config/config.default.php
xuwhite: done

xuwhite:
> Another point is that
> 
>     * a user currently can remove some data after sign up in their profile
> 
>     * There is no hint implemented that notifies that some fields are missing when the config gets changed during an evnet

that is another feature pls make an issue for that

tests/Utils/SignUpConfig.php
xuwhite: done

MyIgel:
> that is another feature pls make an issue for that

The notification: Maybe.
the fact that the settings have different requirements on the profile page not so much.

xuwhite:
> > that is another feature pls make an issue for that
> 
> The notification: Maybe. the fact that the settings have different requirements on the profile page not so much -> That's just inconsistent

done

MyIgel (commented):
@xuwhite  I am still getting a "plz configure your tshirt" message even when tshirt is not required, do we want that?
Screenshot_20231103_134938

And as we are on it: should we eventually implement that message for any required field?

PR #1226, opened 2023-10-01
rename rooms to locations

### LINKED ISSUE #1005: Rename `Räume` into `Örtlichkeiten`, `Lokalitäten` or `Standorte`
this would be more general for example something outside

xuwhite:
fixes #1005

### DIFF
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -293,8 +293,8 @@ msgstr ""
 "Diese Schicht wurde aus einem Fahrplan importiert. "
 "Dadurch werden einige Änderungen beim nächsten Import überschrieben."
 
-msgid "Please select a room."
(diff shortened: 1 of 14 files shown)


resources/lang/de_DE/default.po
```diff
@@ -930,9 +901,6 @@ msgstr "Zusätzliche Beschreibung"
 msgid "This description is for single shifts, otherwise please use the description in shift type."
 msgstr "Diese Beschreibung ist für einzelne Schichten, ansonsten nutze bitte die Beschreibung im Schichttyp."
 
-msgid "Please select a location."
-msgstr "Bitte einen Ort auswählen."
```
MyIgel, line 934: This is still used if  see that correctly

xuwhite: but it was double 
its still there in line 296

PR #1232, opened 2023-10-15
removing pt translation

xuwhite:
the pt translation hasn't been updated for a long time and was completely outdated


MyIgel:
Fair point, if anyone wants to translate that again it should be done after most of the current rewrite is completed (and can be based on this file)

PR #1247, opened 2023-11-05
Confirmation popup for deletion

### LINKED ISSUE #1136: Client-side confirmation of actions (e.g. deletion of a shift)
**Problem description:**

Sometimes people have thick fingers and accidentally submit a change or delete something.

**Proposed solution:**

There should be a general client-side function that can be added to forms to ask for confirmation. This issue proposes a JavaScript function that hooks into form submissions if the attribute `data-confirm="text"` is present on a form element.

A confirmation modal should be displayed with the text of the attribute on form submission. [It should use the default Bootstrap modal.](https://getbootstrap.com/docs/5.3/components/modal/) The buttons should be in this order from left to right „No“ (default button) and „Yes“ (primary button). If `data-confirm` is preceded by another word, such as `data-confirm-danger`, this is used as the „Yes“ button type.

**Known problems / scope:**

The solution only works if the user has JavaScript enabled. Target are users with shift coordinator (or greater) power levels. These must have JavaScript enabled to work with the angel system anyway. A server-side solution is out of scope of this issue.

MyIgel:
This PR creates confirmation popups before deleting some entities from their overview pages.

Screenshot_20231105_020628

Closes #1136

### DIFF
--- a/resources/views/pages/news/news.twig
+++ b/resources/views/pages/news/news.twig
@@ -33,7 +33,7 @@
                                 action="{{ url('/news/comment/' ~ comment.id) }}" enctype="multipart/form-data"
                                 method="post">
                                 {{ csrf() }}
-                                {{ f.submit(m.icon('trash'), {'name': 'delete', 'btn_type': 'danger', 'size': 'sm', 'title': __('form.delete')}) }}
+                                {{ f.delete(null, {'size': 'sm', 'confirm_title': comment.text[:40]|e}) }}
                             </form>
                         </div>
                     {% endif %}
--- a/resources/views/pages/news/news.twig
+++ b/resources/views/pages/news/news.twig
@@ -33,7 +33,11 @@
                                 action="{{ url('/news/comment/' ~ comment.id) }}" enctype="multipart/form-data"
                                 method="post">
                                 {{ csrf() }}
-                                {{ f.delete(null, {'size': 'sm', 'confirm_title': comment.text[:40]|e}) }}
+                                {{ f.delete(null, {
+                                    'size': 'sm',
+                                    'confirm_title': __('news.comments.delete.title', [comment.text[:40]|e]),
+                                    'confirm_button_text': __('form.delete'),
+                                }) }}
                             </form>
                         </div>
                     {% endif %}
--- a/resources/views/pages/questions/overview.twig
+++ b/resources/views/pages/questions/overview.twig
@@ -54,7 +54,7 @@
                                                 >
                                                     {{ csrf() }}
                                                     {{ f.hidden('id', question.id) }}
-                                                    {{ f.submit(m.icon('trash'), {'name': 'delete', 'btn_type': 'danger', 'size': 'sm', 'title': __('form.delete')}) }}
+                                                    {{ f.delete(null, {'size': 'sm', 'confirm_title': question.text[:40]|e}) }}
                                                 </form>
                                             {% endif %}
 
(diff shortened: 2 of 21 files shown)


xuwhite:
> This PR creates confirmation popups before deleting some entities from their overview pages.
> 
> Screenshot_20231105_020628

I think it would look better if you aline the two buttons

resources/views/pages/news/news.twig
```diff
@@ -33,7 +33,7 @@
                                 action="{{ url('/news/comment/' ~ comment.id) }}" enctype="multipart/form-data"
                                 method="post">
                                 {{ csrf() }}
-                                {{ f.submit(m.icon('trash'), {'name': 'delete', 'btn_type': 'danger', 'size': 'sm', 'title': __('form.delete')}) }}
+                                {{ f.delete(null, {'size': 'sm', 'confirm_title': comment.text[:40]|e}) }}
```
xuwhite, line 36: Ive tried it and in news it doesnt work for me at the moment

resources/views/pages/questions/overview.twig
```diff
@@ -54,7 +54,7 @@
                                                 >
                                                     {{ csrf() }}
                                                     {{ f.hidden('id', question.id) }}
-                                                    {{ f.submit(m.icon('trash'), {'name': 'delete', 'btn_type': 'danger', 'size': 'sm', 'title': __('form.delete')}) }}
+                                                    {{ f.delete(null, {'size': 'sm', 'confirm_title': question.text[:40]|e}) }}
```
xuwhite, line 57: it would be nice if it would show which type I am deleting (here Question ...)

resources/views/admin/rooms/index.twig
```diff
@@ -57,7 +57,7 @@
                                             <form method="post" class="ps-1">
                                                 {{ csrf() }}
                                                 {{ f.hidden('id', room.id) }}
-                                                {{ f.button(m.icon('trash'), {'title': __('form.delete'), 'name': 'delete', 'type': 'submit', 'btn_type': 'danger', 'size': 'sm'}) }}
+                                                {{ f.delete(null, {'size': 'sm','confirm_title': room.name|e}) }}
```
xuwhite, line 60: it would be nice if it would show which type I am deleting (here Location ...)

xuwhite: the popup should also pop up if xout try to delete it in the edit view

resources/views/pages/questions/overview.twig
xuwhite: the popup should also pop up if xout try to delete it in the edit view

xuwhite (changes requested):

xuwhite:
the delete worklog page should also be a popup for consistency

xuwhite:
instead of the trashcan i would like "delete" better

MyIgel:
Only remaining task here could be to show what type of something gets deleted but i don't think that would be that useful as you normally know that "Saal 1" is a room, "Angel" an angeltype and "Please get your shirts while they are hot" a news

xuwhite:
> Only remaining task here could be to show what type of something gets deleted but i don't think that would be that useful as you normally know that "Saal 1" is a room, "Angel" an angeltype and "Please get your shirts while they are hot" a news

It's not always that clear and I think it would improve usability and make more clear what impact the deletion could have

weeman1337 (approved):
Looks good :beers:  tested and works :rocket:

PR #1254, opened 2023-11-12
disable needed angels input if shift is imported

xuwhite:
while editing a shift if the shift is imported from a schedule it should not be possible to change the amount of angels needed because that breaks the Engelsystem


MyIgel:
"breaking" here means that a) the needed angels might be overwritten when re-importing the schedule and b) the current implementation of which angeltype is loaded when is not that solid so it might just assume that it must be loaded from a room while another code loads them first from the needed angeltypes list

PR #1256, opened 2023-11-13
refactor page_link_to() to url()

### LINKED ISSUE #1253: Change page_link_to  to url
Change `page_link_to($path, $args)` to `url($path, $args)` as the old function only replaces `_` with `-` (which have to be done statically here).
Also fixing URLs that don't start with a `/` would be great.

xuwhite:
fixes #1253

### DIFF
--- a/includes/view/UserAngelTypes_view.php
+++ b/includes/view/UserAngelTypes_view.php
@@ -25,12 +25,12 @@ function UserAngelType_update_view(UserAngelType $user_angeltype, User $user, An
         form([
             buttons([
                 button(
-                    page_link_to('angeltypes', ['action' => 'view', 'angeltype_id' => $angeltype->id]),
+                    url('/angeltypes', ['action' => 'view', 'angeltype_id' => $angeltype->id]),
                     icon('x-lg') . __('form.cancel')
                 ),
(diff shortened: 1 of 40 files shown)


resources/assets/themes/theme13.scss
```diff
@@ -32,15 +32,15 @@ $link-color: #fff;
 
 @font-face {
   font-family: 'Montserrat';
-  src: url('theme13/Montserrat-Regular.woff2') format('woff2'), url('theme13/Montserrat-Regular.woff') format('woff');
+  src: url('/theme13/Montserrat-Regular.woff2') format('woff2'), url('/theme13/Montserrat-Regular.woff') format('woff');
```
MyIgel, line 35: too much replaced (see also following lines + files)

includes/view/UserAngelTypes_view.php
```diff
@@ -50,15 +50,15 @@ function UserAngelTypes_delete_all_view(AngelType $angeltype)
         form([
             buttons([
                 button(
-                    page_link_to(
+                    url(
                         'angeltypes',
```
MyIgel, line 54: Missing `/` ;)

MyIgel (changes requested):

PR #1264, opened 2023-11-24
Signup advance hours for shift types

MyIgel:
Implements parts of #372: Make signup hours in advance configurable by shift type (useful for example for bar shifts which are already full long before the average angel arrives on location).
It also contains some cleanup on migrations ;)


xuwhite:
migration is broken please fix

PR #1270, opened 2023-12-06
Fix pages & translation

### LINKED ISSUE #1268: <title> in news contains raw html
Hi,
it looks like #1245 introduced a bug adding html to the `<title>` element.
image
image

### LINKED ISSUE #1276: Safari doesnt allow secure cookies in localhost
in the localhost dev environment cookies are set in secure flag 
Safari seems to not allow that

MyIgel:
* Fix html in page titles, fixes #1268
* Fix broken translations & added CI and pre-commit hooks to check them
* Fix secure cookie when using http: fixes #1276

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -155,6 +155,15 @@ yarn lint:
     - apk add --no-cache git
     - yarn lint
 
+translations lint:
+  image: alpine
+  stage: prepare
+  before_script:
+    - apk add gettext
+  script:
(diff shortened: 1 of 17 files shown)


.gitlab-ci.yml
```diff
@@ -155,6 +155,15 @@ yarn lint:
     - apk add --no-cache git
     - yarn lint
 
+translations lint:
+  image: alpine
```
marudor, line 159: Using an explicit version would be advised. But considering none of the images are pinned to a version *shrug*

marudor (approved):

weeman1337 (approved):
Tested and works :+1:

PR #1272, opened 2023-12-08
Implement schedule deletion

### LINKED ISSUE #867: Imports should be disabled/deleted
Currently, there is no possibility to remove an import (e.g. if the fahrplan is merged). There should be an option to deactivate or remove those imports.

MyIgel:
fixes #867


weeman1337 (approved):
Tested as far as I can. Looks good

PR #1273, opened 2023-12-08
show user-info icon with admin_arrive permission

xuwhite:


includes/view/User_view.php
```diff
@@ -934,17 +941,20 @@ function render_user_freeloader_hint()
 }
 
 /**
- * Hinweis für Engel, die noch nicht angekommen sind
+ * hint for angels, which are noch arrived yet
```
MyIgel, line 944: u sure?

includes/view/User_view.php
```diff
@@ -92,7 +92,11 @@ function Users_view(
     $usersList = [];
     foreach ($users as $user) {
         $u = [];
-        $u['name'] = User_Nick_render($user) . User_Pronoun_render($user);
+        $u['name'] = User_Nick_render($user)
+            . User_Pronoun_render($user)
+            . ($user->state->user_info
+                ? ' <small><span class="bi bi-info-circle-fill text-info" data-bs-toggle="tooltip" title=""></span></small>'
```
MyIgel, line 98: `data-bs-toggle="tooltip" title="` not required here either

MyIgel (changes requested):
There are some errors, see below. Also the `please arrive for signup` text on shifts should be imho changed depending on the comment

PR #1275, opened 2023-12-12
highlight actual time in user myshifts view

xuwhite:
for example during a running shift
Screenshot 2023-12-12 at 14 13 18


MyIgel (approved):
Looks good to me (even if a little bit hacky), @weeman1337 do you have an idea how to make the css a bit prettier? :D

PR #1279, opened 2023-12-16
Icons

xuwhite:


resources/views/admin/locations/index.twig
```diff
@@ -8,7 +8,9 @@
     <div class="container">
         <h1>
             {% if not is_index|default(false) %}
-                {{ m.button(m.icon('chevron-left'), location ? url('/locations', {'action': 'view', 'location_id': location.id}) : url('/admin/locations'), 'secondary', 'sm') }}
+                {{ m.button(m.icon('chevron-left'), location
+                    ? url('/locations', {'action': 'view', 'location_id': location.id})
+                    : url('/admin/locations'), 'secondary', 'sm', __('Back')) }}
```
MyIgel, line 13: This (and all following) should be translation strings like `general.back`

resources/views/pages/design.twig
```diff
@@ -159,6 +159,68 @@
...
+            <div class="row">
+                <div class="col-md-6 col-lg-4 mb-4">
+                    cancel: {{ m.icon('x-lg') }}<br>
+                    delete: {{ m.icon('trash', 'danger') }}<br>
+                    edit: {{ m.icon('pencil') }}<br>
+                    preview: {{ m.icon('eye','info') }}<br>
+                    reset: {{ m.icon('arrow-counterclockwise') }}<br>
+                    save: {{ m.icon('save', 'primary') }}<br>
+                    <br>
+                    {{ f.button('Cancel', {'icon_left': 'x-lg'}) }}<br>
```
MyIgel, line 174: You could add a `class="mb-3"` to the br so that it has better spacing

MyIgel (changes requested):
That looks great! <3

Eventually the "delete all sessions" button should have `'title': ' '` set and two smaller things listed below:

resources/views/pages/design.twig
xuwhite: wont do

PR #1280, opened 2023-12-16
Implement random shift button

MyIgel:

PR #1285, opened 2023-12-20
Allow loading angels by shift type on schedule import

### LINKED ISSUE #649: Frab import changes to support Translation shifts
Hi, I had a discussion at Camp with @MyIgel about what the Translation angels would need in order to manage our shifts in the Engelsystem. On the technical side, the only blocker is related to the Frab import. Per our discussion, here's a list of what we would need:

* The Frab import currently can only be run by one group: it creates 1 shift for each event in the Fahrplan, with a count of angels configured per-room, and a before/after time buffer common to all. Translation team needs to reflect prep time in our shift, which differs from the 15/15 used today for room operation shifts. Ideally we'd like to run the import just to create Translation shifts, independently from the VOC who runs the import on their side to create the "room operation" shifts as they do today.

  I think a clean way to do this would be to configure each shifttype to be linked to N angeltypes, and create only shifts for these angeltypes during the Frab import. This way we would have a "Translation" shift type, enabling shifts for Translation angels, and VOC would keep the "Room Operation" shift type, enabling shifts for Camera, Signal, Crowd Control, etc. angels. (Note that shifttypes are already linked to 1 and only 1 angeltype but I have not found where that is used: it does not impact the Frab import).

* Our prep time is generally speaking a function of the talk length. Most of that prep doesn't necessarily happen during the shift, but we want to reflect it in the shift duration anyway. The cleanest way I can think of this would be:
  * Create our shifts with a lead time of 15min (because that's when we want folks in the cabin)
  * Add a buffer of `talk length * ratio - 15min` at the end, where `ratio` would be configurable

  If this is too complicated, we're open to suggestions on how to best approximate this.

I'm happy to bring clarifications on any of this. Thanks for all your help!

MyIgel:
* Load angels from shift type instead of room if needed
  * Cleanup of shift loading code to make it a bit clearer
* Added shift update event to send templated emails

Definitely has side effects with #1282, #1277 and #1280

fixes #649


msquare (approved):
Looks fine. 🔥

PR #1289, opened 2023-12-22
Show "not marked as arrived" Notification only from set arrival date

### LINKED ISSUE #1288: Show "You are not marked as arrived" Notification only from set arrival date and not before
When a user has set an arrival date, don't display the "You are not marked as arrived" hint before this date.

xuwhite:
fixes #1288

### DIFF
--- a/includes/sys_menu.php
+++ b/includes/sys_menu.php
@@ -24,7 +24,7 @@ function header_render_hints()
 
         // Important hints:
         $hints_renderer->addHint(render_user_freeloader_hint(), true);
-        $hints_renderer->addHint(render_user_arrived_hint(), true);
+        $hints_renderer->addHint(render_user_arrived_hint(true), true);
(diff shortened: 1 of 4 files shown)


includes/sys_menu.php
```diff
@@ -24,7 +25,9 @@ function header_render_hints()
 
         // Important hints:
         $hints_renderer->addHint(render_user_freeloader_hint(), true);
-        $hints_renderer->addHint(render_user_arrived_hint(), true);
+        if ($user->personalData->planned_arrival_date && (Carbon::now() > $user->personalData->planned_arrival_date)) {
+            $hints_renderer->addHint(render_user_arrived_hint(), true);
+        };
```
msquare, line 30: The logic if a hint has to be displayed has to be in the render_user_arrived_hint(). These methods decide if the hint has to be rendered.

PR #1290, opened 2023-12-23
Further API fixes & improvements

MyIgel:
* Fixed event time..
* Documented `x-api-key` header for authentication

### DIFF
--- a/db/migrations/2023_05_21_000001_cleanup_short_api_keys.php
+++ b/db/migrations/2023_05_21_000001_cleanup_short_api_keys.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Engelsystem\Migrations;
 
 use Engelsystem\Database\Migration\Migration;
+use Illuminate\Support\Str;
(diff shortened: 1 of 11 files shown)


xuwhite:
put put put

xuwhite:
pls add doku

xuwhite (changes requested):
Barer auth doesnt work for me

MyIgel:
The bearer auth now also trims any additional spaces added and thus should resolve your issue (also added `access-control-allow-origin` to not authorized responses so that requesting applications can properly react instead of triggering browser protection mechanics)

xuwhite (changes requested):
pls fix that on a new angelsystem you have to reset the api key first before getting an api key

MyIgel:
That is only relevant for the admin user and thus another issue that should be solved on user creation -> @xuwhite  can you open a new issue for that?

weeman1337 (approved):
Looks good

PR #1292, opened 2023-12-23
Add tags to FAQ items

MyIgel:
Allows faq items to be grouped & linked to sub-sections.

Also fixed twig filter tests & added slug filter


xuwhite:
Tags im Deutschen dur Kategorien ersetzten?

xuwhite:
Screenshot 2024-02-17 at 17 32 15
eine überschrift über den Tags wäre sinnvoll

xuwhite:
pls delete old unused tags

MyIgel:
> pls delete old unused tags

@xuwhite  thats imho a new feature

xuwhite:
> > pls delete old unused tags
> 
> @xuwhite thats imho a new feature

I dont think so!!!

MyIgel:
The point with deleting unused tags is that they might be used in the future for example for news too (and other things eventually).
Implementing that feature here would mean doing double the work instead of implementing it the proper way with some "tag manager" page in the future which i do prefer -> You can create a new issue for that if you want :)

PR #1306, opened 2023-12-27
small bug fixes

xuwhite:
"You have done enogh" user or admin
shift creation now 1 log
all angel pencil button fix

### DIFF
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -1187,8 +1187,14 @@ msgstr "Dauer"
 msgid "Name & Workmates"
 msgstr "Name & Kollegen"
 
-msgid "You have done enough."
-msgstr "Du hast genug gemacht."
+msgid "myshifts.sucess"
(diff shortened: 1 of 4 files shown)


resources/lang/de_DE/default.po
```diff
@@ -1187,8 +1187,14 @@ msgstr "Dauer"
 msgid "Name & Workmates"
 msgstr "Name & Kollegen"
 
-msgid "You have done enough."
-msgstr "Du hast genug gemacht."
+msgid "myshifts.sucess"
+msgstr "%s genug gemacht."
```
MyIgel, line 1191: That is some magical translation mushing which i can not recommend as it creates issues when only partial transslations get translated (and does not make sense for some other languages) -> please translate whole sentences / output groups at once!

MyIgel: as in: Use `%s hat genug gemacht` and `Du hast genug gemacht` explicitly

PR #1313, opened 2023-12-28
37c3 Theme

weeman1337:
image


MyIgel:
Known issues: Disabled selects have the wrong color

PR #1327, opened 2024-01-05
show users with info as arrived

MyIgel:
This "improves" the workaround introduced in the 37C3 by making the output more consistent.

The interesting question here is if we have more "information leaks" somewhere else when implementing this state output.

### DIFF
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -735,7 +735,7 @@ function User_view_state_user($user_source)
         User_shift_state_render($user_source),
     ];
 
-    if ($user_source->state->arrived) {
+    if ($user_source->state->arrived || $user_source->state->user_info) {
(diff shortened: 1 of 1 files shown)


includes/view/User_view.php
```diff
@@ -899,7 +899,7 @@ function User_Nick_render($user, $plain = false)
     return render_profile_link(
         '<span class="icon-icon_angel"></span> ' . htmlspecialchars($user->displayName) . '</a>',
         $user->id,
-        ($user->state->arrived ? '' : 'text-muted')
+        ($user->state->arrived || $user->state->user_info ? '' : 'text-muted')
```
xuwhite, line 902:
```suggestion
        ($user->state->arrived ? '' : 'text-muted')
```

xuwhite (commented):
I would like it that way a bit more

PR #1336, opened 2024-01-16
note that shift length is in minutes, extend `form_text()` to do this nicely

Lotterleben:
As a user, when adding shifts, it wasn’t quite clear to me in which unit the shift length would be provided.

This PR clears this up by adding a little “in minutes” note under the text form. To make this note less obtrusive and create a visual information hierarchy, this PR also extends the `form_text()` function with a new `description()` parameter that lets us add such information as an optional [form text](https://getbootstrap.com/docs/5.3/forms/form-control/#form-text).

this is my first time writing PHP, so I hope what I did makes sense…

formatting note:
I’ve set up my editor to format according to PSR-12 on save, which messes with the existing indentation as you can see. However, `phpcs` does not seem to find a fault with this, so I’m not sure which is wrong – my editor settings or the pre-existing indentation

### DIFF
--- a/includes/pages/admin_free.php
+++ b/includes/pages/admin_free.php
@@ -117,7 +117,7 @@ function admin_free()
             div('row', [
                 div('col-md-12 form-inline', [
                     div('row', [
-                        form_text('search', __('form.search'), $search, '', null, null, null, 'col', []),
+                        form_text('search', __('form.search'), $search, null, null, null, 'col'),
                         form_select('angeltype', __('Angeltype'), $angel_types, $angelType, '', 'col'),
                         form_submit('submit', icon('search') . __('form.search')),
                     ]),
--- a/includes/pages/admin_free.php
+++ b/includes/pages/admin_free.php
@@ -117,7 +117,7 @@ function admin_free()
             div('row', [
                 div('col-md-12 form-inline', [
                     div('row', [
-                        form_text('search', __('form.search'), $search, null, false, null, null, null, ['col']),
+                        form_text('search', __('form.search'), $search, '', null, null, null, 'col', []),
                         form_select('angeltype', __('Angeltype'), $angel_types, $angelType, '', 'col'),
                         form_submit('submit', icon('search') . __('form.search')),
                     ]),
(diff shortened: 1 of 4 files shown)


includes/pages/admin_free.php
```diff
@@ -117,7 +117,7 @@ function admin_free()
             div('row', [
                 div('col-md-12 form-inline', [
                     div('row', [
-                        form_text('search', __('form.search'), $search, null, null, null, 'col'),
+                        form_text('search', __('form.search'), $search, null, false, null, null, null, ['col']),
```
Lotterleben, line 120: this is a bit of a drive-by fix: when I checked all `form_text()` calls, I noticed that this call was missing an another argument in addition to the newly introduced one and that the last argument had a type mismatch. So I updated that on the go – the page looks fine in my browser, but that's the full extent of my testing. so, happy to reverse if desired.

includes/pages/admin_shifts.php
```diff
@@ -564,8 +565,9 @@ function admin_shifts()
                             'change_hours',
                             __('Shift change hours'),
                             $request->has('change_hours')
-                                ? ($change_hours ? implode(', ', $change_hours) : $request->input('change_hours'))
-                                : '00, 04, 08, 10, 12, 14, 16, 18, 20, 22',
+                            ? ($change_hours ? implode(', ', $change_hours) : $request->input('change_hours'))
+                            : '00, 04, 08, 10, 12, 14, 16, 18, 20, 22',
+                            null,
```
xuwhite, line 570:
```suggestion
                            '',
```
should be the default value

xuwhite (changes requested):
Thanks for your pull request 
I have some smallish request changes

xuwhite:
Overall I would prefer having the hint 'in minutes' in the title of the input field like 'Length (in minutes)'

Lotterleben:
> Overall I would prefer having the hint 'in minutes' in the title of the input field like 'Length (in minutes)'

Makes sense, I can also do that! Want me to also remove the change to `form_text()` to avoid unused code or keep it for future use?

includes/pages/admin_free.php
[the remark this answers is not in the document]
Lotterleben: > 'col' is a class not a data-attribute please reverse that
ah, yup, I got the count wrong – what was missing is the last argument, i.e. `$data_attributes`. I've set this to `[]` now, hope that makes sense.

xuwhite:
Please remove it

PR #1340, opened 2024-01-22
Improvements

xuwhite:
- shifts_view hint if no self signout possible
- shifts_view disable signout buttons if no self signout possible
- button() is now able to be disabled
- allow supporter to add other supporters with config option
- User view Pronoun after name
- force active is a config option

### DIFF
--- a/includes/pages/user_myshifts.php
+++ b/includes/pages/user_myshifts.php
@@ -20,10 +20,18 @@ function user_myshifts()
 {
     $user = auth()->user();
     $request = request();
+    $is_angeltype_supporter = false;
+    if ($request->has('edit')) {
+        $id = $request->input('edit');
+        $shiftEntry = ShiftEntry::where('id', $id)
+            ->where('user_id', User::find($request->input('id'))->id)
+            ->first();
+        $is_angeltype_supporter = $shiftEntry && auth()->user()->isAngelTypeSupporter($shiftEntry->angelType);
+    }
 
     if (
         $request->has('id')
-        && auth()->can('user_shifts_admin')
+        && (auth()->can('user_shifts_admin') || $is_angeltype_supporter)
         && preg_match('/^\d+$/', $request->input('id'))
         && User::find($request->input('id'))
     ) {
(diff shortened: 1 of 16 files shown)


includes/view/Shifts_view.php
```diff
@@ -166,7 +167,10 @@ function Shift_view(
     }
 
     if ($shift_signup_state->getState() === ShiftSignupStatus::SIGNED_UP) {
-        $content[] = info(__('You are signed up for this shift.'), true);
+        $content[] = info(__('You are signed up for this shift.')
+            . (($shift->start->subHours(config('last_unsubscribe')) < Carbon::now() && $shift->end > Carbon::now())
+                ? ' ' . __('shift.sign_out.hint', config('last_unsubscribe'))
```
MyIgel, line 172:
```
 TypeError
__(): Argument #2 ($replace) must be of type array, int given, called in /var/www/includes/view/Shifts_view.php on line 172
```

includes/pages/user_myshifts.php
```diff
@@ -20,10 +20,19 @@ function user_myshifts()
 {
     $user = auth()->user();
     $request = request();
+    $is_angeltype_supporter = false;
+    if ($request->has('edit')) {
+        $id = $request->input('edit');
+        $shiftEntry = ShiftEntry::where('id', $id)
+            ->where('user_id', User::find($request->input('id'))->id)
+            ->with(['shift', 'user'])
```
MyIgel, line 28:
```suggestion
            ->with(['angelType'])
```
should be enough here but can also be completely omitted as its only one model
```suggestion
```

includes/pages/user_myshifts.php
```diff
@@ -20,10 +20,19 @@ function user_myshifts()
...
+            ->first();
+        $is_angeltype_supporter = auth()->user()->isAngelTypeSupporter($shiftEntry->angelType);
```
MyIgel, line 30: We should check here if the shift entry really exists:
```
Attempt to read property "angelType" on null
```

```suggestion
        $is_angeltype_supporter = $shiftEntry && auth()->user()->isAngelTypeSupporter($shiftEntry->angelType);
```

MyIgel (changes requested):
Looks great, i have just some smaller changes / bugs.
Additionally, the `User_edit_vouchers_view` should too contain the force active check.

PR #1343, opened 2024-01-27
Move shifts, news, answers, worklogs etc. to deleting user

MyIgel:
* User deletion: Move shifts, news, answers, worklogs etc. to deleting user: fixes 1335
* User model: Add logs & LogEntry factory

### DIFF
--- a/db/factories/LogEntryFactory.php
+++ b/db/factories/LogEntryFactory.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Database\Factories\Engelsystem\Models;
+
+use Engelsystem\Helpers\Carbon;
+use Engelsystem\Models\LogEntry;
+use Engelsystem\Models\User\User;
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Psr\Log\LogLevel;
+
+class LogEntryFactory extends Factory
+{
+    /** @var string */
+    protected $model = LogEntry::class; // phpcs:ignore
+
+    public function definition(): array
+    {
+        return [
+            'user_id' => User::factory(),
+            'level' => $this->faker->randomElement([
+                    LogLevel::EMERGENCY,
+                    LogLevel::ALERT,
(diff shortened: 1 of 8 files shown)


db/factories/LogEntryFactory.php
```diff
@@ -0,0 +1,37 @@
...
+                    LogLevel::EMERGENCY,
+                    LogLevel::ALERT,
+                    LogLevel::CRITICAL,
+                    LogLevel::ERROR,
+                    LogLevel::WARNING,
+                    LogLevel::NOTICE,
+                    LogLevel::INFO,
+                    LogLevel::DEBUG,
+                ]
+            ),
```
xuwhite, line 32: phpcs doesn't like this

includes/controller/users_controller.php
```diff
@@ -79,6 +79,13 @@ function user_delete_controller()
         }
 
         if ($valid) {
+            // Move user created news/answers/worklogs/shifts  etc. to deleting user
+            $user_source->news()->update(['user_id' => $user->id]);
+            $user_source->questionsAnswered()->update(['answerer_id' => $user->id]);
+            $user_source->worklogsCreated()->update(['creator_id' => $user->id]);
+            $user_source->shiftsCreated()->update(['created_by' => $user->id]);
+            $user_source->shiftsUpdated()->update(['updated_by' => $user->id]);
+
```
xuwhite, line 88: logs could be reassigned too
confirmed_id of user angel types is missing

xuwhite (changes requested):
looks good, I request some minor changes

includes/controller/users_controller.php
MyIgel: Having the logs there too would, imho, move too much "other users content" over so i don't think that this is a good idea

db/factories/LogEntryFactory.php
MyIgel: Fixed

PR #1346, opened 2024-01-29
Enable modal confirmation with enter

### LINKED ISSUE #1345: Confirmation modal should be confirmable by enter key
When a dialogue opens to confirm something like a deletion, it should be possible to use `enter` to confirm it and thus doing the original action like deleting news. (at the moment its only possible to close it using `Esc`)

Screenshot of the modal confirmation dialog

weeman1337:
- Tweaks the design page to be able to debug the confirmation modal
- Focus the button, after the modal has been shown

closes #1345

### DIFF
--- a/resources/assets/js/forms.js
+++ b/resources/assets/js/forms.js
@@ -318,6 +318,7 @@ ready(() => {
                 </div>
                 <div class="modal-footer">
                   <button type="button" class="${element.className}"
+                    autofocus
                     title="${element.title}" data-submit="">
                     ${element.dataset.confirm_button_text ?? element.innerHTML}
                   </button>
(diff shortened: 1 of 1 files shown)


resources/assets/js/design.js
```diff
@@ -10,4 +10,9 @@ ready(() => {
     element.addEventListener('submit', preventDefault);
     element.addEventListener('click', preventDefault);
   });
+
+  document.getElementById('delete-form').addEventListener('submit', (event) => {
```
MyIgel, line 14:
```
Uncaught TypeError: document.getElementById(...) is null
    js design.js:14
    ready ready.js:6
    js design.js:3
    Webpack 4
[design.js:14:11](webpack://engelsystem/resources/assets/js/design.js)
    js design.js:14
    (Async: EventListener.handleEvent)
    ready ready.js:6
    js design.js:3
    Webpack 4
```

on any page which not the design page

```suggestion
  document.getElementById('delete-form')?.addEventListener('submit', (event) => {
```

MyIgel (changes requested):

resources/assets/js/design.js
weeman1337: I would rather say that the `design.js` file should not be loaded on non-design pages...

MyIgel: Thats an idea that can be implemented eventually but at the moment that's not the case :D

MyIgel: & its fixed

PR #1353, opened 2024-02-18
some smaller fixes

xuwhite:
- add target="_blank" rel="noopener" to links
- add number of emails to be send to news edit page


includes/pages/user_shifts.php
```diff
@@ -374,7 +374,7 @@ function ical_hint()
 
     return heading(__('iCal export and API') . ' ' . button_help('user/ical'), 2)
         . '<p>' . sprintf(
-            __('Export your own shifts. <a href="%s">iCal format</a> or <a href="%s">JSON format</a> available (please keep secret, otherwise <a href="%s">reset the api key</a>).'),
+            __('Export your own shifts. <a href="%s" target="_blank" rel="noopener">iCal format</a> or <a href="%s" target="_blank" rel="noopener">JSON format</a> available (please keep secret, otherwise <a href="%s" target="_blank" rel="noopener">reset the api key</a>).'),
```
MyIgel, line 377: In that case the noopener is not needed as they link to the same site ^^

Also the api key reset should not have a target=_blank as its the same page and i don't see a usecase where it would help to open it in another tab?

src/Controllers/Admin/NewsController.php
```diff
@@ -52,6 +54,7 @@ protected function showEdit(?News $news, bool $sendNotification = true, bool $is
                 'is_pinned'      => $news ? $news->is_pinned : false,
                 'is_highlighted' => $news ? $news->is_highlighted : false,
                 'send_notification' => $sendNotification,
+                'mails' => $mails,
```
MyIgel, line 57: Eventually `notifications_count` would be clearer?

MyIgel (changes requested):

includes/pages/user_shifts.php
xuwhite: done

src/Controllers/Admin/NewsController.php
xuwhite: done

resources/lang/de_DE/default.po
```diff
@@ -837,12 +837,12 @@ msgid "iCal export and API"
 msgstr "iCal Export und API"
 
 msgid ""
-"Export your own shifts. <a href=\"%s\">iCal format</a> or <a href=\"%s"
-"\">JSON format</a> available (please keep secret, otherwise <a href=\"%s"
-"\">reset the api key</a>)."
+"Export your own shifts. <a href=\"%s\" target=\"_blank\">iCal format</a> or <a href=\"%s"
+"\" target=\"_blank\">JSON format</a> available (please keep secret, otherwise <a href=\"%s"
+"\" target=\">reset the api key</a>)."
```
MyIgel, line 842: I don't think that this is valid HTML ;)

src/Controllers/Admin/NewsController.php
```diff
@@ -44,6 +45,7 @@ public function edit(Request $request): Response
 
     protected function showEdit(?News $news, bool $sendNotification = true, bool $isMeetingDefault = false): Response
     {
+        $notifications_count = Settings::whereEmailNews(true)->count();
```
MyIgel, line 48: As a variable name you should use `snakeCase`

MyIgel (changes requested):

src/Controllers/Admin/NewsController.php
xuwhite: done

PR #1354, opened 2024-02-18
Fix night shift calculation and smaller fixes

MyIgel:
* Fix night shift calculation
  As long as the calculations are now kept in sync they are test covered too :tada:
* add hours as title to "enough" score
* Shifts view: Highlight current time
* Move footer to the bottom on any page
* Make group rights edit buttons clickable again
* Change translations when removing users from angeltype to be less confusing
* Fix disabled choices select color on dark theme


xuwhite (changes requested):
Please use current night shift default config

PR #1357, opened 2024-02-25
explain freeloads for users

xuwhite:


MyIgel:
@xuwhite this needs a rebase ;)

Imho it would be also great to show an error message on the profile page when the angel visits it as it is the redirect target when opening the shifts page

resources/lang/de_DE/default.po
```diff
@@ -2105,3 +2098,19 @@ msgstr "Es wurde noch kein Engeltyp erstellt. Ohne können keine Schichten erste
...
 "Wenn du nicht zu deiner Schicht kommen kannst, lass dich vom Himmel austragen."
+
+msgid "freeload.info"
+msgstr "Du warst bei dieser Schicht nicht anwesend. "
+"Die doppelte Länge der Schicht wurde von der Summe deiner Stunden abgezogen. "
+"Bitte wende dich bei Fragen an den Himmel."
+
+msgid "freeload.freeloader.info"
+msgstr ""
+"Du warst bei mindestens %s Schichten nicht anwesend. Deshalb ist Schicht-Registrierung "
```
MyIgel, line 2109:
```suggestion
"Du warst bei mindestens %s Schichten nicht anwesend. Deshalb ist die Schicht-Registrierung "
```

resources/lang/en_US/default.po
```diff
@@ -978,3 +978,16 @@ msgstr "No angeltype has been created yet. Shifts can't be created without one."
...
+msgstr "You were not present for this shift. "
+"The double length of the shift was deducted from your total hours. Please contact heaven if you have questions."
+
+msgid "freeload.freeloader.info"
+msgstr "You were not present for at least %s shifts. Therefore, shift registration is blocked. "
+"Please go to heaven to be unlocked again."
+
+msgid "freeload.freeloaded.info"
+msgstr ""
+"If an angel was not present for a shift, the double length of the shift is deducted from the total hours. "
```
MyIgel, line 992:
```suggestion
"If an angel was not present for a shift, the doubled length of the shift is deducted from the total hours. "
```
maybe?

MyIgel (changes requested):

MyIgel:
Screenshot_20240302_155147
@xuwhite looks like the doubling is only effective for the goodie score so that should be somehow fixed too :D

xuwhite:
> Screenshot_20240302_155147 Screenshot_20240302_155340
> 
> @xuwhite looks like the doubling is only effective for the goodie score so that should be somehow fixed too :D

I tried to fix it

resources/lang/en_US/default.po
```diff
@@ -981,3 +981,25 @@ msgstr "No angeltype has been created yet. Shifts can't be created without one."
...
+"Please go to heaven to be unlocked again."
+
+msgid "freeload.freeloaded.info.goodie"
+msgstr ""
+"If an angel was not present for a shift, the double length of the shift is deducted from the %s. "
+"If %s shifts are freeloaded, the shift registration is blocked for the angel."
+
+msgid "freeload.freeloaded.info"
+msgstr ""
+"If an angel was not present for a shift. "
```
MyIgel, line 1004:
```suggestion
"The angel was not present for a shift. "
```

I would skip the if here as its pretty hard to understand (seems to be an unanswered question) -> this would make it easier to parse.

resources/lang/de_DE/default.po
```diff
@@ -2111,3 +2104,28 @@ msgstr "Du kannst dich bis %s Stunden vor dem Start der Schicht austragen. "
...
+
+msgid "freeload.freeloaded.info.goodie"
+msgstr ""
+"War ein Engel bei einer Schicht nicht anwesend, "
+"wird die doppelte Länge der Schicht von dem %s abgezogen. "
+"Bei %s geschwänzten Schichten wird die Schicht-Registration für den Engel gesperrt."
+
+msgid "freeload.freeloaded.info"
+msgstr ""
+"Ein Engel war bei einer Schicht nicht anwesend. "
```
MyIgel, line 2130:
```suggestion
"Der Engel war bei einer Schicht nicht anwesend. "
```

I would skip the "questioning" part here as this makes it easier to parse.

includes/view/User_view.php
```diff
@@ -350,16 +351,27 @@ function User_view_myshift(Shift $shift, $user_source, $its_me)
     }
 
     if ($shift->freeloaded) {
-        $myshift['duration'] = '<p class="text-danger">'
-            . sprintf('%.2f', -($shift->end->timestamp - $shift->start->timestamp) / 3600 * 2) . '&nbsp;h'
```
MyIgel, line 354: The subtraction & doubling should be only removed for non-goodie-configs or it is pretty confusing, especially when combined with night shifts

MyIgel (changes requested):

resources/lang/en_US/default.po
MyIgel: (a not gendered variation might be better but dunno)

resources/lang/de_DE/default.po
MyIgel: (A not gendered variation would be better but dunno)

xuwhite: Engel is not genderd

includes/view/User_view.php
xuwhite: I like it this way more because minus the double length doesnt count for the sum underneath 
night shifts are also not shown with double length

MyIgel: Imho for the goodie none case its fine to just show the min red and cross them.
For goodie / shirt enabled the shown hours imho should indicate the shirt score as the total "real" hours can be easily counted.

So I would suggest having the shirt score value displayed and the "real" hours in round brackets after it like `-4.8h (2.4h)` when the user has missed a shift etc., eventually with a `title="T-Shirt score: -4.8h, 2.4h worked"`.

PR #1359, opened 2024-03-02
fix table on admin_active for length and score

xuwhite:


includes/pages/admin_active.php
```diff
@@ -229,14 +229,30 @@ function admin_active()
             }
         }
 
+        $user_worklogs = $usr->worklogs()
+            ->with(['user', 'creator'])
+            ->get();
+        $shifts = Shifts_by_user($usr->id, auth()->can('user_shifts_admin'));
```
MyIgel, line 235:
```suggestion
        $shifts = Shifts_by_user($usr->id);
```
the permission part can be ignored here as the data is not used.

In general it would be nice to have an SQL statement here that just fetches the users shifts as the `Shifts_by_user` function adds additional informations via joins which makes the response slower (esp. as its run for every user).

includes/pages/admin_active.php
```diff
@@ -229,14 +229,30 @@ function admin_active()
             }
         }
 
+        $user_worklogs = $usr->worklogs()
+            ->with(['user', 'creator'])
```
MyIgel, line 233: That one can be ignored too as its unused data and thus two useless queries for every user

MyIgel (changes requested):

PR #1367, opened 2024-03-18
Add support for file-based docker secrets

nafur:
The recommendation for passwords is to use secrets instead of writing them to compose files or environment variables.
This adds support for `MYSQL_PASSWORD_FILE`, `MAIL_PASSWORD_FILE` and `SETUP_ADMIN_PASSWORD`.

This is in line with https://docs.docker.com/compose/use-secrets/ and works like `mariadb`. Note that this PR only changes the php side. We might want to change the compose file as well like we do over here: https://github.com/froscon/engelsystem/pull/5


MyIgel:
Doing this just for one secret seems odd, wouldn't it be better to have some loader or something that does the file reading on the fly?

nafur:
> Doing this just for one secret seems odd, wouldn't it be better to have some loader or something that does the file reading on the fly?

You think of adding it for the mail password as well, and introducing a new utility that manages the precedence logic? Fine with me. Where should I put this utility?

MyIgel:
Yes i mean adding some utilitarian wrapper for it :D
(Which should be located in the `helpers.php` file then)

Imho we could add it directly to the `env` function itself so that its universally useable for all secrets? (and there could also be a check that the file really exists before using it)

In addition it might be a good idea to also have a look how other projects handle the "magical" loading of files here, do you have an overview over that?

nafur:
I moved the wrapper to `helpers.php` and check whether the file exists. I'm not sure where `env` even comes from.
I've just seen this file secrets in the `mariadb` container and it seems like a good idea.

Also, I've actually tested it now :-)

MyIgel:
The `env` function is defined in `vendor/illuminate/support/helpers.php` and could be overwritten to directly use `Env::get`.
Speaking of tests: any code in he `src` directory must have unit tests ;)

nafur:
I added a unit test. Given that we use this `env()` everywhere else, it seems more consistent to keep using it.

nafur:
@MyIgel can we make some progress here? :-)

MyIgel:
Hi @nafur sure we can, i had some other ToDo's, so maintenance was pretty low on the List :D
This branch needs a rebase and there are some linter errors that need to be fixed, that would be the start.
I don't see the consistency argument here but sure, I'll redo it later at some point :shrug:

src/helpers.php
```diff
@@ -73,6 +73,24 @@ function config_path(string $path = ''): string
...
+ * called `${var}_FILE` is set, read the password from that file. Otherwise
+ * returns the content of the `$var` environment variable.
+ */
+function env_password(string $var): string|null {
+    $filename = env("{$var}_FILE", null);
+    if ($filename) {
+        if (file_exists($filename)) {
+            return file_get_contents($filename);
+        } else {
+            echo "The password file $filename does not exist. Please fix your config.";
```
MyIgel, line 87: This line is not covered by unit tests, throwing a `RuntimeException` would be the better (testable) way with the same outcome.

MyIgel (changes requested):

src/helpers.php
MyIgel: (Or eventually better: Just ignore the file if it can't be read)

Btw, using variables in `"` strings is not allowed according to our linter && it would be helpful to still being able to set a default value for `env(...)`

PR #1374, opened 2024-03-27
add 'enable_email_goody' config and fix wording

xuwhite:


resources/views/pages/registration.twig
```diff
@@ -193,10 +193,11 @@
         <div class="mb-5">
             <h2>{{ __('registration.event_data') }}</h2>
             <div class="row">
-                {% if isGoodieEnabled %}
+                {% if isGoodieEnabled and config('enable_email_goody') %}
```
MyIgel, line 196: Imho that check should be done in the controller

resources/views/pages/settings/profile.twig
```diff
@@ -133,9 +133,10 @@
                 {{ f.checkbox('email_human', __('settings.profile.email_by_human_allowed'), {
                     'checked': userdata.settings.email_human,
                 }) }}
-                {% if goodie_enabled %}
+                {% if goodie_enabled and config('enable_email_goody') %}
```
MyIgel, line 136: (same here and it has side effects on the saving logic too)

config/config.default.php
```diff
@@ -88,6 +88,10 @@
     # Your privacy@ contact address
     'privacy_email' => env('PRIVACY_EMAIL', null),
 
+    # Enable checkbox to consent for storage of nick, email address and worked hours (and T-shirt size)
```
MyIgel, line 91: I would prefer a shorter text here as its currently a copy of the whole translation

MyIgel (changes requested):

config/config.default.php
xuwhite: do you have suggestions

MyIgel:
```suggestion
    // Show opt in to long term saving of personal data on user profile and registration pages
```

src/Controllers/SettingsController.php
```diff
@@ -46,7 +46,8 @@ public function profile(): Response
             [
                 'settings_menu' => $this->settingsMenu(),
                 'userdata' => $user,
-                'goodie_enabled' => $this->config->get('goodie_type') !== GoodieType::None->value,
+                'goodie_enabled' => ($this->config->get('goodie_type') !== GoodieType::None->value)
```
MyIgel, line 49:
```suggestion
                'goodie_enabled' => $this->config->get('goodie_type') !== GoodieType::None->value
```

src/Controllers/RegistrationController.php
```diff
@@ -99,7 +99,7 @@ private function renderSignUpPage(): Response
                 'isPasswordEnabled' => $this->userFactory->determineIsPasswordEnabled(),
                 'isDECTEnabled' => $this->config->get('enable_dect'),
                 'isShowMobileEnabled' => $this->config->get('enable_mobile_show'),
-                'isGoodieEnabled' => $goodieType !== GoodieType::None,
+                'isGoodieEnabled' => ($goodieType !== GoodieType::None) && config('enable_email_goody'),
```
MyIgel, line 102:
```suggestion
                'isGoodieEnabled' => $goodieType !== GoodieType::None && config('enable_email_goody'),
```
does not spark joy ;)

PR #1384, opened 2024-04-16
Fixes

xuwhite:
- angeltype view member dect to link
- throw error on goodie page if goodie is disabled
- fix schedule logs


src/Controllers/Admin/UserGoodieController.php
```diff
@@ -35,8 +36,16 @@ public function __construct(
     ) {
     }
 
+    private function checkActive(): void
```
MyIgel, line 39: This function is not mentioned in any test

MyIgel (changes requested):

PR #1389, opened 2024-04-20
refactor permissions and groups

xuwhite:
based on:
Screenshot 2024-04-20 at 22 31 35

Legend 
red - delete
green - insert
yellow - do something else

### DIFF
--- a/db/migrations/2024_04_20_000000_refactor_permissions_and_groups.php
+++ b/db/migrations/2024_04_20_000000_refactor_permissions_and_groups.php
@@ -1,0 +1,205 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Engelsystem\Migrations;
+
+use Engelsystem\Database\Migration\Migration;
+use Illuminate\Database\Connection;
+use Illuminate\Database\Schema\Builder as SchemaBuilder;
+
+class RefactorPermissionsAndGroups extends Migration
+{
+    protected int $developer = 90;
+    protected int $bureaucrat = 80;
+    protected int $shiCo = 60;
+    protected int $newsAdmin = 85;
+    protected int $teamCoordinator = 65;
+    protected int $angel = 20;
+
+    protected int $active;
+    protected int $driveEdit;
+    protected int $eventConfig;
+    protected int $goodieEdit;
+    protected int $ifsgEdit;
+    protected int $log;
+    protected int $news;
+    protected int $register;
+    protected int $scheduleImport;
+    protected int $shifts;
+    protected int $user;
+    protected int $userAngeltypes;
+    protected int $userShifts;
+
+    protected string $shiftentry = 'shiftentry_edit_angeltype_supporter';
+    protected string $language = 'admin_language';
+    protected string $userEdit = 'user.edit';
+    protected string $userNickEdit = 'user.nick.edit';
+    protected string $shifttypes = 'shifttypes';
+    protected string $shifttypesView = 'shifttypes.view';
+
+    protected Connection $db;
+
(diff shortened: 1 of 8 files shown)


includes/pages/admin_arrive.php
```diff
@@ -206,15 +209,17 @@ function admin_arrive()
...
+            ['name' => __('general.name'),],
+            ($admin_arrive ? ['rendered_planned_arrival_date' => __('Planned arrival')] : []),
+            ['arrived' => __('Arrived?')],
+            ($admin_arrive ? [
+                'rendered_arrival_date' => __('Arrival date'),
+                'rendered_planned_departure_date' => __('Planned departure'),
+                'actions' => __('general.actions'),
+            ] : [])
+        ), $users_matched),
+        div('row', $admin_arrive ?? [
```
MyIgel, line 222: That prints a 1 at the end instead of the statistics

includes/pages/admin_arrive.php
```diff
@@ -19,48 +19,51 @@ function admin_arrive()
...
     $search = '';
     $request = request();
+    $admin_arrive = auth()->can('admin_arrive');
 
     if ($request->has('search')) {
         $search = strip_request_item('search');
         $search = trim($search);
     }
 
     $action = $request->get('action');
```
MyIgel, line 29: Can too be evaluated inside the if

src/Middleware/LegacyMiddleware.php
```diff
@@ -51,10 +51,14 @@ public function process(
             $page = str_replace('-', '_', $page);
         }
 
+        $arriveList = false;
```
MyIgel, line 54: I would name it somewhat more generic like `$pageAllowed` to allow for further permission overwrites on legacy pages

db/migrations/2024_04_20_000000_refactor_permissions_and_groups.php
```diff
@@ -0,0 +1,205 @@
...
+    protected string $shifttypesView = 'shifttypes.view';
+
+    protected Connection $db;
+
+    public function __construct(SchemaBuilder $schema)
+    {
+        parent::__construct($schema);
+        $this->db = $this->schema->getConnection();
+
+        $this->active = $this->getPrivilegeId('admin_active');
```
MyIgel, line 48: Having a mapping of `'permission.name' => 'id'` would make the code below imho more readable but thats an option too

MyIgel (changes requested):

MyIgel:
Other todos from above list:
* Bureaucrat & ShiCo admin_log
* Bureaucrat are missing the schedule.import but Devs have it now?

includes/pages/admin_arrive.php
xuwhite: -v

xuwhite: yes I could add it to both if and else if

xuwhite: move down would've helped me more

src/Middleware/LegacyMiddleware.php
```diff
@@ -51,10 +51,14 @@ public function process(
...
+        if ($page === 'admin_arrive') {
+            $arriveList = $this->auth->can('users.arrive.list');
```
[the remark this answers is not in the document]
xuwhite: please implement it

includes/pages/admin_arrive.php
MyIgel: I meant the `action` part but yeah

PR #1390, opened 2024-04-22
supporters can view users_myshifts

xuwhite:
angeltype supporters can freeload users
now they can view user myshifts from angels in their angeltype and can edit the shifts from that view
they can see freeloaded comments within their angeltype


MyIgel:
It would be great to then only show those shifts where the user is from the supported angeltype ;)

PR #1396, opened 2024-05-20
add requires info to angeltype view for not members

xuwhite:
Screenshot 2024-05-20 at 20 43 27

### DIFF
--- a/includes/view/AngelTypes_view.php
+++ b/includes/view/AngelTypes_view.php
@@ -569,6 +569,13 @@ function AngelType_view_info(
     $admin_angeltypes,
     $supporter
 ) {
+    $required_info_show = !auth()->user()
+            ->userAngelTypes()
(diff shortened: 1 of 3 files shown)


includes/view/AngelTypes_view.php
```diff
@@ -569,6 +569,10 @@ function AngelType_view_info(
     $admin_angeltypes,
     $supporter
 ) {
+    $required_info_show = !auth()->user()->userAngelTypes()->pluck('angel_types.id')->contains($angeltype->id)
```
MyIgel, line 572:
```suggestion
    $required_info_show = !auth()->user()
        ->userAngelTypes()
        ->where('angel_types.id', $angeltype->id)
        ->count()
```
would fit a bit better but its fine as it is

MyIgel (changes requested):

PR #1411, opened 2024-08-03
Add nix flake

6543:


xuwhite:
If you still work on the changes please mark the PR as draft until it is ready to be reviewed

6543:
the flake works for me to get it working ... so I guess it's ready

MyIgel:
Imho configuration for packaging should be provided in a separate repository and maintained by someone that actually uses it (which i don't), do you want to commt to that maintenance? If yes i can create a repository for the nix config

6543:
well this is not about packaging it's about providing an dev env via nix

MyIgel:
Fair point, its still about configuring some special software / env which would clutter the project root even more.
So for the same reason i don't want to add any  `.vscode` or `.idea directories even if it might have some slight benefits i also don't want to add flake configs, especially as the lock file needs additional maintenance and i can definitely use that time on other PRs more effectively.

But to make configuring your dev env setup easier we can definitely add `flake.*` to the `.gitignore` so that it can be used independently :)
If you want you can update your PR accordingly.

6543:
well I think this pull is documented enough ... as you don't want it in, I just close it... interested people can pick it up if wanted :)

I personally get why you want to have a rather clean root but dont get why that's more important than having a good onboarding for new devs ... but It's not my project so i wont argue :)

PR #1413, opened 2024-08-07
rename users' Shifts to My Shifts

Lotterleben:
As a user, I've found it super confusing that "Shifts" does not show all available shifts, but rather the shifts I signed up for – which were none. This empty view led me to assume that the shift plan just wasn't available yet, so I didn't sign up for any shifts. I have a hunch that some of our users* had the same misconception as well, leading us to miss out on helping hands. This PR aims to resolve this by making it clear that this overview shows *my* shifts, not all shifts.

\* at https://engel.wutzrock.de/

### DIFF
--- a/includes/sys_menu.php
+++ b/includes/sys_menu.php
@@ -58,7 +58,7 @@ function make_navigation()
     $pages = [
         'news'           => __('news.title'),
         'meetings'       => [__('news.title.meetings'), 'user_meetings'],
-        'user_shifts'    => __('general.shifts'),
+        'user_shifts'    => __('general.my_shifts'),
         'angeltypes'     => __('angeltypes.angeltypes'),
         'questions'      => [__('Ask the Heaven'), 'question.add'],
     ];
--- a/includes/sys_menu.php
+++ b/includes/sys_menu.php
@@ -55,18 +55,10 @@ function make_navigation()
 {
     $page = current_page();
     $menu = [];
-
-    # figure out if the menu is displayed to an admin account or a regular user
-    # and pick the correct name for the shift overview link
-    $shift_menu_name = 'general.my_shifts';
-    if (auth()->can('admin_shifts')) {
-        $shift_menu_name = 'general.shifts';
-    }
-
     $pages = [
         'news'           => __('news.title'),
         'meetings'       => [__('news.title.meetings'), 'user_meetings'],
-        'user_shifts'    => __($shift_menu_name),
+        'user_shifts'    => __('general.shifts'),
         'angeltypes'     => __('angeltypes.angeltypes'),
         'questions'      => [__('Ask the Heaven'), 'question.add'],
     ];
--- a/includes/sys_menu.php
+++ b/includes/sys_menu.php
@@ -55,10 +55,18 @@ function make_navigation()
 {
     $page = current_page();
     $menu = [];
+
+    # figure out if the menu is displayed to an admin account or a regular user
+    # and pick the correct name for the shift overview link
+    $shift_menu_name = 'general.my_shifts';
+    if (auth()->can('admin_shifts')) {
+        $shift_menu_name = 'general.shifts';
+    }
+
     $pages = [
         'news'           => __('news.title'),
         'meetings'       => [__('news.title.meetings'), 'user_meetings'],
-        'user_shifts'    => __('general.my_shifts'),
+        'user_shifts'    => __($shift_menu_name),
         'angeltypes'     => __('angeltypes.angeltypes'),
         'questions'      => [__('Ask the Heaven'), 'question.add'],
     ];
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -539,9 +539,6 @@ msgstr "Nr."
 msgid "general.shifts"
 msgstr "Schichten"
 
-msgid "general.my_shifts"
-msgstr "Meine Schichten"
-
 msgid "Length"
 msgstr "Länge"
 
@@ -1647,6 +1644,12 @@ msgstr "Mehr Informationen"
 
 msgid "settings.profile.angeltypes.info"
 msgstr "Du kannst deine Engeltypen <a href=\"%s\">auf der Engeltypen-Seite</a> verwalten."
+
+msgid "profile.my_shifts"
+msgstr "Meine Schichten"
+
+msgid "settings.profile"
+msgstr "Profil"
 
 msgid "settings.password"
 msgstr "Passwort"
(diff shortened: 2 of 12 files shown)


resources/lang/de_DE/default.po
```diff
@@ -539,6 +539,9 @@ msgstr "Nr."
 msgid "general.shifts"
 msgstr "Schichten"
 
+msgid "general.my_shifts"
+msgstr "Meine Schichten"
+
```
xuwhite, line 544:
```suggestion
```
try moving the existing translation from `additional.po` here

resources/lang/en_US/default.po
```diff
@@ -37,6 +37,9 @@ msgstr "Name"
 msgid "general.shifts"
 msgstr "Shifts"
 
+msgid "general.my_shifts"
+msgstr "My Shifts"
+
```
xuwhite, line 42:
```suggestion
```
same here move the existing translation

xuwhite (changes requested):
Thanks for the PR. The idea is very good, unfortunately your changes are on the shifts overview and the admin-active page. if you wanted to change the heading on the user page, you can find it at `includes/view/User_view.php:740`.
Please note that this page can be viewed by 2 groups, firstly admins for whom the heading "Shifts" is correct and secondly the user himself, where the heading "My Shifts" makes sense.
Additionally there already exists a translation for "My shifts" at `resources/lang/de_DE/additional.po:172` and `resources/lang/en_US/additional.po:171` its sadly is in the wrong files and should be in `resources/lang/de_DE/default.po` and `resources/lang/en_US/default.po`.
If you have any questions solving your issue feel free to ask and I will be more than happy to help you.

Lotterleben:
Got it! Thank you for the super helpful review & tips, I'll get started on this next week. :)

includes/sys_menu.php
```diff
@@ -58,7 +58,7 @@ function make_navigation()
     $pages = [
         'news'           => __('news.title'),
         'meetings'       => [__('news.title.meetings'), 'user_meetings'],
-        'user_shifts'    => __('general.shifts'),
+        'user_shifts'    => __('general.my_shifts'),
```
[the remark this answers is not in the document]
Lotterleben: wait – but changing the menu was precisely my intention – at least for users (I see how the wording doesn't apply to admins 😅). Imo, this improves the discoverability of the menu.
keeping the menu item as "shifts", but then having the shift page heading be "my shifts" would imo just be an additional source of confusion because it's inconsistent.

I've implemented a suggestion on how to handle this in https://github.com/engelsystem/engelsystem/pull/1413/commits/bf0f3f374ec8fcf37d8c8544b931ca5129354202 , do you think this approach would be feasible? (I have 0 php knowledge, will be happy to research more elegant ways to implement this if you're happy with the general direction)

resources/lang/de_DE/default.po
Lotterleben: would you keep the `profile.` prefix from `additional.po` (as far as I can see there's no other item with this exact prefix) or would you convert current uses of `profile.my_shifts`to `general.my_shifts` – or pick a different prefix entirely?

includes/sys_menu.php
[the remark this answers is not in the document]
xuwhite: Why do you want to change this page's title to "My shifts"? 

This page shows all (available) shifts. That's what I meant by "shifts overview" page. `/user-shifts`

Screenshot 2024-08-17 at 15 05 48


The page that shows all shifts you signed up for is your user profile `/users?action=view`

Screenshot 2024-08-17 at 15 05 08

I interpreted your comment "I've found it super confusing that "Shifts" does not show all available shifts, but rather the shifts I signed up for " to mean that you want to change the heading "Shifts" of the user page `/users?action=view` to "my shifts" for not admins, which is a good idea.

However changing the shifts overview page title to "My shifts" makes no sense, since this is an overview of all shifts (and avialability state) for all users.

includes/sys_menu.php
```diff
@@ -55,10 +55,18 @@ function make_navigation()
 {
     $page = current_page();
     $menu = [];
+
+    # figure out if the menu is displayed to an admin account or a regular user
+    # and pick the correct name for the shift overview link
+    $shift_menu_name = 'general.my_shifts';
```
xuwhite, line 61: the implementation is fine, sadly it makes no sense for this page (see other review comments)

resources/lang/de_DE/default.po
xuwhite: just move them from `additional.po` to `default.po`

includes/sys_menu.php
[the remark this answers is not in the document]
Lotterleben: 🤯 Oohh I think there's been both a miscommunication between us and a misunderstanding on our (organizations') part on how to see & sign up for shifts.
I've been solving the wrong UX bug all along! :D

So, from my current understanding, what's been happening is the following:
In our shift system, there are so many rooms that the list of them pushes the shift table off the screen. Additionally, Changing a date in the drop down menus does not change anything on the page (I just noticed now that you's need to click the "filter" button, which is different to the behaviour of the room page date drop-down menus). So, this overview looked empty to us:

Screenshot 2024-08-26 at 14 33 06


Which is why we instructed all of our helpers to use the "Rooms" drop down, pick an area of the festival they wanted to help with, and then sign up for a shift from there. This is not a path that's easy to discover on your own, so with this PR I wanted to guide users away from the path I *thought* was wrong – turns out it *was* right, the relevant information was just hiding off the screen 😅.

And this wasn't just me, it was *all* of our organization – I've been taught to go via the room view by peers, if I remember it correctly. (If you'd like to see for yourself, the shifts for august 9th-11th are still online at https://engel.wutzrock.de/)

my suggestion for next steps would be:
1. I'll re-work this PR to do what you interpreted it to be intended for ( change the heading "Shifts" of the user page /users?action=view to "my shifts" for not admins)
2. I'll open a separate PR implementing an immediate page update after the user changes something in the drop-down menu or clicks one of the "gestern","heute",.... buttons (just like the drop down menu in the room menu behaves), obsoleting the "Filter" button (at least for the time overview – if it's too slow to update on every room/angel type checkbox change, I'd visually & functionally move the Filter button there).
3. I start working on a proposal to re-arrange the Shifts page that the shifts aren't pushed so far down that you can't see them 

What do you think?

xuwhite:
since there was no update for over a month I have pushed the changes I suggested

includes/sys_menu.php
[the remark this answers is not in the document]
MyIgel: The suggestions sound good, but we have to make sure that its not making the problem "worse" (for example side effects when users add themselves to an angel type and don't see the filter to add them there too)

Do you want to create a new issue for this to discuss a possible solution?

PR #1416, opened 2024-08-09
Fix user-shifts bug

xuwhite:
add angel types from shift entries, which have no reference from needed angel types, to user-shifts

### DIFF
--- a/includes/view/ShiftCalendarRenderer.php
+++ b/includes/view/ShiftCalendarRenderer.php
@@ -3,6 +3,7 @@
 namespace Engelsystem;
 
 use Engelsystem\Helpers\Carbon;
+use Engelsystem\Models\AngelType;
 use Engelsystem\Models\Shifts\Shift;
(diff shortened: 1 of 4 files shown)


includes/view/ShiftCalendarRenderer.php
```diff
@@ -181,9 +182,39 @@ private function renderLane(ShiftCalendarLane $lane)
                 $rendered_until += ShiftCalendarRenderer::SECONDS_PER_ROW;
             }
 
+            $needed_angeltypes = $this->needed_angeltypes[$shift->id];
+            // add angel types from shift entries without reference from needed angel types
+            $needed_angels = new Collection($shift->neededAngels);
+//            if($shift->id == 732){
+//                dd($needed_angeltypes);
```
MyIgel, line 189: ???

MyIgel (changes requested):

PR #1418, opened 2024-08-11
Added dynamic configuration page, refactored form validation

MyIgel:
* Removed unused not operator from validation and simplified flow
* Allow for nested rules in validation by using arrays
* Added  validations:
  * `date_time`
  * `before` datetime
  * `after` datetime
  * `truthy` helper
* Removed ValidationResult from angeltypes page
* Added new Application Configuration page to replace old event config


msquare:
Filename `setings.twig` is missing a `t` ;-)

MyIgel:
@msquare fixed :D

PR #1421, opened 2024-08-16
DEVELOPMENT.md: add hints on why your docker (watch) setup may not be working

Lotterleben:
I've stubbed my toes on the way `:watch` was(n't) working for me for the third time now, so this PR adds some hints that I think might be helpful to other new contributors (and future me)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -81,10 +81,10 @@ docker compose exec es_workspace bin/migrate
 While developing you may use the watch mode to rebuild the system on changes
 
 ```bash
-# Run a front-end build and update on change. Note that this doesn't apply to changed translation files – see above on how to update them manually
+# Run a front-end build and update every time a JS or CSS file is changed (not translation files!)
 docker compose exec es_workspace yarn build:watch
 
-# Or run a front-end build and update on change for specific themes only, e.g.
+# Or build and update on change for specific themes only to save build time, e.g.
 docker compose exec -e THEMES=0,1 es_workspace yarn build:watch
 ```
 
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -89,7 +89,7 @@ docker compose exec -e THEMES=0,1 es_workspace yarn build:watch
 ```
 
 > [!NOTE]
-> Wait for a few minutes after running `docker compose exec` – it may look like it's stalling, but it's not.
+> Wait some time (up to a few minutes) after running `yarn build:watch` – it may look like it's stalling, but it's not.
 
 
 It might also be useful to have an interactive database interface for which a phpMyAdmin instance can be startet at [http://localhost:8888](http://localhost:8888).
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -81,13 +81,17 @@ docker compose exec es_workspace bin/migrate
...
-# Run a front-end build and update on change
+# Run a front-end build and update on change. Note that this doesn't apply to changed translation files – see above on how to update them manually
 docker compose exec es_workspace yarn build:watch
 
 # Or run a front-end build and update on change for specific themes only, e.g.
 docker compose exec -e THEMES=0,1 es_workspace yarn build:watch
 ```
 
+> [!NOTE]
+> Wait for a few minutes after running `docker compose exec` – it may look like it's stalling, but it's not.
```
MyIgel, line 92: As the relevant command here is `yarn build`, i would reference that:
```suggestion
> Wait some time (up to a few minutes) after running `yarn build:watch` – it may look like it's stalling, but it's not.
```

DEVELOPMENT.md
```diff
@@ -81,13 +81,17 @@ docker compose exec es_workspace bin/migrate
 While developing you may use the watch mode to rebuild the system on changes
 
 ```bash
-# Run a front-end build and update on change
+# Run a front-end build and update on change. Note that this doesn't apply to changed translation files – see above on how to update them manually
```
MyIgel, line 84: "Frontend build" should have no connotation of "doing translations" (as the don't have to be build; using `.mo` files is only recommended for prod; dev works without problems when directly parsing the `.po`.

So something like
```suggestion
# Run a front-end build and update on change (JS & CSS, no translations)
```

MyIgel (changes requested):

MyIgel:
It pretty much depends on the environment how fast it is (thus the option to filter at least the theme you are working with to reduce the runtime) :D

DEVELOPMENT.md
Lotterleben: As someone who is not very familiar with frontend development and just wanted to see if I could tweak some things we were struggling with, all of these distinctions were not obvious to me (I changed some wording in a .po file to see if my setup was working, nothing changed -> I looked for the error in my setup). The wording I added here was an attempt to help other newbies like me – I'm very open to finding shorter ways to express this, but `(JS & CSS, no translations)` would not have helped me made the connection. 

How about something like...

```suggestion
# Run a front-end build and update every tome a JS or CSS file is changed (not translation files!)
```
?

Additionally, your comment makes me question whether I messed up my setup somehow because I still have to do the .po->.mo updates manually in order for changes in .po files to show up in my dev instance. I'll investigate and might come back with an additional PR :D

MyIgel: The `.mo` files are binary files that are easier to parse and thuis give a small performance boost on prod systems but the angelsytem is totally useable when they are not present, it uses the text `.po` files automatically.

So yees, you should delete your `.mo` files for the dev env ;)

MyIgel:
Sure why not, we can mention it more explicit :D

MyIgel:
Thanks for the clarifications! <3

PR #1422, opened 2024-08-17
fix wording and translations

xuwhite:
"angeltype" => "angel type"
"shifttype" => "shift type"

# Conflicts:
#	includes/controller/user_angeltypes_controller.php #	includes/pages/admin_shifts.php


includes/controller/user_angeltypes_controller.php
```diff
@@ -38,19 +38,18 @@ function user_angeltypes_unconfirmed_hint()
...
             . '</a>';
     }
 
     $count = $unconfirmed_user_angeltypes->count();
     return
         _e(
-            'There is %d unconfirmed angeltype.',
-            'There are %d unconfirmed angeltypes.',
+            'There is %d unconfirmed angel type. Angel type that needs approval:',
+            'There are %d unconfirmed angel types. Angel types that need approvals:',
```
MyIgel, line 51: Technically, users are confirmed, not the angel type

MyIgel (changes requested):
Only one small thing (and i would not word the error commit that way), other than that its ok and can be merged

includes/controller/user_angeltypes_controller.php
xuwhite: true but this is the original code. I'll fix it anyway.

PR #1427, opened 2024-09-07
Bugfix/40 missing quote in german translation file

kshade:
https://github.com/eurofurence/crittersystem/issues/40


MyIgel:
Hi @kshade, i don't think that this PR is supposed to go to this repo? :D

& as I'm looking through your changes: You should have a look at https://engelsystem.de/doc/admin/translation/#overwriting-translations to dynamically replace the translations instead of editing the code, that helps a lot with maintainability ;)

PR #1434, opened 2024-10-07
add angel types metrics with membership state

### LINKED ISSUE #1432: Some information in /metrics missing
Hey,

I just looked at the /metrics endpoint and there are quite a few metrics which are questionable to me (how many user use which skin? :D), but while building some visualizations in Grafana, I quickly found two rather important metrics that are missing:

- The **total** amount of work hours in the event. When looking at the dashboard I see "Hours to be worked" but this is number is not part of the /metrics endpoint. I get how many hours are done and planned by angels, but without knowing the total, that's not really useful.

- For "alerting" purposes of the admins it's pretty cool that I get the number of pending questions as a metric. Similar to that, getting the number of of pending angels for an angeltype that requires introduction would also be cool. E.g. with the angeltype as a label so the teams could filter out only what they are interested in.

I had some more ideas for missing metrics the other days, so maybe this list will grow a bit over the next few days :D

Best,
fleaz

xuwhite:
see #1432 

Screenshot 2024-10-07 at 22 54 51

### DIFF
--- a/src/Controllers/Metrics/Controller.php
+++ b/src/Controllers/Metrics/Controller.php
@@ -52,6 +52,31 @@ class Controller extends BaseController
             $userOauth[$key]['labels']['name'] = $name;
         }
 
+        $angelTypes = [];
+        foreach ($this->stats->angelTypes() as $angelType) {
+            $angelTypes[] = [
+                'labels' => [
+                    'name' => $angelType['name'],
+                    'restricted' => $angelType['restricted'],
+                    'members' => 'unconfirmed',
+                ], 'value' => $angelType['unconfirmed'],
+            ];
+            $angelTypes[] = [
+                'labels' => [
+                    'name' => $angelType['name'],
+                    'restricted' => $angelType['restricted'],
+                    'members' => 'confirmed',
+                ], 'value' => $angelType['confirmed'],
+            ];
+            $angelTypes[] = [
+                'labels' => [
+                    'name' => $angelType['name'],
+                    'restricted' => $angelType['restricted'],
+                    'members' => 'supporters',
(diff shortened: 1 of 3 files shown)


src/Controllers/Metrics/Stats.php
```diff
@@ -302,21 +303,47 @@ public function locations(): int
...
     {
-        return AngelType::query()
-            ->count();
+        return AngelType::query()->count();
+    }
+
+    public function angelTypes(): array
+    {
+        $angelTypes = [];
+        $rawAngelTypes = AngelType::query()->select(['id', 'name', 'restricted'])->orderBy('id')->get();
```
MyIgel, line 320:
```suggestion
        $rawAngelTypes = AngelType::query()->select(['id', 'name', 'restricted'])->orderBy('name')->get();
```

src/Controllers/Metrics/Controller.php
```diff
@@ -52,6 +52,33 @@ public function metrics(): Response
...
         }
 
+        $angelTypes = [];
+        foreach ($this->stats->angelTypes() as $angelType) {
+            $angelTypes[] = [
+                'labels' => [
+                    'name' => $angelType['name'],
+                    'restricted' => $angelType['restricted'],
+                    'members' => 'unconfirmed',
+                ], 'value' => $angelType['restricted'] ? ($angelType['members'] - $angelType['confirmed']) : 0,
```
MyIgel, line 62: this should eventually be calculated in the Stats class

MyIgel (changes requested):

xuwhite:
I'm merging for @MyIgel

PR #1442, opened 2024-11-01
simplify e-mail preferences on registration

xuwhite:
Screenshot 2024-11-01 at 15 22 55


MyIgel (changes requested):
Some smaller changes, the rest looks great!

PR #1473, opened 2024-12-15
Smaller fixes and updates

### LINKED ISSUE #1463: "What do you want to do" is not explained
How to reproduce.
* want to be an engel for the first time
* Go to https://engel.events.ccc.de/register and try to fill the form. Ensure it's in English.
* Read the question "what do you want to do", and try to understand what is the Späti Angel that is preselecetd, or what is FOC salesperson that is not preselected
* Read the angel guide https://angelguide.c3heaven.de/

Expected result:
* finding the words Späti or FOC somewhere.(This is the only explanaiton page linked here. It's linked from "T-shirt size" only)

Actual result:
* Filling with random selection and hoping it'll be easier to understand what the choices meant.

I'd suggest adding a link to the description, or adding the description to the angel guide. Or, if the angel guide already contains the description, ensure that the same vocabulary is used in the form and in the document so that details can be found

### LINKED ISSUE #1464: Add member to angeltype as supporter
Currently, it's possible to add a member to an angeltype with a checkbox (confirm user). Especially on setting up an event the most used task is to create an angeltype and add the supporters for it. This is a multistep process currently: add member and then add supporter rights. It would be neat to do this in a single step.

Idea: Instead of the current "confirm user" checkbox 3 radio boxes could be helpful: Supporter, Member, Unconfirmed. Keeping the default the same would be Member.

### LINKED ISSUE #1465: Edit Shift Needed Angeltypes list is full-width
`/user-shifts?edit_shift=` uses full width needed angeltypes while `/admin-shifts` displays multiple per row which is way easier to navigate with many angeltypes.

### LINKED ISSUE #1470: Create/Edit Angel type field order
When viewing an angel type the contact is first, then the description.

When editing it its the other way around, first description, then contact. Personally I would put the one line thingies (contact) up top and then the description there too.

### LINKED ISSUE #1471: Create Angel Type default should require introduction
Angel types should be as few as possible. If there is no introduction required, there is probably no need for a dedicated angel type. Therefore the default is to require an introduction. Its also easier to open up an existing angel type rather than closing it again.

So I propose having it enabled by default.

### LINKED ISSUE #1480: Double Signup as Supporter
A user in the supporter role can sign up the same angel to the same shift in the same role twice.
This also might happen for themself by accident.

To reproduce:
Be supporter of any angeltype. Sign up to shift. Sign up to shift again.

Expected Behaviour:
The same angel should not be added to the shift more than once.

MyIgel:
* Default angel types to restricted, fixes  #1471
* Move checkboxes on edit angel type page to the right, fixes #1470
* Update shift edit to match creation, fixes #1465
* Show link to team/angel type description on registration, fixes #1463
* Allow adding angels as supporter for angel type when adding, fixes #1464
* Ported comment fixes made by @chipuni (happy to have more input if you need some other stuff that can be added here)
* Updated translation of message that angel user needs to arrive
* Fixed freeload metrics & flaky tests
* Show SSO menu and entries in profile settings when entered, even if they are normally hidden
* Schedule creation: Show form content on validation errors
* Show total configured t-shirts/goodies on active page
* Show info that user is already in a shift when adding, fixes #1480
* Link FAQ on angeltypes about page

### DIFF
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -1914,7 +1914,8 @@ msgid "angeltypes.about"
 msgstr "Teams-/Aufgabenbeschreibung"
 
 msgid "angeltypes.about.text"
-msgstr "Hier findest Du die Liste der Teams und ihrer Aufgaben. Wenn Du weitere Fragen hast, schaue in den FAQ nach."
+msgstr "Hier findest Du die Liste der Teams und ihrer Aufgaben. Wenn Du weitere Fragen hast, "
+"schaue in den <a href=\"%s\">FAQ</a> nach."
 
 msgid "angeltypes.restricted.hint"
 msgstr "Dieser Engeltyp benötigt eine Einweisung bei einem Einführungstreffen. "
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -282,12 +282,6 @@ msgstr "Titel"
 msgid "Location:"
 msgstr "Ort:"
 
-msgid "Start:"
-msgstr "Start:"
-
-msgid "End:"
-msgstr "Ende:"
-
 msgid "Needed angels"
 msgstr "Benötigte Engel"
 
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -974,7 +974,7 @@ msgid "ended"
 msgstr "vorbei"
 
 msgid "please arrive for signup"
-msgstr "Ankommen zum Eintragen"
+msgstr "Komme an um dich einzutragen"
 
 msgid "not yet possible"
 msgstr "noch nicht möglich"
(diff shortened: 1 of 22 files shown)


resources/lang/en_US/default.po
```diff
@@ -73,7 +73,7 @@ msgid "page.419.title"
 msgstr "Authentication expired"
 
 msgid "page.419.text"
```
xuwhite, line 75: please also  add this to the German translation

resources/lang/de_DE/default.po
```diff
@@ -974,7 +974,7 @@ msgid "ended"
 msgstr "vorbei"
 
 msgid "please arrive for signup"
-msgstr "Ankommen zum Eintragen"
+msgstr "Du bist noch nicht angekommen"
```
xuwhite, line 983: "bitte komme an um dich einzutragen"

resources/lang/de_DE/default.po
```diff
@@ -527,6 +527,9 @@ msgstr "T-Shirt Statistik"
 msgid "Given T-shirts"
 msgstr "Ausgegebene T-Shirts"
 
+msgid "Configured T-Shirts"
```
xuwhite, line 533: "Configured T-shirts"

Issue #565, opened 2019-01-01
Maximum count of consecutive shifts (configurable by shift type)


der-michik:
A lot of tasks in shifts done at congress are physically and/or mentally challenging and currently, there is no safeguard against angels trying to sign up e.g. for 12 hours of bar runner shifts in a row (completely fictional example). Some angels don't know their own limits and become overworked easily. With the congress growing, neither Heaven nor most other teams have a chance to oversee this and react proactively.

It would be great to have a setting, global and/or per shift type, that prohibits angels from signing up for a too large number of demanding shifts (either as number of shifts or better and more flexible in units of time) in a row and forces a mandatory break of a configurable duration after that amount. This would not only protect these angels from themselves but also ensure that shifts are distributed more fairly.

olebittner:
I am interested in working on this issue. Therefore I propose a possible implementation below. I've tried to cover @MichiK's ideas from above as well as from [#746](https://github.com/engelsystem/engelsystem/issues/746#issuecomment-672302642). I'm happy for any feedback or discussion, especially for the open questions regarding edge cases and communication. I've split the issue into *basic* and *advanced* implementations as I think the topic is quite large and can be implemented well in multiple steps.

## Basic Implementation
An angel may work a maximum of X hours. After that, a break of Y hours must be taken before the next shift can begin.

This explicitly does not prevent angels from signing up for a lot of shifts in advance, as long as they leave the required breaks in between. As discussed in #746 this safeguard and fair distribution of shifts should not be handled by the same system.

### Global Config
- toggle to enable/disable this system
- maximum working hours (X)
- rest period hours (Y)

## Advanced Implementation (demanding work)
Shift types can be marked as ‘demanding’ work. Of the X hours of work mentioned above, a maximum of Z hours may be performed in ‘demanding’ shifts.

In my opinion, this approach is a good compromise between simplicity and the fact that different tasks are demanding in different ways. The threshold value (Z) needs to be well chosen here so as not to discourage teams from marking their shifts as ‘demanding’ by setting it too low. Another approach could be an individual hour limit per shift type, but I think that would be too intransparent and raise questions like 'Why can I sign-up for this shift but not that one'. It also makes the answer to the simple question 'How much can I work today?' really complicated i.e. '8h, but only 5h of A or 2h of B...'

### Global Config
- maximum 'demanding' working hours (Z)

### Config per shift typ
- is this shift type 'demanding'? True/False

## Edge Cases
[...]

PR #1504, opened 2025-01-27
WIP: limit consecutive working time

olebittner:
This PR implements #565 and has the goal to prevent angels from overworking themselfs by limiting consecutive work to configured hours before requiring a rest.

- [ ] Basic Implementation
    - [x] Global Config: feature enabled, max working time, rest period, toggle for daily rest
    - [x] Limit consecutive working time
    - [x] Enforce daily rest period
    - [ ] Finalize user-facing messages and visualization (shift calendar)
- [ ] Demanding Work implementation
    - [ ] Global Config: max demanding working time
    - [ ] Shift Type Config: is shift type demanding
    - [ ] Limit consecutive demanding work
    - [ ] Finalize user-facing messages and visualization (shift calendar)
- [ ] Unit Tests

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -388,6 +388,18 @@ return [
     # Instruction only onsite in accordance with § 43 Para. 1 of the German Infection Protection Act (IfSG)
     'ifsg_light_enabled'           => env('IFSG_LIGHT_ENABLED', false) && env('IFSG_ENABLED', false),
 
+    // Prevent Angles from overworking themselves by working too many consecutive shifts
+    'maximum_working_time' => [
+        // enable to enforce work time limit / disable to allow unlimited consecutive work
+        'enabled' => (bool) env('MAX_WORKING_TIME_ENABLED', false),
+        // maximum hours of work before a rest period is required
+        'max_working_hours' => env('MAX_WORKING_TIME_WORK_HOURS', 10),
+        // length of required rest period
+        'min_rest_hours' => env('MAX_WORKING_TIME_REST_HOURS', 11),
+        // enforce one rest period every 24h, even if max_working_hours limit is not reached
+        'enforce_daily_rest' => env('MAX_WORKING_TIME_ENFORCE_DAILY_REST', true),
+    ],
+
     // Available locales in /resources/lang/
     // To disable a locale in config.php, you can set its value to null
     'locales'                 => [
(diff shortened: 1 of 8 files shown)


olebittner:
Hi @MyIgel & @xuwhite,

this is not ready for review yet, but could one of you give me a quick hint as to whether this is going in the right direction and is in the spirit of the engelsystem before I spend more time on it?

MyIgel:
Some thoughts about those changes:

Imho we should split this feature in a MVP and parts of a "full blown work time enforcement and tracking system".

The MVP should contain the grouping of shifts in "time blocks" to make further comparison easier:
* This must be done only once per request and not more (takes time for the same calculations, no need to repeat it)
* Definition of a block:
  * One or more shifts that have no (0s) pause between them
  * Parallel or overlapping shifts are merged into a block
  * Length equals start of the first shift till the end of the last shift
* Supporters / shift coordinators / bureaucrats must be able to sign in users even if it collides with above rules
  * Should get a warning that this is the case
      * which could be implemented with a popup that requires confirmation (not feasible at the moment -> performance)
      * as a notification after signing up the user
      * as an additionally confirmation (javascript?) pop up or page

Then it should check that a given shift where a user could sign up for does not extend / merge to a given slot so that it results in a combined work time that's bigger than the configured time. It should still be possible to sign up for shifts that are longer than the configured max. hours of consecutive work of a time block.

Later extensions might contain:
* Min. pause times, after a certain amount of hours in a time block (for later, as it makes the MVP more complicated)
* Min "night" time / time between time blocks (aka sleep) / Minimum time between "work blocks" (group of time blocks with pause times between them)
* Max time of a work block
* Eventually some rating/weighting of shiffts ("pause", "normal", "demanding") with special handling (phone support could for example count as a pause even if its a shift)


To the code itself:
I would appreciate if the config format / comments and translation texts could be more precise regarding above requirements / steps.
Generally calculation with seconds is possible (its old code anyways) but directly working with the `DateTime` objects and thus `DateInterval`s would make it easier to be moved to `src/` after refactoring.

xuwhite (changes requested):
Thanks for the PR and sorry for the late answer. @MyIgel and I discussed this quite a lot. At the end we both agreed on the text above. I just have a minor change request to add ;D

olebittner:
As discussed elsewhere, this feature is not as relevant at the moment because very few angles actually would be affected by this. If this becomes relevant later, I'm happy to help with the implementation.

PR #1508, opened 2025-02-02
Add schedule event guid to shift in api and origin_url to schedule.xml parser

MyIgel:
* Added schedule event guid to api to make translation mapping easier
  Screenshot_20250202_182532
* Schedule: Add new `origin_url` to schedule.xml parser
* API: Add /users endpoint


msquare (approved):
Approved! 👌

PR #1509, opened 2025-02-02
Add pagination to "All Users" view

MyIgel:


config/config.default.php
```diff
@@ -257,9 +257,12 @@
     // Must be one of news, meetings, user_shifts, angeltypes, questions
     'home_site'               => env('HOME_SITE', 'news'),
 
-    // Number of News shown on one site and for feed readers (minimum 1)
+    // Number of news shown on one site and for feed readers (minimum 1)
     'display_news'            => env('DISPLAY_NEWS', 10),
 
+    // Number of users shown on one page in the backend
+    'display_users'           => env('DISPLAY_USERS', 100),
```
xuwhite, line 285: I would love if this would be changeable on the all users page itself with some pre defined numbers, so that every user could decide how many works for them
I like this as a default though

includes/view/User_view.php
```diff
@@ -201,6 +202,7 @@ class="btn btn-sm btn-secondary ms-2 js-only"
     return page_with_title(__('All users') . ' ' . $link, [
         msg(),
         table($user_table_headers, $usersList),
+        pagination($users, true),
```
xuwhite: I would love having this on the top of the table too

xuwhite: and if there is only one page please don't show this at all

includes/view/User_view.php
```diff
@@ -201,6 +202,7 @@ class="btn btn-sm btn-secondary ms-2 js-only"
     return page_with_title(__('All users') . ' ' . $link, [
         msg(),
         table($user_table_headers, $usersList),
```
xuwhite, line 220: I found a bug:
I clicked the "all" button, and got the whole page but the pagination at the end of page changed to only one page 

in pictures: 
before clicking "all"
Screenshot 2025-06-01 at 15 51 01

after clicking "all":
Screenshot 2025-06-01 at 15 51 15

config/config.default.php
MyIgel: That would be a great idea for an additional MR but is here out of scope

includes/view/User_view.php
MyIgel: It is technically one page then?

Will be fixed by simply showing no buttons as requested above.

xuwhite: add a dropdown on how many entries should be shown on one page. This dropdown should include all as an option

MyIgel: that makes the code much more ugly but sure...

config/config.default.php
MyIgel: implemented -> its ugly now..

PR #1512, opened 2025-02-12
small improvements

### LINKED ISSUE #563: Show Next/Previous Shift in Shift View
When viewing a certain shift a dropdown or button to immediately jump to the next or previous shift would be cool. The heaven often has to do work like "Can you put me into the next shift as well?", "There is no one at this door. Who had the previous shift?", etc.

For example when viewing the Access Control Main Entrance 7:00 - 9:00 Shift it should be possible to immediately jump to the 9:00 - 11:00 and the 5:00 - 7:00 shift at the main entrance.

We need to think about the behaviour of overlapping shifts in this as well. So I'd like to propose two drop downs showing the 10 next and 10 previous shifts respectively in the same location sorted by time.

xuwhite:
* new permission for view user info hint
* memory limit in pre-commit script
* next shift and previous shift buttons on shift view closes #563 
* update and translate permission descriptions
* edit user delete log message

### DIFF
--- a/bin/pre-commit
+++ b/bin/pre-commit
@@ -16,7 +16,7 @@ testing 'PHP ⚙️'
 composer validate
 composer phpcs
 composer phpstan
-./vendor/bin/phpunit
+php -d memory_limit=-1 ./vendor/bin/phpunit
 
 testing 'translations 🗺️'
 find resources/lang -type f -name '*.po' -exec sh -c 'msgfmt "${1%.*}.po" -o"${1%.*}.mo"' shell {} \;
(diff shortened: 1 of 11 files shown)


xuwhite:
Copy and paste bug, should be fixed now.

> I have a failing test? Can you reproduce it? There was 1 error:
> 
> 1. Engelsystem\Test\Unit\Models\Shifts\ShiftTest::testPreviousShift
>    Attempt to read property "id" on null
> 
> /var/www/tests/Unit/Models/Shifts/ShiftTest.php:194

PR #1518, opened 2025-02-26
Make alert in shifts view harder to miss

Lotterleben:
This PR depends on https://github.com/engelsystem/engelsystem/pull/1517
It intends to make the alert that is shown if you can't sign up for a shift even harder to miss by embedding it between the shift view header and description. My hope is that this will decrease angel churn as well support requests by angels.

I've also taken the liberty to refactor the alert rendering into its own method for readability.

this is what the before/after looks like (the change in color comes from #1517)

### before
Screenshot 2025-02-26 at 14 22 45

### after
Screenshot 2025-02-26 at 14 21 16

### DIFF
--- a/includes/view/Shifts_view.php
+++ b/includes/view/Shifts_view.php
@@ -200,25 +200,6 @@ function Shift_view(
 
     $content = [msg()];
 
-    if ($shift_signup_state->getState() === ShiftSignupStatus::COLLIDES) {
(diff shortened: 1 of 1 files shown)


xuwhite:
Thank you very much for this PR.
I looked at it and like the placement, but the message block is quite long, with not much text in it, so I changed it a bit to make the message only appear on the left half of the screen above the needed angels heading 
Screenshot 2025-04-09 at 16 46 55
I hope you like it.

Issue #1482, opened 2024-12-25
add "own" option to shift occupancy filter


Kruemmelspalter:
currently, the only two options in the occupancy filter are "free" and "occupied"
![image](https://github.com/user-attachments/assets/542653ee-c70d-477c-b715-e77c51f5d0d2)

I don't know if this is a sensible or widely needed use case, but I would really like to see all free shifts and all shifts that I signed up for, even those that are already fully occupied.

My suggestion is to add an additional option to the occupancy filter that says "own" which adds all shifts one signed up for.

PR #1522, opened 2025-03-18
Add own shifts filter to shift calendar view

Enterprize1:
This adds the feature requested in #1482 by adding a new occupancy filter to the calendar view called "own" like this:

image

### DIFF
--- a/includes/controller/users_controller.php
+++ b/includes/controller/users_controller.php
@@ -433,6 +433,11 @@ function shiftCalendarRendererByShiftFilter(ShiftsFilter $shiftsFilter)
         return new ShiftCalendarRenderer($shifts, $needed_angeltypes, $shift_entries, $shiftsFilter);
     }
 
+    $own_shift_ids = [];
+    if (in_array(ShiftsFilter::FILLED_OWN, $shiftsFilter->getFilled())) {
+        $own_shift_ids = Shifts_by_user(auth()->user()->id, false)->pluck('id')->toArray();
+    }
+
     $filtered_shifts = [];
     foreach ($shifts as $shift) {
         $needed_angels_count = 0;
@@ -463,11 +468,20 @@ function shiftCalendarRendererByShiftFilter(ShiftsFilter $shiftsFilter)
             && $needed_angels_count > 0
         ) {
             $filtered_shifts[] = $shift;
+            continue;
         }
 
         if (
             in_array(ShiftsFilter::FILLED_FILLED, $shiftsFilter->getFilled())
             && $needed_angels_count == 0
+        ) {
+            $filtered_shifts[] = $shift;
+            continue;
+        }
+
+        if (
+            in_array(ShiftsFilter::FILLED_OWN, $shiftsFilter->getFilled())
+            && in_array($shift->id, $own_shift_ids)
         ) {
             $filtered_shifts[] = $shift;
         }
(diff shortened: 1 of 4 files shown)


includes/model/ShiftsFilter.php
```diff
@@ -21,6 +21,11 @@ class ShiftsFilter
      */
     public const FILLED_FREE = 0;
 
+    /**
```
xuwhite, line 24: while you are on it please sort them by id ;)

includes/pages/user_shifts.php
```diff
@@ -280,6 +280,10 @@ function view_user_shifts()
             'id'   => '0',
             'name' => __('free'),
         ],
+        [
+            'id'   => '2',
```
xuwhite, line 284: I have a problem with this, this option is only useful if you want to show only your own shifts without the options free or full. Since this is the case I think this would need some javascript to make the checkboxes reactive and deselect full and free when own is selected and deselect own if full or free is selected.

includes/controller/users_controller.php
```diff
@@ -433,6 +433,11 @@ function shiftCalendarRendererByShiftFilter(ShiftsFilter $shiftsFilter)
         return new ShiftCalendarRenderer($shifts, $needed_angeltypes, $shift_entries, $shiftsFilter);
     }
 
+    $own_shift_ids = [];
+    if (in_array(ShiftsFilter::FILLED_OWN, $shiftsFilter->getFilled())) {
+        $own_shift_ids = Shifts_by_user(auth()->user()->id, false)->pluck('id')->toArray();
```
xuwhite, line 438:
```suggestion
        $own_shift_ids = Shifts_by_user(auth()->user()->id)->pluck('id')->toArray();
```
false is the default so it's not needed

xuwhite (changes requested):
Thanks for the PR
I like the feature. I found some minor change requests and some javascript magic missing...

includes/pages/user_shifts.php
Enterprize1: I understood the original request from the issue to be able to view free and own shifts at the same time. The proposed javascript would prevent this.

xuwhite:
I had some fun and tried to improve the feature, what are your thoughts on this? 
https://github.com/engelsystem/engelsystem/commit/6b0bcfa669320c1d051599ec088dda2d4f62922c

Screenshot 2025-07-13 at 15 04 00

Enterprize1:
I quite like the change, that should clarify what the filters will cause and the relationship between them.

agowa:
> I quite like the change, that should clarify what the filters will cause and the relationship between them.

The initial proposal was way better. This one looks kinda like a CSS error... I also think that it is clear enough to just have the users select multiple themselves. What exactly is the issue with just going with the initial one?

Also the current state is kinda unintuitive as own shifts are currently partially shown within the dashboard. As long as there is still at least one additional helper needed. But as soon as all of the spots have been taken it just vanishes (as then it is considered an occupied one and not a "fee" one anymore.

This here is the more clean implementation I think:
image


Alternatively (or maybe even additionally) to this proposal all "own" shifts could be hidden from the "free" view. For a user that already signed up to a specific shift that won't be "up for taking" I.E. it's not "free to sign up for". So it should be filtered out.

MyIgel:
Eventually adding a `<hr>` between own and the others would be a good way here (and then implementing it as "and additionally all my own"

PR #1524, opened 2025-03-23
Add sticky headers to tables

### LINKED ISSUE #289: fixed column heads for tables
Hey,

It would be nice to have fixed column heads for table views like 'arrived angels', 'all angels' and such, since these pages tend to involve of lot of scrolling and they have a lot of columns.

Regards.

Enterprize1:
Fixes #289 

This MR introduces sticky headers to tables where possible. The sticky headers will remain visible at the top of the viewport (below the navbar) when scrolling down.

{354667E2-96D1-4506-A335-4F22FCA5E636}

### DIFF
--- a/includes/sys_template.php
+++ b/includes/sys_template.php
@@ -315,7 +315,7 @@ function render_table($columns, $rows, $data = true)
         return info(__('No data found.'), true);
     }
 
-    $html = '<table class="table table-striped' . ($data ? ' data' : '') . '">';
+    $html = '<table class="table table-striped table-sticky-header' . ($data ? ' data' : '') . '">';
     $html .= '<thead><tr>';
     foreach ($columns as $key => $column) {
         $html .= '<th class="column_' . $key . '">' . $column . '</th>';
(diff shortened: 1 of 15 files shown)


xuwhite:
Thank you so much, I love it. the locations table header isn't sticky yet, could you fix that too?

Enterprize1:
Adding the same header stickiness to the location table required some JavaScript as the table is wrapped in a `.table-responsive` element for the horizontal scroll. This combination is not supported using just CSS (https://github.com/w3c/csswg-drafts/issues/865#issuecomment-350585274).

xuwhite:
Thank you so much it looks really good. I had a look and found even more tables with non sticky headers, would you mind to also make them sticky? 
- history
- shift types 
- sessions
- oauth (even though I hope we never need it ;D)
- schedule
- messages
- load (schedules)
- tag

and also please add this to design.twig

Enterprize1:
I finally got around to add that to all the tables, should work now

xuwhite:
Thanks.

PR #1530, opened 2025-04-18
Allow adding angels to angel type when already awaiting confirmation

MyIgel:
At the moment angels can't be confirmed for an angel type if they already requested to join.


xuwhite (approved):
code looks nice 
feature test in progress

SettRaziel:
Tested with docker dev build. Works as intended.

SettRaziel:
New Bug: Additional confirmation as supporter for this user afterwards with the same method does not work either. It seems the same issue as confirming the unconfirmed user.

PR #1532, opened 2025-04-22
make admin_arrive buttons not reload page with js magic

xuwhite:


resources/assets/js/arrive.js
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * @type {HTMLButtonElement}
+     */
+    const element = event.currentTarget;
+    const dataset = element.dataset;
+
+    const action = String(dataset.arriveAction);
+    const userId = Number(dataset.arriveUserId);
+
+    if (Number.isInteger(userId) === false) {
```
rehleinBo: Wouldn't `if (!Number.isInteger(userId))` do the job properly?

The builtin function ([link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger)) is defined te deliver a boolean result

resources/assets/js/forms.js
```diff
@@ -329,16 +329,22 @@ ready(() => {
         `
       );
 
+      console.log(element.dataset)
```
rehleinBo: debug code?

resources/assets/js/forms.js
```diff
@@ -329,16 +329,22 @@ ready(() => {
...
+      // if title of confirmation popup is empty don't open it
+      if (!element.dataset['confirm_submit_title'] || ('modalShow' in element.dataset && !element.dataset.modalShow)){
+        console.log('test')
```
rehleinBo: Also debug code?

src/Controllers/Admin/UserArriveController.php
```diff
@@ -0,0 +1,83 @@
...
+use Engelsystem\Models\User\User;
+use Psr\Log\LoggerInterface;
+
+class UserArriveController extends BaseController
+{
+    use HasUserNotifications;
+
+    /** @var array<string> */
+    protected array $permissions = [
+        'admin_arrive',
```
rehleinBo: Do this array magically control the access level via $auth?

You know what I said about PHP ^^

xuwhite: yes it does.

resources/assets/js/arrive.js
```diff
@@ -0,0 +1,147 @@
...
+    event.preventDefault();
+    event.stopPropagation();
+
+    /**
+     * @type {HTMLButtonElement}
+     */
+    const element = event.currentTarget;
+    const dataset = element.dataset;
+
+    const action = String(dataset.arriveAction);
```
rehleinBo: As far as I know a dataset value is always a string or undefined.  
See: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

So 
```suggestion
    const action = dataset.arriveAction || ''; // The ?? operator is unfortunately a TypeScript thing
```
would be more clean from my point of view. 

`String(undefined)` would result in `'undefined'`

resources/assets/js/arrive.js
```diff
@@ -0,0 +1,147 @@
...
+    const userId = Number(dataset.arriveUserId);
+
+    if (!Number.isInteger(userId)) {
+        console.error('User ID is not an integer', userId);
+        return;
+    }
+    if (action !== 'arrive' && action !== 'reset' && action !== '') {
```
rehleinBo: This suggestion is untested and I'm not 100% sure if it's proper JS... - But it would be easier to read instead of the if statement.
```suggestion
   switch(action) {
     case 'arrive':
     case 'reset':
     case '':
       break;
     default:
        console.error('Invalid action', action);
        return;
   }
```

src/Controllers/Admin/UserArriveController.php
MyIgel: looks removed ;)

resources/assets/js/arrive.js
MyIgel: not sure if it makes it better? :D
But seems to work so :shrug:

includes/pages/admin_arrive.php
```diff
@@ -121,26 +121,47 @@ function admin_arrive()
         $usr['rendered_planned_arrival_date'] = $plannedArrivalDate ? $plannedArrivalDate->format(__('general.date')) : '-';
         $usr['rendered_arrival_date'] = $arrivalDate ? $arrivalDate->format(__('general.date')) : '-';
         $usr['arrived'] = icon_bool($usr->state->arrived);
-        $usr['actions'] = form([
```
MyIgel: Imho we should keep the "form" functionality as arriving culd be done by "normal" angels

resources/assets/js/arrive.js
```diff
@@ -0,0 +1,152 @@
...
+            const arriveIconElement = tableRow.querySelector('.column_arrived').firstElementChild;
+
+            element.dataset.arriveAction = arrived ? '' : 'arrive';
+            arrivalDateElement.innerHTML = arrivalDate;
+
+            if (arrived) {
+                element.dataset.modalShow = '1';
+                handleActionButton(true, element);
+                handleArriveIcon(true, arriveIconElement);
+                // element.classList.replace(btnType.arrive, btnType.reset);
```
MyIgel: Unused code should be removed

rehleinBo:
LGTM to me now but I'm not a PHP guy ^^

MAybe should @MyIgel  have a look at the PHP parts too.

src/Controllers/Admin/UserArriveController.php
```diff
@@ -0,0 +1,83 @@
...
+        protected Translator $translator,
+    ) {
+    }
+
+
+    public function saveArrive(Request $request): Response
+    {
+        $userId = (int) $request->getAttribute('user_id');
+        /** @var User $user */
+
```
MyIgel:
```suggestion
```

src/Controllers/Admin/UserArriveController.php
```diff
@@ -0,0 +1,83 @@
...
+                ->withHeader('content-type', 'application/json')
+                ->withContent(json_encode([
+                    'state' => $user->state->arrived,
+                    'arrival_date' => $user->state->arrival_date?->format(
+                        $this->translator->translate('general.date')
+                    ) ?: '-',
+                ]));
+        }
+
+        $this->addNotification($action == 'arrive'
```
MyIgel: I don't like that formatting but it würgs so 🤷

resources/assets/js/arrive.js
```diff
@@ -0,0 +1,152 @@
+import { getCSRFToken } from './utils';
+import { ready } from './ready';
+
+ready(() => {
+    // Add plus 1 voucher click handler to all plus 1 voucher buttons
```
MyIgel: `eslint` is not happy about this file's indentation (should be only two spaces)

src/Controllers/Admin/UserArriveController.php
```diff
@@ -0,0 +1,83 @@
...
+class UserArriveController extends BaseController
+{
+    use HasUserNotifications;
+
+    /** @var array<string> */
+    protected array $permissions = [
+        'admin_arrive',
+    ];
+
+    public function __construct(
```
MyIgel: The constructor should be covered in a test too

src/Controllers/Admin/UserArriveController.php
MyIgel: This is not used by a test with matching `@covers` annotation

includes/pages/admin_arrive.php
MyIgel: (it ain't outdated, just part of an older review ;))

resources/assets/js/forms.js
```diff
@@ -287,72 +273,75 @@ ready(() => {
...
+    });
 
-      /**
-       * After the modal has been shown, focus on the "Submit" button in the modal
-       * so that it can be confirmed with "Enter".
-       */
-      modal.addEventListener('shown.bs.modal', () => {
-        modalSubmitButton.focus();
-      });
+    const modalSubmitButton = modal.querySelector('[data-submit]');
```
rehleinBo, line 323: I really am a big fan of element IDs. This is always fuzzy and can maybe become a problem.

resources/assets/js/forms.js
```diff
@@ -287,72 +273,75 @@ ready(() => {
...
-      const modalSubmitButton = modal.querySelector('[data-submit]');
-      modalSubmitButton.addEventListener('click', () => {
-        element.type = oldType;
-        element.click();
-      });
+        </div>
+      `
+    );
+
+    const modal = document.getElementById('confirmation-modal');
```
rehleinBo, line 317: Instead of a static ID I would suggest a dynamic one. 
`<paranoidmode>`
If anything goes wrong afterward you maybe have two or more modals in your DOM and every instance will have the same ID.
Why not use [`randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) or at least the current timestamp?
`</paranoidmode>`

resources/assets/js/forms.js
```diff
@@ -287,72 +273,75 @@ ready(() => {
...
-                <div class="modal-footer">
-                  <button type="button" class="${element.className}"
-                    autofocus
-                    title="${element.title}" data-submit="">
-                    ${element.dataset.confirm_button_text ?? element.innerHTML}
-                  </button>
-                </div>
+    event.preventDefault();
+
+    document.getElementById('confirmation-modal')?.remove();
```
rehleinBo, line 290: Wouldn’t it much cleaner if the modal is removed in the click routine. This would also spare another query select.

This would also work perfecly with my other comments ;)

resources/assets/js/forms.js
```diff
@@ -435,3 +424,55 @@ ready(() => {
...
+        method: 'POST',
+        headers: {
+          Accept: 'application/json',
+          'X-CSRF-TOKEN': data.get('_token'),
+        },
+        body: data,
+      });
+
+      if (!response.ok) {
+        console.warn('Broken response, submitting form anyways');
```
rehleinBo, line 456: Why that?
As a fallback?

This log is not very helpful - At least the failed action and the response code should be included.

resources/assets/js/forms.js
```diff
@@ -435,3 +424,55 @@ ready(() => {
...
+        form.submit();
+        return;
+      }
+
+      const body = await response.json();
+      arrival.innerText = body.arrival_date;
+      const newButton = new DOMParser().parseFromString(body.button, 'text/html').body.firstChild;
```
rehleinBo, line 463: That's kind of dangerous. You create a DOM element from maybe spoofed text.
I know it's paranoid but...

`const btn = document.createElement('button')` would do the job also. It would be more JS code to set all the button attributes, but more save

resources/assets/js/forms.js
```diff
@@ -435,3 +424,55 @@ ready(() => {
...
+      button.replaceWith(newButton);
+
+      if (body.state === 'arrived') {
```
rehleinBo, line 466: Would'nt it be cleaner to set the classes in the two spans instead of replacing the whole HTML?

This wold also work if something is changend on the PHP side of life

resources/assets/js/forms.js
MyIgel: In this case there are many forms (as in: number of angels availabe) so any ID would not be feasible

resources/assets/js/forms.js
MyIgel: It makes the state tracking easier (and is here only changed because I removed two spaces of indentation so imho out of scope)

resources/assets/js/forms.js
MyIgel: its the most "useful" fallback indeed (besides of showing an error message which is currently not implemented)

resources/assets/js/forms.js
MyIgel: it is much more code as you have to include all the translations and set them to attributes -> this is kind of legacy code anyways so the "easier" way.

ALternative would be to remove any magic, parse the whole returning html and just pick out the whole row to replace it 1:1 -> that makes no difference here imho?

resources/assets/js/forms.js
MyIgel: its not only the classes but also the content so "replacing" is technically more correct? Or am i missing something?

PR #1533, opened 2025-04-23
Fixing the anoying small stuff

MyIgel:


xuwhite:
please provide a description

includes/view/User_view.php
```diff
@@ -683,6 +683,11 @@ function User_view(
                                 form_hidden('user', $user_source->id),
                                 form_submit('submit', icon('house') . __('user.arrive'), '', false),
                             ], url('/admin-arrive'), 'float:left') : '',
+                        $auth->can(['admin_log', 'logs.all']) ?
```
xuwhite: I would prefer moving this button to the right end of the button group to e.g. make it less confusing if a buerocrat wants to show something to a shico

.gitlab-ci.yml
```diff
@@ -120,6 +120,7 @@ generate-version:
           (git describe --abbrev=0 --tags | tr -d '\n')\
           && echo "-${CI_COMMIT_REF_NAME}+${CI_PIPELINE_ID}.${CI_COMMIT_SHORT_SHA}"\
         )\
+        || echo "0-${CI_COMMIT_REF_NAME}+${CI_PIPELINE_ID}.${CI_COMMIT_SHORT_SHA}"\
```
xuwhite, line 123: I'm sorry I can't test this

includes/view/User_view.php
MyIgel: I could move it to the second or third position from the right but work logs are more important and thus should be "easier findable" than a random log link

PR #1535, opened 2025-04-27
user-shifts: collapse Location/Type/Occupation Filter by default

Lotterleben:
We've had the problem that people don't notice the shifts view because it is pushed out of the frame by extensive location lists. this PR quick-fixes this by collapsing this filter by default. while we're at it, it also attempts to clarify the language around filters.

before/after:

Screenshot 2025-04-27 at 21 18 07

↓
Screenshot 2025-04-27 at 21 06 34

### DIFF
--- a/resources/views/pages/user-shifts.html
+++ b/resources/views/pages/user-shifts.html
@@ -3,6 +3,7 @@
     <input type="hidden" name="tag" value="%tag_id%">
     <div class="row">
         <div class="col-md-6">
+            <h1>%title% %add_link%</h1>
             <div class="row align-items-center">
                 <div class="col-12 col-xxl-auto row pe-xxl-0">
                     <div class="col-auto pe-xxl-0">%start_select%</div>
@@ -62,9 +63,6 @@
                 </div>
                 <div class="btn-group mb-1" role="group">%random%</div>
                 <div class="btn-group mb-1" role="group">%dashboard%</div>
-                <div class="btn-group mb-1" role="group">
-                    %dashboard%
-                </div>
             </div>
             <div class="row d-print-none">
                 <div class="col-md-12">
@@ -74,7 +72,7 @@
         </div>
         <div class="col-md-6">
             <button
-                class="btn btn-info mb-3 d-print-none"
+                class="btn btn-info btn-sm mb-3 d-print-none"
                 type="button"
                 data-bs-toggle="collapse"
                 data-bs-target="#collapseShiftsFilterSelect"
--- a/includes/pages/user_shifts.php
+++ b/includes/pages/user_shifts.php
@@ -383,7 +383,7 @@ function view_user_shifts()
                 'tag_id'  => $tagId,
                 'shifts_table'  => $shiftCalendarRenderer->render(),
                 'ical_text'     => div('mt-3', ical_hint()),
-                'filter'        => __('Look For Shifts'),
+                'filter'        => __('Filter shifts'),
                 'filter_toggle' => __('shifts.filter.toggle'),
                 'expand_toggle_title' => __('shifts.toggle.title'),
                 'set_yesterday' => __('Yesterday'),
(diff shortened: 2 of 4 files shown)


resources/views/pages/user-shifts.html
```diff
@@ -60,14 +63,14 @@ <h1>%title% %add_link%</h1>
             </div>
         </div>
         <div class="col-md-6">
-            <button class="btn btn-info btn-sm mb-3 d-print-none" type="button"
+            <button class="btn btn-info mb-3 d-print-none" type="button"
```
xuwhite, line 66: Can you tell me why you want this button to be bigger? Since this takes up more space on the page and moves the shifts even more down on the screen, this looks worse to ma especially using a small screen.

resources/views/pages/user-shifts.html
```diff
@@ -1,8 +1,8 @@
 <form class="form-inline" action="">
     <input type="hidden" name="p" value="user_shifts">
+    <h1>%title% %add_link%</h1>
```
xuwhite, line 3: I like the look of the title being out of the rows, but sadly it takes up way more space on the page if the filters are shown, which moves the shifts down. That's why this sadly feels impractical to me.

resources/views/pages/user-shifts.html
```diff
@@ -52,6 +52,9 @@ <h1>%title% %add_link%</h1>
                 <div class="btn-group mb-1" role="group">
                     %dashboard%
                 </div>
+                <div class="btn-group mb-1" role="group">
```
xuwhite, line 55: We now have two public dashboard buttons 😄 Could you reduce it to one please.

xuwhite (changes requested):
Thanks for the PR. I like the idea of making it easier for angels to see shifts by default collapsing the filters and making the language of the Filter button more clear. Some of your design changes make it harder to work with shown filters, I'm interested in your thoughts why you prefer this way. Additionally I would prefer a shorter Filter button text and made some suggestions.

resources/views/pages/user-shifts.html
Lotterleben: ah, my reasoning was that the different button sizes add visual noise and a hierarchy that doesn't exist functionally (at least from my view), making the page harder to skim. I think once a user *knows* that there's a table below the long list, it's less critical if the list pushes the table down a bit further. I think in the long run, maybe the list could be made scrollable instead of displaying it all at once (similar to filter options in fashion online shops– they're really hard to screenshot because the scroll bar keeps disappearing obviously) to resolve the problem even better, but that's for another PR :D

resources/views/pages/user-shifts.html
Lotterleben: whoops, that wasn't on purpose :D will do!

resources/views/pages/user-shifts.html
Lotterleben: >  the list could be made scrollable instead of displaying it all at once 

orrrrr they could be moved to the side as an "overlay" and be made toggleable there?

IMG_9830

but now I'm just throwing ideas around, that's all for a future PR.

xuwhite: I like your ideas. But since we still have the long lists and there are people that have the filters expanded all the time, I would prefer the smaller button for now. Could you open an issue and move your ideas to that issue (if you want you can start the title with "[RFC]", which stands for "request for comments", so that more people can share their ideas and opinions)?

xuwhite: I made these filter selections scrollable in my fixes PR #1538

xuwhite:
Hi, did you have time to implement the changes?

resources/views/pages/user-shifts.html
```diff
@@ -86,7 +87,7 @@ <h1>%title% %add_link%</h1>
                 <i class="bi bi-chevron-bar-expand" title="%expand_toggle_title%"></i>
             </button>
 
-            <div class="collapse show d-print-none" id="collapseShiftsFilterSelect">
+            <div class="collapse d-print-none" id="collapseShiftsFilterSelect">
```
MyIgel, line 89: That needs a change in the filter JS as filters are now hidden again after page relaod

PR #1536, opened 2025-04-29
add api endpoints for updating vouchers

MrKrisKrisu:
This PR adds two new API endpoints:
- `POST /admin/users/{id}/voucher`: Allows setting the got_voucher value directly.
- `POST /admin/users/{id}/voucher/increment`: Allows incrementing or decrementing the got_voucher value by a given amount (as long as the resulting value is not negative).

These endpoints are restricted to users with the api and voucher.edit permissions.

I've also deactivated CSRF checks on the `/api` path.

I tried to follow the structure of existing tests, but I couldn't find any API tests related to my case. I attempted to create some tests similar to how I would in Laravel, but ran into issues due to unfamiliarities with the test setup. I’d be happy if there is any input on this...


MrKrisKrisu:
I've added some tests - can you please check whether I have done this according to your requirements? (@xuwhite @MyIgel)

MyIgel:
Hi @MrKrisKrisu, implementing a POST'able API is planned for the near future and needs prior conceptualisation and a more generic implementation which must correspond to the existing API schema.
Adding some random endpoints to the admin sub-path would not solve that issue and looks pretty odd as it does not integrate in the current api endpoints.
I would suggest leving this MR for as long as the POST API is not implemented and then refactor it to match the general implementation.

PR #1537, opened 2025-04-29
show `got_voucher` and `got_goodie` if permissions apply

MrKrisKrisu:


src/Controllers/Api/Resources/UserResource.php
```diff
@@ -24,5 +23,16 @@ public function toArray(): array
             'contact' => $this->model->contact->only(['dect', 'mobile']),
             'url' => url('/users', ['action' => 'view', 'user_id' => $this->model->id]),
         ];
+
+        // if current user has permission: inject got_voucher and got_goodie
+        $currentUser = auth()->user();
+        if($currentUser->privileges->where('name', 'voucher.edit')->first() || $currentUser->id === $this->model->id) {
```
MyIgel, line 29: Please use `phpcbf` to fix formatting errors:
```
FILE: /var/www/src/Controllers/Api/Resources/UserResource.php
-------------------------------------------------------------------------
FOUND 2 ERRORS AND 1 WARNING AFFECTING 2 LINES
-------------------------------------------------------------------------
 29 | ERROR   | [x] Expected 1 space after IF keyword; 0 found
 32 | ERROR   | [x] Expected 1 space after IF keyword; 0 found
 32 | WARNING | [ ] Line exceeds 120 characters; contains 123 characters
-------------------------------------------------------------------------
PHPCBF CAN FIX THE 2 MARKED SNIFF VIOLATIONS AUTOMATICALLY
-------------------------------------------------------------------------
```

MyIgel, line 29:
```suggestion
        if(auth()->can('voucher.edit') || $currentUser->id === $this->model->id) {
```

`auth()->can('voucher.edit')` should be used to check permissions

src/Controllers/Api/Resources/UserResource.php
```diff
@@ -24,5 +23,16 @@ public function toArray(): array
...
+            $data['got_voucher'] = $this->model->state->got_voucher;
+        }
+        if($currentUser->privileges->where('name', 'user.goodie.edit')->first() || $currentUser->id === $this->model->id) {
```
MyIgel, line 32:
```suggestion
        if(auth()->can('user.goodie.edit') || $currentUser->id === $this->model->id) {
```
Same here of course

MyIgel (changes requested):
Hi, thanks for the API extension!

It just needs fixed formatting, a small cleanup and tests (currently it breaks them as it returns an unspecified format / not in the `openapi.yml`).

(As the API is currently only available to ShiCos it even would be fine to just add them to all instances and implement the optional part later (esp. as that makes the api output more consistent / does not show it on the users data); if not the fields should eventually be `null` if permissions do not match which would even make the api definition easier)

PR #1540, opened 2025-05-25
make shift filter selections scrollable

xuwhite:
* make shifts filter selection scrollable and add max height too in user-shifts
Screenshot 2025-06-01 at 12 04 01
* because @MyIgel wanted it: there now are two sections one with own angel types and one with others, separated with a line 
  * for the record: I hate this sorting and its confusing to me

### DIFF
--- a/includes/pages/user_shifts.php
+++ b/includes/pages/user_shifts.php
@@ -194,7 +194,14 @@ function load_types()
                         NOT `user_angel_type`.`confirm_user_id` IS NULL
                         OR `user_angel_type`.`id` IS NULL
                     )
-                ) AS `enabled`
+                ) AS `enabled`,
+                (
+                    `user_angel_type`.`id` IS NOT NULL
+                    AND (
+                        `angel_types`.`restricted`=0
+                        OR `user_angel_type`.`confirm_user_id` IS NOT NULL
+                    )
+                ) AS `own`
             FROM `angel_types`
             LEFT JOIN `user_angel_type`
                 ON (
@@ -204,7 +211,7 @@ function load_types()
             . ($isShico ? '' :
             'WHERE angel_types.hide_on_shift_view = 0
                 OR user_angel_type.user_id IS NOT NULL ') .
-            'ORDER BY `angel_types`.`name`
+            'ORDER BY `own` DESC, `angel_types`.`name`
         ',
         [
             $user->id,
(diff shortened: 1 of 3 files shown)


includes/pages/user_shifts.php
```diff
@@ -204,7 +208,7 @@ function load_types()
             . ($isShico ? '' :
             'WHERE angel_types.hide_on_shift_view = 0
                 OR user_angel_type.user_id IS NOT NULL ') .
-            'ORDER BY `angel_types`.`name`
+            'ORDER BY `user_angel_type`.`user_id` IS NOT NUll AND `user_angel_type`.`confirm_user_id` IS NOT NULL DESC, `angel_types`.`name`
```
MyIgel: Technically this should check if the confirmation is even needed (or remove the confirm check anyways) ;)

resources/assets/themes/base.scss
```diff
@@ -208,6 +208,17 @@ table .border-bottom {
   }
 }
 
+/* make selections scrollable and add a max height of 4.5 entries */
+.selection {
+  overflow-y: auto;
+  max-height: ($form-check-min-height + $form-check-margin-bottom) * 4.5;
+}
+
+/* add devider for confusing sorting of angel types selection */
```
MyIgel:
```suggestion
/* add divider between sorting groups of angel types selection */
```

resources/assets/themes/base.scss
```diff
@@ -208,6 +208,17 @@ table .border-bottom {
   }
 }
 
+/* make selections scrollable and add a max height of 4.5 entries */
+.selection {
```
MyIgel: Imho eventually we should be more specific in the css selector here (eventually include "for shifts page?")

MyIgel (changes requested):

includes/pages/user_shifts.php
xuwhite: ???

MyIgel: meaning: the user_angel_type.confirm_user_id is only set if the angel type is restricted which introduces additional sorting for "not restricted" angel types of the user; this could be solved in the query imho but also possible in code if needed

(also the order by clause should be formatted/put on multiple lines here ;))

xuwhite: `ORDER BY 'own' DESC, 'angel_types'.'name'` better?

PR #1547, opened 2025-07-19
Add Dutch translation

shizmob:
This is my initial pass at a Dutch (`nl_NL`) translation for engelsystem, hopefully useful in the leadup to WHY2025. :-)

Here's some of the involved design decisions:
- Native Dutch terms are used where possible, but for certain technical and non-technical terms that are used by the community in English already, the English usage is retained for clarity and consistency with existing usage. Notable examples:
  * Angel
  * Heaven
  * API-key
  * Login
- Login vs. registration vs. sign-up is a difficult one. The following terms were chosen:
  * Registration (as in, making an account): registratie/registreren
  * Login (as in, starting a session on an existing account): log in/inloggen
  * Sign up (as in, indicating willingness to perform a shift): aanmelden
  * Sign off (as in, indicating lack of willingness to perform a shift): afmelden
- Other notable chosen terms:
  * Schedule: rooster
  * Shift: dienst
  * Job: taak
  * Enter (as in information): invullen
  * Adding/Creating: aanmaken
  * Editing: bijwerken (if information, think updating), veranderen (if setting, think changing)
  * Deleting: verwijderen
  * Work log entry: werklogvermelding

The `.po` file was created by merging the possible entries from the `en_US` and `de_DE` `.po` (as neither seemed to be a full superset of the other).

Looking forward to your comments and thoughts!

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -403,6 +403,7 @@ return [
     'locales'                 => [
         'de_DE' => 'Deutsch',
         'en_US' => 'English',
+        'nl_NL' => 'Nederlands',
     ],
 
     // The default locale to use
--- a/resources/lang/nl_NL/additional.po
+++ b/resources/lang/nl_NL/additional.po
@@ -1,0 +1,344 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: Engelsystem 2.0\n"
+"Language-Team: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"Last-Translator: Shiz <hi@shiz.me>\n"
+"Language: nl_NL\n"
+
+msgid "auth.not-found"
+msgstr "Gebruiker bestaat niet of het opgegeven wachtwoord komt niet overeen. Probeer het nogmaals. "
+"Neem contact op met Heaven bij verdere problemen."
+
+msgid "validation.password.required"
+msgstr "Het wachtwoord is vereist."
+
+msgid "validation.password.length"
+msgstr "Het ingevoerde wachtwoord is te kort."
+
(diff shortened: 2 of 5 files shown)


MyIgel:
Hi @shizmob,
oof, that's definitely one of those PR's that should have been an issue beforehand to discuss the general points around it :D (esp. as there was another one, #1207).
Generally a new language is of course appreciated and great to have! But on the other side it needs continuous upkeep when the translation changes, is used in another way than expected or new parts of the Engelsystem are added.
As we keep the main branch mostly "production ready stable" so should be any texts/translations. At the moment no one of the Engelsystem dev team speaks Dutch and thus we can't either provide new translations nor even have any clue about if it is somehow consistent / useful / useable.
Any language needs (at least?) one to two, (better three?) people that can have a look at it and to merge any updates / keep up with changes.
Is it possible that this would happen? If yes i would be happy to merge it here or else we have to remove it later again (as done with pt_br at some point) as it will be outdated and thus the whole system just a mashup between Dutch and English.
For the WHY2025 I am more than happy to merge it though (in the hope that any changes until that event can be made on-site if needed ;))

To come to some of your points:
* The German .po file should contain all texts that can be translated, if not that’s actually a bug which should be fixed :D
* English .po files should be a subset of it as there are some (legacy) texts that are not identified by a translation key (dotted text) but by the English text itself and thus don't need translation

MyIgel:
As no volunteers where volunteering for this translation I'll close the MR for now; if someone want to contribute / maintain it feel free to comment here / create an updated one :)

shizmob:
Hi! When I made this MR I was in the middle of a rush for WHY2025, so the follow-up messages got a bit lost in the chaos after. Apologies!

We've successfully deployed this translation at WHY2025! I'm also very much willing to keep the nl_NL translation up-to-date with new texts when those become available.
Not sure who'd else would be willing to to reach the two people threshold, but maybe someone (like from the original MR) will hop in once it's a reality in upstream? :)

PR #1548, opened 2025-07-21
Schedule: Fixed room patching created duplicates

MyIgel:
The current schedule import logic patches only the last room instance to the new name with previous instances of other days not being changed.
This leads to duplicated room names (full length and shortened)

### DIFF
--- a/src/Controllers/Admin/ScheduleController.php
+++ b/src/Controllers/Admin/ScheduleController.php
@@ -569,7 +569,7 @@ class ScheduleController extends BaseController
 
     protected function patchSchedule(Schedule $schedule): Schedule
     {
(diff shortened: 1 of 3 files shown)


xuwhite:
Since we had so many problems with long room names could you please add a test for long room names?

xuwhite:
CR: LGTM 🙂

MyIgel:
The room name change itself was already covered & the diplicated rooms are now tested too with the changed test in this PR, what test du you want to have in addition @xuwhite ?

xuwhite:
> The room name change itself was already covered & the diplicated rooms are now tested too with the changed test in this PR, what test du you want to have in addition @xuwhite ?

true Ignore my comment

PR #1557, opened 2025-08-26
Shift Removal Explanation

ski7777:
Send explanation to users who are removed from a shift

### DIFF
--- a/includes/controller/shift_entries_controller.php
+++ b/includes/controller/shift_entries_controller.php
@@ -336,8 +336,14 @@ function shift_entry_delete_controller()
     }
 
     if ($request->hasPostData('delete')) {
+        $reason = $request->postData('reason');
+        if (empty($reason)) {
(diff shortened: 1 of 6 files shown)


includes/controller/shift_entries_controller.php
```diff
@@ -336,8 +336,14 @@ function shift_entry_delete_controller()
     }
 
     if ($request->hasPostData('delete')) {
+        $reason = $request->postData('reason');
+        if (empty($reason)) {
+            error(__('Please provide a reason for removing the shift entry.'));
```
MyIgel, line 341: This should be covered by a required attribute on the form and thus any error here would only need something like `Error: reason missing` ;)

On the other side i don't see the need to enforce a reason everywhere; just let the shico decide if they want to use it (and log it when it exists too).

MyIgel:
Hi, any updates?

PR #1555, opened 2025-08-26
change time format in worklog

ski7777:
Changes as requested in #1552 
Set format to hh:mm at:

- goodie score in worklog
- worklog sum
- worklog entries

PR #1562, opened 2025-10-17
show worked time as duration string

### LINKED ISSUE #1552: Show worked time in HH:MM instead of decimal
Currently the work-log hours and sum are shown in decimal time, which is a bit confusing to convert to regular time.
Image

It should show the times in HH:MM, so 09:55h instead of 9.92h

olebittner:
This PR formats the durations on the /user page as duration strings, resolves #1552 

I used CarbonIntervals to format the time, this allows adjusting the format by patching the translations.
I implemented ``%hh %Imin`` (de) and ``%hh %Im`` (en) as defaults as @xuwhite suggested, resulting in:
grafik
grafik

while still allowing to change the translation to ``%H:%I`` resulting in:
grafik

I didn't adjust the Adjust the active angle page, as the format is quite different there, if you want me to adjust it, just let me know.

### DIFF
--- a/includes/controller/users_controller.php
+++ b/includes/controller/users_controller.php
@@ -1,6 +1,8 @@
 <?php
 
+use Carbon\CarbonInterval;
 use Engelsystem\Database\Db;
+use Engelsystem\Helpers\Carbon;
 use Engelsystem\Helpers\Goodie;
 use Engelsystem\Models\AngelType;
 use Engelsystem\Models\Shifts\ShiftEntry;
@@ -189,7 +191,10 @@ function user_controller()
         auth()->resetApiKey($user_source);
     }
 
-    $goodie_score = sprintf('%.2f', Goodie::userScore($user_source)) . '&nbsp;h';
+    $goodie_score = Carbon::formatDuration(
+        CarbonInterval::minutes(round(Goodie::userScore($user_source) * 60)),
+        __('general.duration')
+    );
     if ($user_source->state->force_active && config('enable_force_active')) {
         $goodie_score = '<span title="' . $goodie_score . '">' . __('user.goodie_score.enough') . '</span>';
     }
--- a/src/Helpers/Carbon.php
+++ b/src/Helpers/Carbon.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Engelsystem\Helpers;
 
+use Carbon\CarbonInterval;
+
 class Carbon extends \Carbon\Carbon
 {
     public const DATETIME_LOCAL = '!Y-m-d\TH:i';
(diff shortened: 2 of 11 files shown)


olebittner:
my bad @ski7777 I just went of the "Available" label on the issue and just now noticed you already started work in #1555 if you want to go ahead I will close my PR

olebittner:
grafik
grafik

olebittner:
this is ready for test/merge @xuwhite

includes/controller/users_controller.php
```diff
@@ -188,7 +189,9 @@ function user_controller()
         auth()->resetApiKey($user_source);
     }
 
-    $goodie_score = sprintf('%.2f', Goodie::userScore($user_source)) . '&nbsp;h';
+    $goodie_score = CarbonInterval::minutes(round(Goodie::userScore($user) * 60))
```
xuwhite: If you want you could add a function to Helpers/Carbon named something like formatDuration

olebittner:
>  I think it could improve readability if only hours were shown when minutes = 00

Hmm, not sure. I like it when numbers align, so I'd align to the right but keep zeros.

quick spreadsheet to visualize it:
grafik
I like the left one better 😄 (also would you hide hours when they are 0, like in the last line?)

olebittner:
@xuwhite 
> sadly the sorting on active angels is broken now

Good news: sorting was never broken.
Bad news: the durations were not correct. 😓 

I misunderstood the CarbonInterval documentation. The strings did only show the remainder of hours, the rest would have been shown as days, weeks and so on but were not displayed. But its fixed now.

olebittner:
grafik

I just implemented the right-align for now, I also centered to boolean columns. But I'll revert it if you don't like it 👼

src/Helpers/Carbon.php
```diff
@@ -47,6 +47,6 @@ public static function formatDuration(CarbonInterval $interval): string
         $hours = floor($interval->totalHours);
         $minutes = $interval->minutes;
 
-        return __('general.duration.format', [$hours, $minutes]);
+        return sprintf(__('general.duration.format'), $hours, $minutes);
```
olebittner: @xuwhite I chose to use sprintf over the translators build-in replacement function, to allow for unit tests. Hope this is fine.

MyIgel: That's not the intension of the translation feature here ;)

In tests you can use the `mockTranslator` from our `TestCase` to check for any replacements / replace it in a testable way.

olebittner: not sure if you are saying this is ok or not 😓 

Im already using ``mockTranslator`` but if I use ``__('general.duration.format', [$hours, $minutes]);`` and ``mockTranslator`` then the whole return value is mocked, which defeats the use of the testcase.

On the other hand I might be able to extract the passed parameter array via ``mockTranslator`` if you prefer that.

MyIgel: The mock calls a callback if you set it (you can have a look at the `BarChartTest` on how to use it ;)).
Extracting it would be the "proper" way (although it is currently not done enywhere else -> its on my ToDo list to use it more at some point)

PR #1564, opened 2025-10-18
vertically center buttons in titles

### LINKED ISSUE #1241: center buttons in titles
look at the plus buttons
Screenshot 2023-10-25 at 14 28 44

olebittner:
resolves #1241 by centering buttons vertically in titles.

``h1``s containing buttons are changed to flexboxes with ``align-items-end``. Buttons are then assigned the ``align-self-center`` class.

Before:
Edit Team - Title - Old

After:
Edit Team - Title - New

As you can see this causes the decenders of the letters to "hang" into the bottom margin. But this actually brings it into line with headings without buttons.

Before:
Meetings - Title - Old

After:
Meetings - Title - New

No Button:
Meetings - Title - No Button

---
I chose to add the ``align-self-center`` class to every button (in titles) seperatly instead of adding it to the defaults in the ``base.twig`` because I didn't want to assume that every button will be inside a flexbox.

### DIFF
--- a/resources/assets/themes/base.scss
+++ b/resources/assets/themes/base.scss
@@ -445,6 +445,17 @@ blockquote {
   padding-left: 0.5em;
 }
 
+/* center buttons in titles */
+h1 {
+  display: flex;
+  gap: 0.5rem;
+  align-items: flex-end;
+
+  & .btn {
+    align-self: center;
+  }
+}
+
 /* Hide the arrow up/down buttons rendered by the browser in the input field */
 /* Chrome, Safari, Edge, Opera */
 input[type='number']::-webkit-outer-spin-button,


MyIgel:
Wouldn't it be easier to just use additional css to fix the buttons? In that wa one has to remember to add additional classes on every title which is a bad idea -> that should be handled by the corresponding macro or explicitly defined in css

olebittner:
good point, I'll give this another try

olebittner:
much better solution, thank you for the hint.

Since titles have no extra class I applied the css to all ``<h1>`` elements. Currently all ``<h1>`` elements are titles, so it doesn't make a difference at the moment and most themes assume the same, I guess. Just wanted to point it out.

@MyIgel please take another look, when you have the time.

PR #1565, opened 2025-10-18
Ignore shifts without angels when populating event days

### LINKED ISSUE #1351: empty dates in romm-shifts-date dropdown
When angeltypes are importet by schedule from shifttype and the shifttype has no required angels but the room does empty dates are displayed in room-shifts-date-dropdown and the rooms are showed in the shifts overview but contain no shifts

olebittner:
resolves #1351 by checking if shifts need any angle types before adding a day to the array.

It is difficult for me to estimate how computationally expensive this is and if it is worth it. ~It will only be checked each day until the first shift which needs angels is found. But since every call to ``NeededAngelTypes_by_shift()`` will trigger three DB queries, there will be atleast three queries for every day on which this location has as any shifts. And if there is a day with many shifts which don't require any angels, this will trigger many additional DB queries.~

### DIFF
--- a/includes/controller/locations_controller.php
+++ b/includes/controller/locations_controller.php
@@ -24,10 +24,9 @@ function location_controller(): array
     $request = request();
     $location = load_location();
 
-    $all_shifts = $location->shifts->sortBy('start');
+    $days_list = Days_by_Location_id($location->id);
     $days = [];
-    foreach ($all_shifts as $shift) {
-        $day = $shift->start->format('Y-m-d');
+    foreach ($days_list as $day) {
         if (!isset($days[$day])) {
             $days[$day] = dateWithEventDay($day);
         }
--- a/includes/model/Shifts_model.php
+++ b/includes/model/Shifts_model.php
@@ -351,6 +351,60 @@ function NeededAngeltype_by_Shift_and_Angeltype(Shift $shift, AngelType $angelty
 }
 
 /**
+ * returns all days with shifts needing angels for a location
+ *
+ * @param int $location_id
+ * @return list<string>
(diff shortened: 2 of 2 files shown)


olebittner:
I think this makes more sense. Days are now no longer determined by iterating over the shifts but instead by a single sql query similiar to ``NeededAngeltypes_by_ShiftsFilter()`` but will return only dates (which have shifts that require angels).

On the downside I think this makes the code harder to read and strictly speaking its new functionality in the ``includes`` section. Also not sure if ``Shifts_model.php`` is the correct file for the query.

MyIgel:
It is correct enough & legacy code anyways so thats fine; needs to be replaced with a "fancier" variation at some point (i am currently trying to implement a wrapper around the shifts rendering so it might be the case soon :tm:)

MyIgel:
This does not resolve the issue completely:

> the rooms are showed in the shifts overview but contain no shifts

In addition to that each angeltype also has the day list so should be afaik affected too?

olebittner:
> In addition to that each angeltype also has the day list so should be afaik affected too?

No because for the angeltype view the list is pulled via [``Shifts_by_angeltype()``](https://github.com/engelsystem/engelsystem/blob/5a3056715f09799482ef51b25e143d8afc5473c6/includes/controller/angeltypes_controller.php#L228)

Which already uses a smiliar sql see: https://github.com/engelsystem/engelsystem/blob/5a3056715f09799482ef51b25e143d8afc5473c6/includes/model/Shifts_model.php#L20

MyIgel:
Fair point, I remembered that wrong :D

PR #1568, opened 2025-10-20
Allow joining angel type via QR codes with JWT tokens

### LINKED ISSUE #1501: Angel type confirmation as self-service
### Background

We had the large evacuation helper introduction on 38C3 in which attendance was checked in a vey cumbersome manual process involving lots of paper.

Also, in other big introduction meetings, a supporter needs to record who is there and needs to confirm everyone manually which takes time and is error-prone.

### Idea

Have some way, e.g. via a QR code that can be shown during/at the end of a meeting which contains a special link that confirms people for an angel type when visited.

This way, the link could be part of the presentation or handed out at the end, making the process a lot easier.

Yes, this can be abused, but chances are good that it will get noticed later when people show up to shifts clueless and violators can be banned, which hopefully is enough to dissuade people from trying.

### Criteria

The link needs to contain some key/secret so it can not be guessed easily. Ideally, the link has a limited lifetime so people can't easily send it to their friends who then would get confirmed without having visited the meeting.

Gold standard: TOTP-like mechanism that changes  the token in the URL frequently, e.g. every 30 seconds and a view that shows a QR code and updates automatically when the token changes. This way, scanning the code and visiting the link immediately will work but sending the link to someone else is tricky.

MyIgel:
* Allows the generation of qr codes for a predefined time span via JWT tokens, closes #1501
* Adds `hasPermission` method to controllers
* Added `qr` function to render QR codes in templates

### DIFF
--- a/includes/view/AngelTypes_view.php
+++ b/includes/view/AngelTypes_view.php
@@ -284,6 +284,15 @@ function AngelType_view_buttons(
             '',
             __('form.edit')
         );
+        if (config('app_key') && config('join_qr_code', true)) {
+            $buttons[] = button(
+                url('/angeltypes/' . $angeltype->id . '/qr'),
+                icon('qr-code'),
+                '',
+                '',
+                __('general.qr')
+            );
+        }
     }
     if ($admin_angeltypes) {
         $buttons[] = button(
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -1307,6 +1307,9 @@ msgstr "Alle löschen"
 msgid "form.updated"
 msgstr "Aktualisiert"
 
+msgid "form.generate"
+msgstr "Generieren"
+
 msgid "form.cancel"
 msgstr "Abbrechen"
 
@@ -1922,6 +1925,21 @@ msgid "angeltypes.hide_on_shift_view.info"
 msgstr "Wenn ausgewählt, können nur Admins und Mitglieder des Engeltyps auf der "
 "Schicht Seite die Filteroption für diesen Engeltyp sehen."
 
+msgid "angeltypes.qr"
+msgstr "Engeltyp \"%s\" beitreten QR Code"
+
+msgid "angeltypes.qr.expires"
+msgstr "Läuft ab %c"
+
+msgid "angeltypes.qr.expired"
+msgstr "Abgelaufen"
+
+msgid "angeltypes.qr.time"
+msgstr "Lebensdauer (Minuten)"
+
+msgid "angeltypes.qr.time.info"
+msgstr "Der Link ist für die gesamte angegebene Zeitspanne gültig und kann nicht zurückgezogen werden!"
+
 msgid "location.location"
 msgstr "Ort"
 
@@ -2007,6 +2025,9 @@ msgstr "Anzahl"
 
 msgid "general.created_at"
 msgstr "Erstellt am"
+
+msgid "general.url"
+msgstr "URL"
 
 msgid "shifts.history"
 msgstr "Schichten Historie"
--- a/resources/lang/en_US/default.po
+++ b/resources/lang/en_US/default.po
@@ -121,6 +121,9 @@ msgstr "Delete"
 msgid "form.delete_all"
 msgstr "Delete all"
 
+msgid "form.generate"
+msgstr "Generate"
+
 msgid "form.updated"
 msgstr "Updated"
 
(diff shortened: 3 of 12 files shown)


includes/view/AngelTypes_view.php
```diff
@@ -284,6 +284,15 @@ function AngelType_view_buttons(
             '',
             __('form.edit')
         );
+        if (config('app_key') && config('join_qr_code', true)) {
```
xuwhite, line 287: I can't find the config 'join_qr_code' in config.php

resources/lang/en_US/default.po
```diff
@@ -118,6 +118,9 @@ msgstr "Delete"
 msgid "form.delete_all"
 msgstr "Delete all"
 
+msgid "form.generate"
+msgstr "Generieren"
```
xuwhite:
```suggestion
msgstr "Generate"
```

resources/lang/de_DE/default.po
```diff
@@ -1913,6 +1916,15 @@ msgid "angeltypes.hide_on_shift_view.info"
 msgstr "Wenn ausgewählt, können nur Admins und Mitglieder des Engeltyps auf der "
 "Schicht Seite die Filteroption für diesen Engeltyp sehen."
 
+msgid "angeltypes.qr"
+msgstr "Engeltyp %s beitreten"
+
+msgid "angeltypes.qr.time"
+msgstr "Gültigkeit (Minuten)"
```
xuwhite:
```suggestion
msgstr "Lebensdauer (Minuten)"
```

resources/lang/en_US/default.po
```diff
@@ -778,6 +781,15 @@ msgid "angeltypes.hide_on_shift_view.info"
 msgstr "If checked only admins and members of the angel type "
 "can see the filter option for this angel type on the shifts page"
 
+msgid "angeltypes.qr"
+msgstr "Join angel type %s"
+
+msgid "angeltypes.qr.time"
+msgstr "Lifetime (Minutes)"
```
xuwhite:
```suggestion
msgstr "Lifetime (minutes)"
```

resources/lang/en_US/default.po
```diff
@@ -778,6 +781,15 @@ msgid "angeltypes.hide_on_shift_view.info"
...
+
+msgid "angeltypes.qr.time.info"
+msgstr "The link will be valid for the whole time span and can't be revoked beforehand!"
```
xuwhite:
```suggestion
msgstr "The link will be valid for the whole time span and can't be revoked!"
```
"beforehand" without an explicit reference confuses me

resources/views/pages/angeltypes/qr.twig
```diff
@@ -0,0 +1,47 @@
...
+        <div class="offset-md-4 col-md-4">
+            <form method="post" class="row row-cols-lg-auto g-3 align-items-center">
+                {{ csrf() }}
+
+                <div class="col-12">
+                    {{ f.input('minutes', __('angeltypes.qr.time'), {
+                        'type': 'number',
+                        'required': true,
+                        'value': f.formData('minutes', 1),
+                        'min': 1,
```
xuwhite, line 46: I would like to also have a max time that could be a config option.

resources/views/pages/angeltypes/qr.twig
```diff
@@ -0,0 +1,47 @@
+{% extends "layouts/app.twig" %}
+{% import 'macros/base.twig' as m %}
+{% import 'macros/form.twig' as f %}
+
+{% block title %}{{ __('angeltypes.qr', [angel_type.name]) }}{% endblock %}
```
xuwhite, line 5: I would prefer a more descriptive title that includes qr

xuwhite (changes requested):
CR: Not complete I didn't get through half of the files.

resources/views/pages/angeltypes/qr.twig
```diff
@@ -0,0 +1,47 @@
...
+
+                {{ f.input('url', __('general.url'), {'value': qr_data, 'readonly': true}) }}
+            </div>
+        {% endif %}
+
+        <div class="offset-md-4 col-md-4">
+            <form method="post" class="row row-cols-lg-auto g-3 align-items-center">
+                {{ csrf() }}
+
+                <div class="col-12">
```
xuwhite: imho it shouldn't be possible to generate more than one code at a time. The input fields should not be visible when the QR code was generated.

src/Controllers/AngelTypesController.php
```diff
@@ -22,4 +53,108 @@ public function about(): Response
...
+
+            $userAngelType->save();
+        }
+
+        return redirect(url('/angeltypes', ['action' => 'view', 'angeltype_id' => $angelType->id]));
+    }
+
+    protected function qrJoinEnabled(): void
+    {
+        if ($this->config->get('app_key') && $this->config->get('join_qr_code', true)) {
```
xuwhite, line 158: I think it should be possible to deactivate this feature. I wouldn't associate app_key with this feature though.

xuwhite: What about the jwt_algorithm here?

xuwhite: could you please add all used config options to the default config

xuwhite:
please add a timer to show how long the QR code is active

xuwhite:
and add a configurable max time with default 2h

xuwhite:
I would like to see a unique identifier of the QR code in each log entry. So it is possible to see in the log which QR code was used.

resources/views/pages/angeltypes/qr.twig
xuwhite: fixed over the log of the QR code identifier

src/Controllers/AngelTypesController.php
MyIgel: already implemented

PR #1569, opened 2025-10-25
replace user state 'force_active' / 'force_food' / 'got_goodie' with 'force_active_by' / 'force_food_by' / 'got_goodie_by'

olebittner:
This PR replaces user state 'force_active' with 'force_active_by' and  'force_food' with  'force_food_by'. And shows the user who applied the state on the user_view:

grafik

- [x] replace ``force_active``
- [x] replace ``force_food``
- [x] replace ``got_goodie``

### DIFF
--- a/db/migrations/2024_08_21_000000_shift_entries_freeloaded_to_freeload_user_id.php
+++ b/db/migrations/2024_08_21_000000_shift_entries_freeloaded_to_freeload_user_id.php
@@ -13,6 +13,7 @@ class ShiftEntriesFreeloadedToFreeloadUserId extends Migration
 
     private function boolToUserId(): void
     {
+        $user_id = $this->schema->getConnection()->table('users')->orderBy('id')->first()?->id;
         $connection = $this->schema->getConnection();
         $connection
             ->table('shift_entries')
@@ -24,7 +25,7 @@ class ShiftEntriesFreeloadedToFreeloadUserId extends Migration
             ->table('shift_entries')
             ->where('freeloaded', true)
             ->update([
-                'freeloaded_by' => $connection->raw('user_id'),
+                'freeloaded_by' => $user_id,
             ]);
     }
 


olebittner:
@xuwhite could you please check, if this looks good to you? Then I'll do the same for ``force_food``.

We could also do the same for ``got_goodie``.

db/factories/User/StateFactory.php
```diff
@@ -17,13 +17,14 @@ class StateFactory extends Factory
     public function definition(): array
     {
         $arrival = $this->faker->optional()->dateTimeThisMonth();
+        $force_active_by = $this->faker->optional(.01)->passthrough(User::factory());
```
xuwhite:
```suggestion
        $force_active_by = $this->faker->optional(.1)->passthrough(User::factory());
```
possibility a mocked user is force active is higher than freeloaded 🙃

tests/Unit/Models/User/UserStateTest.php
```diff
@@ -38,4 +39,52 @@ public function testScopeWhereArrived(): void
...
     }
+    /**
+     * @covers \Engelsystem\Models\User\State::getForceActiveAttribute
+     */
+    public function testGetForceActiveAttribute(): void
+    {
+        $state = new State();
+        $this->assertFalse($state->force_active);
+
+        $state->force_active_by = 1;
```
xuwhite:
```suggestion
        $user = User::factory()->create();
        $state->force_active_by = $user->id;
```

tests/Unit/Models/User/UserStateTest.php
```diff
@@ -38,4 +39,52 @@ public function testScopeWhereArrived(): void
...
+     */
+    public function testScopeWhereForceActive(): void
+    {
+        $state = State::factory()->create([
+            'force_active_by' => null,
+        ]);
+        $this->assertCount(0, State::whereForceActive(true)->get());
+        $this->assertCount(1, State::whereForceActive(false)->get());
+
+        $state->force_active_by = 1;
```
xuwhite:
```suggestion
        $user = User::factory()->create();
        $state->force_active_by = $user->id;
```

tests/Unit/Models/User/UserStateTest.php
xuwhite: Please create an actual user and use the user id

tests/Unit/Models/User/UserTest.php
```diff
@@ -76,7 +76,7 @@ public function hasOneRelationsProvider(): array
                 State::class,
                 'state',
                 [
-                    'force_active' => true,
+                    'force_active_by' => 1,
```
xuwhite: same here

tests/Unit/Controllers/Metrics/StatsTest.php
```diff
@@ -597,7 +597,7 @@ protected function addUsers(): void
             ['language' => 'lo_RM']
         );
         $this->addUser(
-            ['arrival_date' => Carbon::now(), 'got_voucher' => 9, 'force_active' => true, 'user_info' => 'Info'],
+            ['arrival_date' => Carbon::now(), 'got_voucher' => 9, 'force_active_by' => 1, 'user_info' => 'Info'],
```
xuwhite: and here

tests/Unit/Controllers/Metrics/StatsTest.php
```diff
@@ -614,7 +614,7 @@ protected function addUsers(): void
                 'ifsg_confirmed' => true,
             ]
         );
-        $this->addUser(['arrival_date' => Carbon::now(), 'active' => 1, 'got_goodie' => true, 'force_active' => true]);
+        $this->addUser(['arrival_date' => Carbon::now(), 'active' => 1, 'got_goodie' => true, 'force_active_by' => 1]);
```
xuwhite: and here

includes/view/User_view.php
```diff
@@ -831,7 +831,7 @@ function User_view_state($admin_user_privilege, $freeloader, $user_source)
                 . '</span>';
 
             if ($user_source->state->force_active && config('enable_force_active')) {
-                $state[] = '<span class="text-success">' . __('user.force_active') . '</span>';
+                $state[] = '<span class="text-success">' . sprintf(__('user.force_active.by'), $user_source->state->forceActiveBy->name) . '</span>';
```
xuwhite:
```suggestion
                $state[] = '<span class="text-success">' . __('user.force_active.by', [User_Nick_render(User::find($user_source->state->forceActiveBy))]) . '</span>';
```
the translation function has a second argument to add inserts, I think it looks nicer if we render a user link

xuwhite:
> @xuwhite could you please check, if this looks good to you? Then I'll do the same for `force_food`.
> 
> We could also do the same for `got_goodie`.

@myigel what do you think about that idea?

tests/Unit/Models/User/UserTest.php
olebittner: I cannot access the database in the dataProvider. I'm now overriding the field in the test-case, but I think this makes it worse 😓  maybe setting ``1`` is the better way here?

see: https://github.com/engelsystem/engelsystem/pull/1569/commits/33b12e40d8271947d9883f02fa3e274d13ddf114#diff-aa294697986a7cfc5d2a6502c6111406d31efbb0931dac2913e9a0a958cfaa15R156

tests/Unit/Controllers/Metrics/StatsTest.php
olebittner: I'm now reusing the first added user to set ``force_active_by`` as adding new users screws with the other tests in this class.

see: https://github.com/engelsystem/engelsystem/pull/1569/commits/33b12e40d8271947d9883f02fa3e274d13ddf114#diff-61c393c43f80547eec6b625b574094724b996a5ecb0f8712acb0a59b6ced7837R582

olebittner:
grafik

>  I would personally also render the name on the user edit page, could you add that in?

Sure:
grafik

olebittner:
@xuwhite please check again 😄 

I just added commits for now, I will squash in the end.

MyIgel:
> > We could also do the same for `got_goodie`.
> 
> @MyIgel what do you think about that idea?

Sure, why not!

db/migrations/2024_08_21_000000_shift_entries_freeloaded_to_freeload_user_id.php
```diff
@@ -13,19 +13,26 @@ class ShiftEntriesFreeloadedToFreeloadUserId extends Migration
 
     private function boolToUserId(): void
     {
+        $user = $this->schema->getConnection()->table('users')->orderBy('id')->first();
```
MyIgel:
```suggestion
        $userId = $this->schema->getConnection()->table('users')->orderBy('id')->first()?->id;
```
is the fastest way

db/migrations/2024_08_21_000000_shift_entries_freeloaded_to_freeload_user_id.php
```diff
@@ -13,19 +13,26 @@ class ShiftEntriesFreeloadedToFreeloadUserId extends Migration
...
             ->update([
                 'freeloaded_by' => null,
             ]);
-        $connection
-            ->table('shift_entries')
-            ->where('freeloaded', true)
-            ->update([
-                'freeloaded_by' => $connection->raw('user_id'),
-            ]);
+        if ($user_id !== null) {
```
MyIgel:
```suggestion
```

When we have shift entries we have users so it should work every time

db/migrations/2025_12_20_000000_users_state_force_active_force_food_got_goodie_to_by_user_ref.php
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Run the migration
+     */
+    public function up(): void
+    {
+        $user = $this->schema->getConnection()->table('users')->orderBy('id')->first();
+        $user_id = null;
+        if ($user !== null) {
+            $user_id = (int) $user->id;
```
MyIgel: Imho if a user somehow creates a system with no users and runs this migration its ok to fail hard -> no explicit check needed (and not used in other migrations too)

includes/pages/admin_user.php
```diff
@@ -360,7 +369,13 @@ function admin_user()
                 $user_source->contact->save();
 
                 if ($goodie_enabled && $user_goodie_edit) {
-                    $user_source->state->got_goodie = $request->postData('goodie');
+                    if ($user_source->state->got_goodie != $request->postData('goodie')) {
+                        if ($request->postData('goodie')) {
```
MyIgel:
```suggestion
                        $request->postData('goodie') ? $user->id : null;
```

& others in other files too

src/Models/User/State.php
```diff
@@ -7,25 +7,32 @@
...
 
 /**
  * @property-read bool   $arrived
  * @property Carbon|null $arrival_date
  * @property string|null $user_info
  * @property bool        $active
- * @property bool        $force_active
- * @property bool        $force_food
- * @property bool        $got_goodie
+ * @property-read bool   $force_active
```
MyIgel, line 18: Explicit boolean fields are nice to have but on the other side any user ID is truthy enough so could be directly used

src/Models/User/State.php
```diff
@@ -7,25 +7,32 @@
...
+ * @property-read bool   $got_goodie
+ * @property int|null    $got_goodie_by
  * @property int         $got_voucher
  *
  * @method static QueryBuilder|State[] whereArrived($value)
  * @method static QueryBuilder|State[] whereArrivalDate($value)
  * @method static QueryBuilder|State[] whereUserInfo($value)
  * @method static QueryBuilder|State[] whereActive($value)
  * @method static QueryBuilder|State[] whereForceActive($value)
+ * @method static QueryBuilder|State[] whereForceActiveBy($value)
```
MyIgel, line 31: (same here and following)

db/migrations/2025_12_20_000000_users_state_force_active_force_food_got_goodie_to_by_user_ref.php
```diff
@@ -0,0 +1,94 @@
...
+namespace Engelsystem\Migrations;
+
+use Engelsystem\Database\Migration\Migration;
+use Illuminate\Database\Schema\Blueprint;
+
+class UsersStateForceActiveForceFoodGotGoodieToByUserRef extends Migration
+{
+    use Reference;
+
+    private const COL_FORCE_ACTIVE = 'force_active';
```
MyIgel: Explicit constants seem a bit "too much" here as they are only used in two places each?

src/Models/User/State.php
olebittner: hmm, unit tests using the user_id
`this->assertTrue($state->force_active_by);`
fail with
`Failed asserting that 1 is true.`

are unit tests more strict regarding truthiness?

src/Models/User/State.php
olebittner: while `whereForceActiveBy(null)` works to select non force active users.
only having `whereForceActiveBy($value)` whould also require a `whereForceActiveByIsNotNull()` or similar to select force active users.

PR #1573, opened 2025-10-30
trigger password reset mail from admin_user.php

olebittner:
This PR allows to trigger the mail based password reset flow from the admin_user.php edit page:
grafik

I moved the existing logic used in the "I forgot my password" flow into the newly created ``PasswordReset`` helper.

- Should this button honor the ``email_human`` setting and only be visible when the user opted-in?
  - currently it is always visible.
- Should there be a confirmation dialog to prevent sending accidental mails?
  - currently this is a one-click action

### DIFF
--- a/includes/pages/admin_user.php
+++ b/includes/pages/admin_user.php
@@ -197,9 +197,23 @@ function admin_user()
             . '</td></tr>' . "\n";
 
         $html .= '</table>' . "\n" . '<br>' . "\n";
-        $html .= '<button type="submit" class="btn btn-primary">'
-            . icon('save') . __('form.save') . '</button>' . "\n";
         $html .= '</form>';
+
+        $html .= '<div class="row justify-content-start">';
+        $html .= '<div class="col-sm-auto">';
+        $html .= '<button type="submit" form="change_password" class="btn btn-primary">'
+            . icon('save') . __('form.save') . '</button>' . "\n";
+        $html .= '</div>';
+        $html .= '<form class="col-sm-auto" action="'
+            . url('/users/' . $user_id . '/password/reset')
+            . '" method="post">' . "\n";
+        $html .= form_csrf();
+        $html .= '<button type="submit" class="btn btn-warning">'
+            . icon('envelope-at')
+            . __('password.recovery.send')
+            . '</button>';
+        $html .= '</form>';
+        $html .= '</div>';
 
         $html .= '<hr>';
 
--- a/resources/lang/en_US/default.po
+++ b/resources/lang/en_US/default.po
@@ -1093,11 +1093,17 @@ msgstr "Angel"
 msgid "title.date"
 msgstr "Date"
 
+msgid "password.recovery.send"
+msgstr "Send password recovery e-mail"
+
 msgid "password.reset.confirm"
 msgstr "Confirm password"
 
 msgid "password.recovery.success"
 msgstr "We sent you an e-mail containing your password recovery link."
+
+msgid "password.recovery.success.admin"
+msgstr "Password recovery e-mail sent."
 
 msgid "password.reset.title"
 msgstr "Password recovery"
--- a/src/Controllers/PasswordResetController.php
+++ b/src/Controllers/PasswordResetController.php
@@ -11,6 +11,7 @@ use Engelsystem\Mail\EngelsystemMailer;
 use Engelsystem\Models\User\PasswordReset;
 use Engelsystem\Models\User\User;
 use Psr\Log\LoggerInterface;
+use Random\RandomException;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 
 class PasswordResetController extends BaseController
@@ -23,6 +24,7 @@ class PasswordResetController extends BaseController
         'postReset'         => 'login',
         'resetPassword'     => 'login',
         'postResetPassword' => 'login',
+        'postResetByUserId' => 'admin_user',
     ];
 
     public function __construct(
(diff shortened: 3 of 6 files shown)


src/Helpers/Password.php
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @param User $user The user to trigger the password reset for.
+     * @throws RandomException
+     */
+    public static function triggerPasswordReset(User $user): void
+    {
+        /** @var EngelsystemMailer $mailer */
+        $mailer = app(EngelsystemMailer::class);
+        /** @var LoggerInterface $logger */
+        $logger = app(LoggerInterface::class);
```
olebittner: @MyIgel not sure if this is the best method to get ahold of the mailer and logger. If there is a better way, please let me know.

MyIgel: The "better" way would be to set it in the constructor which is not used here as it is a static helper so that's fine :+1:

src/Helpers/Password.php
```diff
@@ -0,0 +1,44 @@
...
+
+        $reset = (new \Engelsystem\Models\User\PasswordReset())->findOrNew($user->id);
```
MyIgel: This should be imported

includes/pages/admin_user.php
```diff
@@ -188,8 +189,14 @@ function admin_user()
             . '</td></tr>' . "\n";
 
         $html .= '</table>' . "\n" . '<br>' . "\n";
-        $html .= '<button type="submit" class="btn btn-primary">'
-            . icon('save') . __('form.save') . '</button>' . "\n";
+        $html .= buttons([
+            '<button type="submit" class="btn btn-primary">' . icon('save') . __('form.save') . '</button>',
+            button(
+                url('/admin-user', ['action' => 'reset_pw', 'id' => $user_id]),
+                icon('envelope-at') . __('password.recovery.send'),
```
MyIgel: Technically this is a legacy text and thus not dotted (and i am not sure if the dotted variation here fits that well, eventually some form of settings.password_recovery.send maybe would fit more? (or just ignore it and use the English text to "fit" in the page ;) )

includes/pages/admin_user.php
```diff
@@ -402,6 +409,12 @@ function admin_user()
                     );
                 }
                 break;
+
+            case 'reset_pw':
+                $user_source = User::find($user_id);
+                Password::triggerPasswordReset($user_source);
+                $html .= success(__('Password reset triggered.'), true);
```
MyIgel: This is missing a German translation

resources/lang/en_US/default.po
```diff
@@ -1034,6 +1034,9 @@ msgstr "Angel"
 msgid "title.date"
 msgstr "Date"
 
+msgid "password.recovery.send"
+msgstr "Send Password Recovery E-mail"
```
MyIgel:
```suggestion
msgstr "Send password recovery e-mail"
```

would fit better afaik?

tests/Unit/Helpers/PasswordResetTest.php
```diff
@@ -0,0 +1,42 @@
...
+use Engelsystem\Helpers\Password;
+use Engelsystem\Mail\EngelsystemMailer;
+use Engelsystem\Models\User\User;
+use Engelsystem\Test\Unit\HasDatabase;
+use Engelsystem\Test\Unit\TestCase;
+use PHPUnit\Framework\MockObject\MockObject;
+use Psr\Log\LoggerInterface;
+use Psr\Log\Test\TestLogger;
+
+class PasswordResetTest extends TestCase
```
MyIgel:
```suggestion
class PasswordTest extends TestCase
```

src/Controllers/PasswordResetController.php
```diff
@@ -47,22 +48,7 @@ public function postReset(Request $request): Response
...
-                ['user' => $user->toJson()]
-            );
-
-            $this->mail->sendViewTranslated(
-                $user,
-                'Password recovery',
-                'emails/password-reset',
-                ['username' => $user->displayName, 'reset' => $reset]
-            );
+            PasswordResetHelper::triggerPasswordReset($user);
```
MyIgel: I would have left it in this Controller and just moved it  to a public method which can be called by the settings page; The Helpers are "not a great idea" design wise / by themselves as its a place where "all that random junk" accumulates (generally it would be great to use them only for stuff that actually has no other place to live).

tldr: I am not completely happy but would let it go through?

MyIgel: @xuwhite implemented an alternative approach in https://github.com/engelsystem/engelsystem/commit/e257c3d715c37369223b50161e4238370fe3e199 by integrating it to the PasswordResetController -> feel free to integrate it / refactor that a bit (eventually switching the method calls around could be even cleaner) :)

MyIgel (changes requested):

olebittner:
Hi @MyIgel 
finally had the time to put some more work into this. It is now much closer to @xuwhite's implementation. Please take another look. Testcases for the new functions are still missing, I will add them once the feature is finalized.

src/Controllers/PasswordResetController.php
```diff
@@ -23,6 +24,7 @@ class PasswordResetController extends BaseController
         'postReset'         => 'login',
         'resetPassword'     => 'login',
         'postResetPassword' => 'login',
+        'postResetByUserId'     => 'admin_user',
```
MyIgel:
```suggestion
        'postResetByUserId' => 'admin_user',
```

src/Controllers/PasswordResetController.php
```diff
@@ -119,4 +106,40 @@ protected function requireToken(Request $request): PasswordReset
...
+        $reset->save();
+
+        $this->log->info(
+            sprintf('Password recovery for %s (%u)', $user->name, $user->id),
+            ['user' => $user->toJson()]
+        );
+
+        $this->mail->sendViewTranslated(
+            $user,
+            'Password recovery',
```
MyIgel:
```suggestion
            'password.reset.title',
```

src/Controllers/PasswordResetController.php
```diff
@@ -119,4 +106,40 @@ protected function requireToken(Request $request): PasswordReset
...
+
+    /**
+     * Trigger a password reset for the given user. This will email the user!
+     *
+     * @param User $user The user to trigger the password reset for.
+     * @throws RandomException
+     */
+    private function triggerPasswordReset(User $user): void
+    {
+
```
MyIgel:
```suggestion
```

MyIgel (commented):
Looks good to me so far; just some smaller formatting nots:

PR #1574, opened 2025-10-31
Limit the height of the angel list in the myshifts view

### LINKED ISSUE #1487: My shifts: Hide assigned angels
In the "my shifts" view, all assigned angels to a shift are displayed. E.g., this causes in a confusing page if you are assigned to the "Angel Meeting (incl Evacuation Training)". It would be nicer if this list is limited to one line or if you can disable or enable it.

Enterprize1:
Resolves #1487 by limiting the height of the shift-info column to 4 lines (6 on smaller screens). When hovering the cell expands just as in the shift calendar view.

image

Note on the implementation: While there is a vendor prefix of `-webkit` it should work in all the browsers: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/line-clamp#browser_compatibility

### DIFF
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -346,6 +346,8 @@ function User_view_myshift(Shift $shift, $user_source, $its_me, $supporter)
         $shift_info .= User_view_shiftentries($needed_angel_type);
     }
 
+    $shift_info = div('table-myshifts-shift-info-limit-height', $shift_info);
+
     $night_shift = '';
     if ($shift->isNightShift() && $goodie_enabled) {
         $night_shift = render_night_shift_hint($nightShiftsConfig);
--- a/resources/assets/themes/base.scss
+++ b/resources/assets/themes/base.scss
@@ -307,6 +307,17 @@ table.table-sticky-header thead {
   text-align: right;
 }
 
+.column_shift_info:not(:hover) .table-myshifts-shift-info-limit-height {
+  display: -webkit-box;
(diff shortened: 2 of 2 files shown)


MyIgel:
That works great! Just a small detail: afaik the `display: box` should be `display: flex` nowadays and aren't the webkit prefixes automatically added on build if needed so could be omitted? @marudor any thoughts here / how to implement this?

Enterprize1:
> That works great! Just a small detail: afaik the `display: box` should be `display: flex` nowadays and aren't the webkit prefixes automatically added on build if needed so could be omitted? @marudor any thoughts here / how to implement this?

I completely agree that it looks pretty weird :) Each of those properties should not be used on their own, they only make sense in combination as the browsers all agreed to support this while the [CSS Overflow Module Level 4](https://drafts.csswg.org/css-overflow-4/#line-clamp) is still being worked on. I think the [note at the top of the mdn page](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/line-clamp) describes it best.

The only other way I could see is to limit the height to four times the line-height in the container or something like that and to show that there is more content implement an ellipsis like here: https://codepen.io/2oldladiescode/pen/OVgQvx but that seems even more hacky :)

xuwhite:
It looks ugly. Thanks for the good explanation. Until the CSS module is supported by all browsers, this seems like a working solution. "If it works, it can't be that stupid."

PR #1575, opened 2025-11-02
Show shifts on shift type page via legacy renderer

MyIgel:
* Added legacy map for shifts renderer and dateWithEventDay filter for twig templates
* Added needsUsers scope to shift for filtering shifts that have needed angel types
* Added shifts view for shift type
* Improved shifts rendering performance by reducing SQL queries for own shifts

### DIFF
--- a/includes/model/Shifts_model.php
+++ b/includes/model/Shifts_model.php
@@ -614,6 +614,12 @@ function Shift_signup_allowed(
  */
 function Shifts_by_user($userId, $include_freeloaded_comments = false)
 {
+    # Cache static content per request
+    static $cached;
+    if (!empty($cached[$userId][$include_freeloaded_comments])) {
+        return $cached[$userId][$include_freeloaded_comments];
+    }
+
     $shiftsData = Db::select(
         '
         SELECT
@@ -647,6 +653,7 @@ function Shifts_by_user($userId, $include_freeloaded_comments = false)
         $shifts[] = (new Shift())->forceFill($data);
     }
     $shifts->load(['shiftType', 'location']);
+    $cached[$userId][$include_freeloaded_comments] = $shifts;
 
     return $shifts;
 }
--- a/resources/views/admin/shifttypes/view.twig
+++ b/resources/views/admin/shifttypes/view.twig
@@ -5,30 +5,93 @@
 {% block title %}{{ shifttype.name }}{% endblock %}
 
 {% block row_content %}
-    {% if shifttype.signup_advance_hours %}
-        <div class="col-md-12">
-            {{ __('shifttype.edit.signup_advance_hours') }} {{ f.info(__('shifttype.edit.signup_advance_hours.info')) }}: {{ shifttype.signup_advance_hours }}
-        </div>
-    {% endif %}
-
     <div class="col-md-12">
-        <h3>{{ __('general.description') }}</h3>{{ shifttype.description|md }}
-    </div>
+        <ul class="nav nav-tabs mb-3" role="tablist">
+            <li class="nav-item" role="presentation">
+                <button class="nav-link{% if not shifts_active %} active{% endif %}" id="home-tab" data-bs-toggle="tab"
+                        data-bs-target="#view-info" type="button" role="tab" aria-controls="view-info"
+                        aria-selected="{% if not shifts_active %}true{% else %}false{% endif %}"
+                >
+                    {{ __('general.info') }}
+                </button>
+            </li>
+            <li class="nav-item" role="presentation">
+                <button class="nav-link{% if shifts_active %} active{% endif %}" id="profile-tab" data-bs-toggle="tab"
+                        data-bs-target="#view-shifts" type="button" role="tab" aria-controls="view-shifts"
+                        aria-selected="{% if not shifts_active %}false{% else %}true{% endif %}"
+                >
+                    {{ __('general.shifts') }}
+                </button>
+            </li>
+        </ul>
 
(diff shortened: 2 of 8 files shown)


db/factories/Shifts/ScheduleShiftFactory.php
```diff
@@ -0,0 +1,25 @@
...
+{
+    /** @var string */
+    protected $model = ScheduleShift::class; // phpcs:ignore
+
+    public function definition(): array
+    {
+        return [
+            'shift_id' =>    Shift::factory(),
+            'schedule_id' => Schedule::factory(),
+            'guid' =>        $this->faker->uuid(),
```
rehleinBo, line 22: A word like `faker` doesn't sound right in a non test file but it is, isn't it?

includes/model/Shifts_model.php
```diff
@@ -614,6 +614,12 @@ function Shift_signup_allowed(
  */
 function Shifts_by_user($userId, $include_freeloaded_comments = false)
 {
+    # Cache static content per request
+    static $cached;
+    if (!empty($cached[$userId][$include_freeloaded_comments])) {
```
rehleinBo, line 619: Would it result in an exception/seg fault if `cached[$userId]` is null?

includes/model/Shifts_model.php
```diff
@@ -647,6 +653,7 @@ function Shifts_by_user($userId, $include_freeloaded_comments = false)
         $shifts[] = (new Shift())->forceFill($data);
     }
     $shifts->load(['shiftType', 'location']);
+    $cached[$userId][$include_freeloaded_comments] = $shifts;
```
rehleinBo, line 656: same here

resources/views/admin/shifttypes/view.twig
```diff
@@ -5,30 +5,93 @@
...
+                    {% if shifts|length %}
+                        <button class="btn btn-secondary dropdown-toggle" type="button"
+                                data-bs-toggle="dropdown" aria-expanded="false"
+                        >
+                            {{ selected_day|dateWithEventDay }}
+                        </button>
                     {% endif %}
-                {% endfor %}
-            </ul>
+                    <ul class="dropdown-menu">
```
rehleinBo, line 78: A classic `select` is not an option?

Disclaimer: It's just a question from a point not knowing the design and component system

src/Models/Shifts/Shift.php
```diff
@@ -133,6 +136,57 @@ public function updatedBy(): BelongsTo
...
+                                            ->where('schedules.needed_from_shift_type', true)
+                                            ->whereColumn('needed_angel_types.shift_type_id', 's2.shift_type_id');
+                                    });
+                            });
+                    })
+                    ->selectRaw('COUNT(*) > 0'),
+            ]);
+
+        if ($query->getConnection()->getQueryGrammar() instanceof SQLiteGrammar) {
+            // SQLite does not support HAVING for non-aggregate queries
```
rehleinBo, line 180: I'm impressed MySQL/Maria does....

Most databases I know do support `having` only as filter on grouped results

src/Helpers/ShiftsRenderer.php
```diff
@@ -0,0 +1,83 @@
...
+     * @param Collection|Shift[] $shifts
+     * @param array[] $neededAngelTypes
+     * @param ShiftEntry[][] $shiftEntries
+     * @codeCoverageIgnore
+     */
+    protected function renderShiftCalendar(
+        array | Collection $shifts,
+        array $neededAngelTypes,
+        array $shiftEntries
+    ): string {
```
rehleinBo, line 65: In other files the opening parenthesis is in a new line.

db/factories/Shifts/ScheduleShiftFactory.php
MyIgel: `db/factories` is used only for tests so its fine here :D

resources/views/admin/shifttypes/view.twig
MyIgel: A select does not allow "directly" linking the different days which the "dropdown select with links" does here afaik?

src/Models/Shifts/Shift.php
MyIgel: If you have a good idea how i can prevent using it at all / designing a query that works on all i am happy to hear it :D
After tinkering and shoving queries around thats the best i could manage after banging my head at the table enough or something ;)
(it is indeed not optimal but seem to work so far and is better than combining three queries into one using union as it is done everywhere else)

The biggest problem i had here is to create a query that
* Works on SELECT
* and when using COUNT
* Does not change the values that are SELECTed for (as would a join do)

And possibly more that i can't remember currently :upside_down_face:

src/Helpers/ShiftsRenderer.php
MyIgel: not when the closing are on their own page ;)

includes/model/Shifts_model.php
MyIgel: > No warning is generated if the variable does not exist. [...] This also applies to nested structures, such as a multidimensional array or chained properties. 
> https://www.php.net/manual/en/function.empty.php#refsect1-function.empty-parameters

So no, there should be no option to have that result in an error.

includes/model/Shifts_model.php
MyIgel: (see above)

olebittner (approved):

PR #1576, opened 2025-11-02
Add 39c3 theme

msquare:
Proposal for 39c3 theme.
grafik


MyIgel:
I am not sure if we should keep the white & green menu, it seems a bit much :D
(and eventually wait a bit until we can get inspired by the hub? see for example pages like https://events.ccc.de/congress/2025/hub/en/me/)

MyIgel:
We moved multiple colors around for a more coherent look and ended up with that configuration.
It is not perfect by any means but is imho at least useable even if you see it all day ^^

Screenshot_20251227_033727

(Improvement ideas welcome!)

PR #1578, opened 2025-11-02
Smaller patches & fixes

MyIgel:
* Add PDO to container to be easier accessible
* Automatically confirm a user when added as a supporter
* Prevent ShiCos from removing the Angel group (only allow it for Developers)
* Log OAuth info on registration and user on questions

### DIFF
--- a/src/Controllers/Admin/QuestionsController.php
+++ b/src/Controllers/Admin/QuestionsController.php
@@ -57,7 +57,15 @@ class QuestionsController extends BaseController
         $question = $this->question->findOrFail($data['id']);
         $question->delete();
 
-        $this->log->info('Deleted question {question} ({id})', ['question' => $question->text, 'id' => $question->id]);
+        $this->log->info(
+            'Deleted question "{text}" ({id}) by {user} ({user_id})',
+            [
+                'text' => $question->text,
+                'id' => $question->id,
+                'user' => $question->user->name,
+                'user_id' => $question->user->id,
+            ]
+        );
         $this->addNotification('question.delete.success');
 
         return $this->redirect->to('/admin/questions');
@@ -90,8 +98,13 @@ class QuestionsController extends BaseController
             $question->delete();
 
             $this->log->info(
-                'Deleted question {question} ({id})',
-                ['question' => $question->text, 'id' => $question->id]
+                'Deleted question "{text}" ({id}) by {user} ({user_id})',
+                [
+                    'text' => $question->text,
+                    'id' => $question->id,
+                    'user' => $question->user->name,
+                    'user_id' => $question->user->id,
+                ]
             );
 
             $this->addNotification('question.delete.success');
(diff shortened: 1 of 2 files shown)


xuwhite:
What the hack is PDO?

includes/controller/user_angeltypes_controller.php
```diff
@@ -286,6 +286,14 @@ function user_angeltype_update_controller(): array
 
     if ($request->hasPostData('submit')) {
         $user_angeltype->supporter = $supporter;
+        if ($supporter && !$user_angeltype->confirm_user_id) {
```
xuwhite, line 289: Should we also add a migration to fix the existing ones?
I don't really understand what you are doing here.

includes/pages/admin_user.php
```diff
@@ -269,6 +270,22 @@ function admin_user()
...
+                        if (
+                            !in_array($defaultGroup, $groupsRequest)
+                            && $angel->groups->where('id', $defaultGroup)->count()
+                        ) {
+                            if (!auth()->can('admin_groups') && !config('default_group_removable')) {
+                                $groupsRequest[] = $defaultGroup;
+                                $html .= warning(__('You cannot remove the default group.'), true);
+                            } else {
+                                $html .= warning(
+                                    __('You removed the default group, this has unintended side effects!'),
```
xuwhite, line 283: I would prefer a confirmation popup
For this solution I would at least change them to errors

includes/pages/admin_user.php
```diff
@@ -269,6 +270,22 @@ function admin_user()
...
                         }
 
+                        $defaultGroup = auth()->getDefaultRole();
+                        if (
+                            !in_array($defaultGroup, $groupsRequest)
+                            && $angel->groups->where('id', $defaultGroup)->count()
+                        ) {
+                            if (!auth()->can('admin_groups') && !config('default_group_removable')) {
+                                $groupsRequest[] = $defaultGroup;
+                                $html .= warning(__('You cannot remove the default group.'), true);
```
xuwhite: I don't think this should be a valid save. 
This should not exist.
Screenshot 2025-11-18 at 19 45 50

includes/controller/user_angeltypes_controller.php
MyIgel: It checks if "supporter" is checked and, when the user is not confirmed, confirms it directly -> the user should never be "unconfirmed" at the end. A migration would be possible but imho not needed to jist capture the edge case of an edge case ;)

includes/pages/admin_user.php
MyIgel: Error on a successful change seems a bit odd; Changing it to an error for ShiCos is a good idea.

includes/pages/admin_user.php
MyIgel: Fixed

MyIgel:
PDO is PHPs database connection wrapper; its not existence in the container is not needed (yet) but provided here for completeness

PR #1580, opened 2025-11-08
tests: Make GoodieTests run successful with disabled night-shifts in config.php

pzii:
This PR makes GoodieTests run successful even in situations when night-shifts are disabled in the main config.php.
The reason for failing tests is/was the non-correct usage of the config() function.


MyIgel:
Good catch, thats a dumb bug to make but happens ^^

PR #1581, opened 2025-11-08
oauth: Implement a config flag to disable disconnecting from provider

pzii:
This PR implements a per-provider option to restrict disconnecting user accounts from the provider.

In our setup we have a single SSO provider that is to handle all authentication; allowing users to disconnect individual accounts from SSO does not make sense in our use case.

### DIFF
--- a/config/config.default.php
+++ b/config/config.default.php
@@ -138,6 +138,8 @@ return [
             'enable_password' => false,
             // Allow registration even if disabled in config (optional)
             'allow_registration' => null,
+            // Allow disconnecting user accounts from the oauth provider (optional)
+            'allow_user_disconnect' => true,
             // Auto join teams
             // Info groups field (optional)
             'groups' => 'groups',
--- a/resources/views/pages/settings/oauth.twig
+++ b/resources/views/pages/settings/oauth.twig
@@ -37,7 +37,7 @@
 
                         {{ f.submit(__('form.connect'), {'size' : 'sm', 'icon_left': 'box-arrow-in-right'}) }}
                     </form>
-                {% else %}
+                {% elseif config.allow_user_disconnect is not defined or config.allow_user_disconnect %}
                     <form method="post" action="{{ url('/oauth/' ~ name ~ '/disconnect') }}">
                         {{ csrf() }}
 
--- a/src/Controllers/OAuthController.php
+++ b/src/Controllers/OAuthController.php
@@ -194,6 +194,12 @@ class OAuthController extends BaseController
     {
         $providerName = $request->getAttribute('provider');
 
+        $this->requireProvider($providerName);
+
+        if (!($this->config->get('oauth')[$providerName]['allow_user_disconnect'] ?? true)) {
+            throw new HttpNotFound();
+        }
+
         $this->oauth
             ->whereUserId($this->auth->user()->id)
             ->where('provider', $providerName)
--- a/tests/Unit/Controllers/OAuthControllerTest.php
+++ b/tests/Unit/Controllers/OAuthControllerTest.php
@@ -574,20 +574,69 @@ class OAuthControllerTest extends TestCase
     /**
      * @covers \Engelsystem\Controllers\OAuthController::disconnect
      */
-    public function testDisconnect(): void
+    public function testDisconnectIfAllowIsTrue(): void
     {
+        $oauthConfig = $this->config->get('oauth');
+        $oauthConfig['testprovider']['allow_user_disconnect'] = true;
+
+        $this->runDisconnectTest($oauthConfig, true);
+    }
+
+    /**
+     * @covers \Engelsystem\Controllers\OAuthController::disconnect
+     */
+    public function testDisconnectIfAllowIsNull(): void
+    {
+        $oauthConfig = $this->config->get('oauth');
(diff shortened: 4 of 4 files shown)


resources/views/pages/settings/oauth.twig
```diff
@@ -37,12 +37,14 @@
 
                         {{ f.submit(__('form.connect'), {'size' : 'sm', 'icon_left': 'box-arrow-in-right'}) }}
                     </form>
-                {% else %}
+                {% elseif config.allow_user_disconnect %}
```
MyIgel:
```
RuntimeError
Key "allow_user_disconnect" for sequence/mapping with keys "name, client_id, client_secret, url_auth, url_token, url_info, nested_info, id, username, email, hidden" does not exist.
```

```suggestion
                {% elseif config.allow_user_disconnect is defined and config.allow_user_disconnect %}
```
(although should be inverted for "default allowed", see above)

resources/views/pages/settings/oauth.twig
```diff
@@ -37,12 +37,14 @@
...
                     <form method="post" action="{{ url('/oauth/' ~ name ~ '/disconnect') }}">
                         {{ csrf() }}
 
                         {{ f.submit(__('form.disconnect'), {'btn_type': 'danger', 'size' : 'sm', 'icon_left': 'box-arrow-right'}) }}
                     </form>
+                {% else %}
+                    {{ f.button(__('form.disconnect.disabled'), {'btn_type': 'danger', 'size' : 'sm', 'icon_left': 'box-arrow-right', attr: {'disabled': 'disabled'}}) }}
```
MyIgel: I would either not show it at all or just display the text "Disabled" instead. (Definitely preferring the first option)

config/config.default.php
```diff
@@ -138,6 +138,8 @@
             'enable_password' => false,
             // Allow registration even if disabled in config (optional)
             'allow_registration' => null,
+            // Allow disconnecting user accounts from the oauth provider (optional)
+            'allow_user_disconnect' => null,
```
MyIgel: i would prefer explicit `true` here as a default that could be changed if needed

resources/lang/en_US/default.po
```diff
@@ -596,6 +596,9 @@ msgstr "All"
 msgid "form.disconnect"
 msgstr "Disconnect"
 
+msgid "form.disconnect.disabled"
+msgstr "Disconnect not allowed"
```
MyIgel:
```suggestion
msgstr "Disconnecting not allowed"
```

src/Controllers/OAuthController.php
```diff
@@ -194,6 +195,11 @@ public function disconnect(Request $request): Response
     {
         $providerName = $request->getAttribute('provider');
 
+        $this->requireProvider($providerName);
+        if (!($this->config->get('oauth')[$providerName]['allow_user_disconnect'] ?? false)) {
+            throw new HttpForbidden('oauth.provider-disconnect-not-allowed');
```
MyIgel: As this should never be shown to the user (the provider either is allowed or not to disconnect), we could throw a 404 here and spare us an error message + translation that would only be shown in edge cases

MyIgel (changes requested):
Hi, thanks for the PR! I have some smaller changes that would make it imho "easier", esp. allowing disconnecting by default and mostly removing he translations as they are redundant or not needed (just handling edge cases of users disconnecting while the static config changes)

resources/views/pages/settings/oauth.twig
pzii: I changed that snipped following your suggestions to allow for null/undefined `allow_user_disconnect` configs

config/config.default.php
pzii: I changed the default configuration to `true`

resources/lang/en_US/default.po
pzii: I removed that translation entirely

src/Controllers/OAuthController.php
pzii: Following your suggestion, I replaced that `HttpForbidden` to a `HttpNotFound`

resources/views/pages/settings/oauth.twig
pzii: I removed the button - it makes sense to not have a button here :)

pzii:
Thank you for your quick review!

I adopted your changes and extended the unit tests regarding the new config flag a little bit to reflect additional 'pitfalls'.

tests/Unit/Controllers/OAuthControllerTest.php
```diff
@@ -574,20 +574,71 @@ public function testConnect(): void
...
 
         $controller->disconnect($request);
-        $this->assertCount(1, OAuth::all());
-        $this->log->hasInfoThatContains('Disconnected');
+
+        if ($shouldDisconnect) {
+            $this->assertCount(1, OAuth::all());
+            $this->assertTrue($this->log->hasInfoThatContains('Disconnected'));
+        } else {
+            $this->assertCount(2, OAuth::all());
```
MyIgel: This is never called (exception only interpreted in tests runner, not directly)

MyIgel (changes requested):

PR #1589, opened 2025-12-12
Bugfixes for rendering, oauth, config and arrival

### LINKED ISSUE #1591: Event config not beeing saved/displayed again
With current HEAD configuring an event saves some data (Seen in DB) but wont display it again opening it again so it looks as if no event has been configured.

This works fine with v3.6.0

Flo

MyIgel:
* md renderer: Handle non-string-values gracefully
* OAuth: Configured mixed case uniqueness for token (currently the unique key produces errors on `Test` vs. `test`)
* Fix all selects being autofocused on page load
* Fix loading config from database.., closes #1591
* Fix arrival button

### DIFF
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -702,7 +702,7 @@ function User_view(
                             form([
                                 form_hidden('action', 'arrived'),
                                 form_hidden('user', $user_source->id),
-                                form_submit('submit', icon('house') . __('user.arrive'), '', false),
+                                form_submit('send', icon('house') . __('user.arrive'), '', false),
                             ], url('/admin-arrive'), 'float:left') : '',
                         ($admin_user_privilege || $auth->can('voucher.edit')) && config('enable_voucher') ?
                             button(


tests/Unit/Config/ConfigServiceProviderTest.php
```diff
@@ -140,14 +140,14 @@ public function testBoot(): void
...
         $this->setExpects($eloquentBuilder, 'get', [['name', 'value']], $configs);
         $this->setExpects($app, 'get', ['config'], $config, $this->exactly(3));
+        $this->setExpects($app, 'make', [EventConfig::class], $eventConfig, $this->exactly(1));
 
         $serviceProvider = new ConfigServiceProvider($app);
         $serviceProvider->boot();
 
         $serviceProvider = new ConfigServiceProvider($app, $eventConfig);
         $serviceProvider->boot();
         $serviceProvider->boot();
```
xuwhite, line 150: ????????
can you please explain this?

MyIgel: this test tests multiple cases how the service provider can be initialised to write less code with the disadvantage of being more complex and thus the multiple boots but different return values above.

In this case the boot does not just return when it has no EventConfig model set but creates an instance by itself and thus shortening the multiple calls by one edge case.

resources/views/admin/log.twig
```diff
@@ -75,7 +75,13 @@
                             {%- endif %}
 
                             <tr>
-                                <td class="table-{{ td_type }} text-nowrap">{{ entry.created_at.format(__('general.datetime')) }}</td>
+                                <td class="table-{{ td_type }} text-nowrap">
+                                    <span id="message-{{ entry.id }}" class="ref-id"></span>
+                                    {{ entry.created_at.format(__('general.datetime')) }}
+                                    <small class="text-muted">
+                                        <a class="ref-link" href="#message-{{ entry.id }}">{{ m.icon('link') }}</a>
```
xuwhite: When I click on the link of this log message:
Screenshot 2025-12-17 at 16 32 16

My page looks like this:
Screenshot 2025-12-17 at 16 32 07

xuwhite: Also I would prefer a column for the link (with no header title to keep it narrow)

src/Controllers/OAuthController.php
```diff
@@ -140,9 +141,18 @@ public function index(Request $request): Response
...
             ]);
-            $oauth->user()
-                ->associate($user)
-                ->save();
+
+            try {
+                $oauth->user()
+                    ->associate($user)
+                    ->save();
+            } catch (UniqueConstraintViolationException) {
```
xuwhite, line 150: This is missing Test coverage.

MyIgel: "Fixed"

resources/views/admin/log.twig
MyIgel: Moved to  #1593

xuwhite (approved):
LGTM!

PR #1590, opened 2025-12-13
New night shift goodie score calculation

xuwhite:
Night shifts hours get multiplied partially. Only hours inside the night time get night shift multiplier.

### DIFF
--- a/src/Helpers/Goodie.php
+++ b/src/Helpers/Goodie.php
@@ -32,11 +32,7 @@ class Goodie
         ));
     }
 
-    /**
-     * Returns the goodie score (number of hours counted for goodie score)
-     * Includes only ended shifts
-     */
-    public static function userScore(User $user): float
+    public static function shiftScoreQuery(User $user): mixed
     {
         /** @var Database $db */
         $db = app(Database::class);
@@ -46,7 +42,7 @@ class Goodie
         $shiftScoreQuery = /** @lang MySQL */
             'SELECT goodie_score(?, ?, ?, ?, ?) / 3600.0 as goodie_score';
 
-        $state = $con->selectOne(
+        return $con->selectOne(
             $shiftScoreQuery,
             [
                 $user->id,
@@ -56,6 +52,18 @@ class Goodie
                 Carbon::now(),
             ]
         );
+    }
+
+    /**
+     * Returns the goodie score (number of hours counted for goodie score)
+     * Includes only ended shifts
+     */
+    public static function userScore(User $user): float
+    {
+        /** @var Database $db */
+        $db = app(Database::class);
+        $con = $db->getConnection();
+        $state = self::shiftScoreQuery($user);
 
         $shiftHours = 0;
         if ($state) {
--- a/tests/Feature/Helpers/GoodieTest.php
+++ b/tests/Feature/Helpers/GoodieTest.php
@@ -76,7 +76,7 @@ class GoodieTest extends ApplicationFeatureTest
 
         $result = Goodie::userScore($user);
 
-        $this->assertEquals(9.87, round($result, 2));
+        $this->assertEquals(8.87, round($result, 2));
     }
 
     private function deleteModels(): void
--- a/tests/Unit/Controllers/Admin/UserGoodieControllerTest.php
+++ b/tests/Unit/Controllers/Admin/UserGoodieControllerTest.php
@@ -31,7 +31,7 @@ class UserGoodieControllerTest extends ControllerTest
     public function testIndex(): void
     {
         $this->config->set('goodie_type', GoodieType::Tshirt->value);
-        $this->config->set('night_shifts', ['enabled' => false]);
+        $this->config->set('night_shifts', ['enabled' => false, 'start' => 2, 'end' => 8, 'multiplier' => 2]);
         $request = $this->request->withAttribute('user_id', 1);
         /** @var Authenticator|MockObject $auth */
         $auth = $this->createMock(Authenticator::class);
--- a/tests/Unit/Helpers/GoodieTest.php
+++ b/tests/Unit/Helpers/GoodieTest.php
@@ -10,23 +10,20 @@ use Engelsystem\Models\User\User;
 use Engelsystem\Models\Worklog;
 use Engelsystem\Test\Unit\HasDatabase;
 use Engelsystem\Test\Unit\TestCase;
-use Illuminate\Database\Connection;
-use Illuminate\Database\Query\Grammars\SQLiteGrammar;
-use PDO;
 
 class GoodieTest extends TestCase
 {
     use HasDatabase;
 
-    /**
-     * @covers \Engelsystem\Helpers\Goodie::shiftScoreQuery
-     */
-    public function testShiftScoreQuery(): void
-    {
-        $result = Goodie::shiftScoreQuery();
-
-        $this->assertEquals('0', $result->getValue(new SQLiteGrammar(new Connection(new PDO('sqlite::memory:')))));
-    }
+//    /**
+//     * @covers \Engelsystem\Helpers\Goodie::shiftScoreQuery
+//     */
+//    public function testShiftScoreQuery(): void
+//    {
+//        $result = Goodie::shiftScoreQuery();
+//
+//        $this->assertEquals('0', $result->getValue(new SQLiteGrammar(new Connection(new PDO('sqlite::memory:')))));
+//    }
 
     /**
      * @covers \Engelsystem\Helpers\Goodie::userScore
@@ -48,6 +45,8 @@ class GoodieTest extends TestCase
         parent::setUp();
 
         $this->initDatabase();
-        $this->app->instance('config', new Config(['night_shifts' => ['enabled' => false]]));
+        $this->app->instance('config', new Config(
+            ['night_shifts' => ['enabled' => false, 'start' => 2, 'end' => 8, 'multiplier' => 2]]
+        ));
     }
 }


db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
```diff
@@ -0,0 +1,107 @@
...
+                    
+                    DECLARE DAYS_DIFF int;
+                
+                    DECLARE shift_cursor CURSOR FOR
+                            SELECT s.start, s.end, se.freeloaded_by IS NOT NULL AS freeloaded 
+                            FROM users u
+                            JOIN shift_entries se ON se.user_id = u.id
+                            JOIN shifts s ON se.shift_id = s.id
+                            WHERE u.id = user_id
+                            AND s.end < CURRENT_TIMESTAMP;
```
olebittner: start and end timestamps are calculated by the engelsystem and then compared to a timestamp calculated by the database. This might cause issues if the database clock and engelsystem clock are not in sync.

Only a thought that came to mind, not sure how much of a problem this is in practice.

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
```diff
@@ -0,0 +1,107 @@
...
+                    OPEN shift_cursor;
+                    
+                    Set GOODY_SUM = 0;
+                    read_loop: LOOP
+                        FETCH shift_cursor INTO shift_start, shift_end, freeloaded;
+                        IF done THEN
+                                LEAVE read_loop;
+                        END IF;
+                        IF freeloaded THEN
+                            SET SHIFT_SUM = TIMESTAMPDIFF(SECOND, shift_start, shift_end) * -2;
```
olebittner: previously the negated night_shift_multiplier was used freeloading. Not sure which is better, just wanted to point it out.

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
```diff
@@ -0,0 +1,107 @@
...
+                                    SECOND, 
+                                    GREATEST(shift_start, NIGHT_START),
+                                    LEAST(shift_end, NIGHT_END)
+                                ));
+                                SET DAYS_DIFF = DAYS_DIFF + 1;
+                                SET START_OF_DAY = TIMESTAMPADD(DAY, 1 , START_OF_DAY);
+                            END WHILE;
+                            
+                            SET SHIFT_SUM = TIMESTAMPDIFF(SECOND, shift_start, shift_end) 
+                                                + (NIGHT_SHIFT_SECONDS * night_shift_multiplier);
```
olebittner: wouldn't the night_shift part be added once too many because it is also part of the TIMESTAMPDIFF?

resources/lang/de_DE/default.po
```diff
@@ -1178,8 +1178,8 @@ msgstr "iCal Export"
 msgid "JSON Export"
 msgstr "JSON Export"
 
-msgid "Night shifts between %d and %d am are multiplied by %d for the goodie score."
-msgstr "Nachtschichten zwischen %d und %d Uhr werden für den Goodie Score mit %3$d multipliziert."
+msgid "night_shift.goodie_score.hint"
+msgstr "Nachtschichtstunden zwischen %d und %d Uhr werden für den Goodie Score mit %3$d multipliziert. Stunden außerhalb der Nachtzeit zählen normal."
```
olebittner: might be nitpicking, but this chage was introduced because shift with as little as 15min during the nightshift would be multiplied. I think specifing hours in the hint might cause confusion in those cases.

```suggestion
msgstr "Die Zeit in der Nacht zwischen %d und %d Uhr wird für den Goodie Score mit %3$d multipliziert. Die Zeit außerhalb der Nacht zählt normal."
```

resources/lang/en_US/default.po
```diff
@@ -662,6 +662,9 @@ msgstr "Delete tag \"%s\""
 msgid "user.edit.goodie"
 msgstr "Edit goodie"
 
+msgid "night_shift.goodie_score.hint"
+msgstr "Night shift hours between %d and %d am are multiplied by %d for the goodie score. Hours outside of the night time are counted regularly."
```
olebittner:
```suggestion
msgstr "Night shift time between %d and %d am is multiplied by %d for the goodie score. Time outside of the night is counted regularly."
```

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
olebittner: nvm, I see you reduce the multiplier when calling the function

olebittner: now that I think of it, I think the multipler should be reduced in the function, so it is self contained and you don't need to remeber to reduce it when calling the function

resources/lang/en_US/default.po
MichiK: Maybe "Time worked between ... and ..." instead of "Night shift hours/time"?

Also, I would leave out the "time outside of the night is counted regularly" - that's the default after all.

MichiK:
I played with this a bit and tried a few exotic shift times and didn't manage to break the calculation.

However, it would be nice to see the hours that a shift is counted with somewhere (maybe in the tooltip and/or in the table in the user profile) to make the system a bit more intuitive and the calculation easier to understand.

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
rehleinBo: You are right this should be changed.
It was `Carbon::now()` - This parameter should be added

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
rehleinBo: AFAIK the multiplier was hard coded so no difference at all ;)

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
rehleinBo: We discussed about it - IMHO the multiplier is now used correctly - Before that change the whole shift was multiplied. 

Now only the night shift part is the amount of time in question.

rehleinBo:
@MichiK - This would unfortunatetly take a little more rework. 

I see it the same way. It will lead to questions and it's not very easy to calculate in your head except you are some kind of genius...

I'm willing to enhance it for 40C3 if @xuwhite will do the PHP part again :-)

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
xuwhite: @obit is right. I'll change this.

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
xuwhite: I think what @obit means is to input the real multiplier into the function and then reduce it by one in the function. When reintroducing the night shift multiplier as freeloading I will probably change that and agree with @obit even though I was the person that was against this in the first place. 😇

resources/lang/en_US/default.po
xuwhite: I like that wording.

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
xuwhite: Will do.

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
```diff
@@ -0,0 +1,108 @@
...
+        parent::__construct($schema);
+        $this->db = $this->schema->getConnection();
+    }
+
+    /**
+     * Run the migration
+     */
+    public function up(): void
+    {
+        $this->db->statement(
```
MyIgel, line 32: This needs a check to not be run during tests, see other migrations for reference:

```
[...]
.......WF...................EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE   61 / 1122 (  5%)
EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE  122 / 1122 ( 10%)
EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE......E  183 / 1122 ( 16%)
EEEEEEEEEEE.EEEEEEEE....................EEEEEEEEEEE...EEEEEEE  244 / 1122 ( 21%)
EEEEEEEEEEEEEEEEEE......EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE  305 / 1122 ( 27%)
EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE  366 / 1122 ( 32%)
EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE..................  427 / 1122 ( 38%)
...........EEEEEEEEEEEEEEEEEEEEEE............................  488 / 1122 ( 43%)
EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE....EEEEEEEEEEEEE...  549 / 1122 ( 48%)
................................................EE...........  610 / 1122 ( 54%)
.......................EEEE.E...........EEEE.................  671 / 1122 ( 59%)
.............................................................  732 / 1122 ( 65%)
....................................................EEEE.....  793 / 1122 ( 70%)
.............................................................  854 / 1122 ( 76%)
EEEEEEEEEEEEEEEEEE.E.........................................  915 / 1122 ( 81%)
...................................E........EEEEEEEEEEEEEEEEE  976 / 1122 ( 86%)
EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE 1037 / 1122 ( 92%)
EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE..........E.......E....... 1098 / 1122 ( 97%)
........................                                      1122 / 1122 (100%).

Time: 03:22.754, Memory: 134.50 MB

There were 571 errors:

1) Engelsystem\Test\Unit\Controllers\Admin\ConfigControllerTest::testIndex
Illuminate\Database\QueryException: SQLSTATE[HY000]: General error: 1 near "DEFINER": syntax error (Connection: default, SQL: CREATE DEFINER=`engelsystem`@`%` FUNCTION `goodie_score` (
[...]
```

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
```diff
@@ -0,0 +1,108 @@
...
+                    DECLARE shift_start TIMESTAMP;
+                    DECLARE shift_end TIMESTAMP;
+                    DECLARE freeloaded BOOLEAN;
+
+                    DECLARE START_OF_DAY TIMESTAMP;
+                    DECLARE NIGHT_START TIMESTAMP;
+                    DECLARE NIGHT_END TIMESTAMP;
+                    DECLARE NIGHT_SHIFT_SECONDS int;
+                    DECLARE SHIFT_SUM int;
+                    DECLARE GOODY_SUM int;
```
MyIgel, line 52: The consistent wording would be goodie
```suggestion
                    DECLARE GOODIE_SUM int;
```

(same for following)

resources/lang/de_DE/default.po
```diff
@@ -1148,8 +1148,8 @@ msgstr "iCal Export"
 msgid "JSON Export"
 msgstr "JSON Export"
 
-msgid "Night shifts between %d and %d am are multiplied by %d for the goodie score."
-msgstr "Nachtschichten zwischen %d und %d Uhr werden für den Goodie Score mit %3$d multipliziert."
+msgid "night_shift.goodie_score.hint"
+msgstr "Gearbeitete Zeit zwischen %d und %d Uhr werden für den Goodie Score mit %3$d multipliziert."
```
MyIgel, line 1152:
```suggestion
msgstr "Gearbeitete Zeit zwischen %d und %d Uhr wird für den Goodie Score mit %3$d multipliziert."
```

resources/lang/en_US/default.po
```diff
@@ -659,6 +659,9 @@ msgstr "Delete tag \"%s\""
 msgid "user.edit.goodie"
 msgstr "Edit goodie"
 
+msgid "night_shift.goodie_score.hint"
+msgstr "Time worked between %d and %d am are multiplied by %d for the goodie score."
```
MyIgel, line 663:
```suggestion
msgstr "Time worked between %d and %d am is multiplied by %d for the goodie score."
```

src/Helpers/Goodie.php
```diff
@@ -7,80 +7,9 @@
...
-use Illuminate\Database\Query\Grammars\MySqlGrammar;
 
 class Goodie
 {
-    /**
-     * Generates the query to sum night shifts
-     *
-     * Shifts and shift entries must be available via join
-     */
-    public static function shiftScoreQuery(): Expression
```
MyIgel: That function is still used in multiple places (admin_active, it is referenced in isNightShift),, please replace those too or re-add it n a compatible way

src/Helpers/Goodie.php
```diff
@@ -7,80 +7,9 @@
...
-    {
-        $nightShifts = config('night_shifts');
-
-        /** @var Database $db */
-        $db = app(Database::class);
-        $connection = $db->getConnection();
-
-        if (!$connection->getQueryGrammar() instanceof MySqlGrammar) {
-            return $connection->raw('0');
```
MyIgel: This will be a still required distinction imho

src/Helpers/Goodie.php
```diff
@@ -112,21 +41,22 @@ public static function userScore(User $user): float
...
-                self::shiftScoreQuery()->getValue($con->getQueryGrammar())
-            ))
-            ->where('users.id', $user->id)
-            ->join('shift_entries', 'users.id', 'shift_entries.user_id')
-            ->join('shifts', 'shift_entries.shift_id', 'shifts.id')
-            ->where('shifts.end', '<', Carbon::now())
-            ->groupBy('users.id')
-            ->first();
+        $shiftScoreQuery = /** @lang MySQL */
+            'SELECT goodie_score(?, ?, ?, ?, ?) / 3600.0 as goodie_score';
```
MyIgel, line 43: that new line seems to be useless?

src/Helpers/Goodie.php
```diff
@@ -112,21 +41,22 @@ public static function userScore(User $user): float
...
+
+        $state = $con->selectOne(
+            $shiftScoreQuery,
+            [
+                $user->id,
+                $nightShifts['start'],
+                $nightShifts['end'],
+                // at the moment supports negative multipliers, on trouble add max(0, x)
```
MyIgel:
```suggestion
```

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
```diff
@@ -0,0 +1,108 @@
...
+    public function up(): void
+    {
+        $this->db->statement(
+        /** @lang MySQL */
+            'CREATE DEFINER=`engelsystem`@`%` FUNCTION `goodie_score` (
+                    `user_id` INT,
+                    `night_shift_start` INT,
+                    `night_shift_end` INT,
+                    `night_shift_multiplier` INT,
+                    `now` TIMESTAMP
```
MyIgel, line 39:
```suggestion
                    `until` TIMESTAMP
```

or something similar, now is imho not the right wording or could be directly replaced with `NOW()`

db/migrations/2025_12_13_000000_add_goodie_score_calculation_function.php
MyIgel: Now there are less so its better debuggable ;)

PR #1593, opened 2025-12-17
Log: Add direkt link to messages and show errors in red

MyIgel:


resources/views/admin/log.twig
```diff
@@ -75,7 +75,13 @@
                             {%- endif %}
 
                             <tr>
-                                <td class="table-{{ td_type }} text-nowrap">{{ entry.created_at.format(__('general.datetime')) }}</td>
+                                <td class="table-{{ td_type }} text-nowrap">
+                                    <span id="message-{{ entry.id }}" class="ref-id"></span>
+                                    {{ entry.created_at.format(__('general.datetime')) }}
+                                    <small class="text-muted">
```
xuwhite: Screenshot 2025-12-18 at 14 05 17

It is very hard to see in theme 1 on errors

xuwhite: Screenshot 2025-12-18 at 14 06 34

on warnings it's even worse.

xuwhite: could you add this to design twig so it is easier to check on which designs this is the case?

MyIgel: I moved it to the right of the table which solves this problem too :)

PR #1596, opened 2025-12-19
New Engelsystem config page

MyIgel:
This MR extends the configuration abilities of the Engelsystem by a fair bit to allow for most config options to be set in the frontend.

The order of Importance of loading config options (first ones take precedence):
* ENV
* Database
* Config files
  * `config.php` (User created config)
  * `config.local.php` (Auto generated file when changing `write_back` configs)
  * `config.default.php` (Will be obsolete soon)
  * `app.php` (App state, no user changes)
* Defaults

Changes:
* Configuration options are defined in `app.php` to dynamically create multiple config pages
  * Have miultiple options like type, validation, info etc.
  * Most options are migrated, some remain for time / implementation complexity reasons
* Added export to automatically generate (parts of) the documentation page
* Updated some config options to match new format / better config
  * `email.port` `587` -> `465`
  * `email.tls` `null` -> `true`
  * `required_user_fields` from dict to list (former keys)
  * `enable_dect` `true` -> `false`
  * `locales` from dict to list (former keys)
* Added more twig fields to `/design`
* Added dotted array access to `Config` class
* Moved ServiceProvider initialization around to not expect configs before loading (from register to boot)
* Added `CarbonDay` to represent a day in Carbon
* Dynamic loading of some values like timezones, locales, ..
* Added validations `InMany` / `NotInMany`
* List items don't have to be "disabled" by `null` any more in `config.php` as the default is only used if it is not configured at all


src/Controllers/Admin/ConfigController.php
```diff
@@ -106,80 +73,155 @@ public function save(Request $request): Response
...
+            $fieldKey = Str::replace('.', '_', $key);
+            $default = $options['default'] ?? null;
+            $value = array_key_exists($fieldKey, $data) ? $data[$fieldKey] : $default;
 
             $value = match ($options['type']) {
                 'datetime-local' => $value ? Carbon::createFromDatetime($value) : $value,
+                'date' => $value ? CarbonDay::createFromDay($value) : $value,
+                'boolean' => !empty($value),
+                'number' => (float) $value,
+                'password' => $value === '**********' ? $this->config->get($key) : $value,
```
QueerCodingGirl: magic string should be a constant or config

src/Http/SessionHandlers/DatabaseHandler.php
```diff
@@ -33,7 +34,11 @@ public function write(string $id, string $data): bool
         $session->id = $id;
         $session->payload = $data;
         $session->last_activity = Carbon::now();
-        $session->user_id = auth()->user()?->id ?? null;
+        try {
+            $session->user_id = auth()->user()?->id ?? null;
+        } catch (Exception) {
+            // There might be some errors on early init
```
QueerCodingGirl, line 40: Catch and handle the exception types you expect, not all exceptions. Exceptions aren't Pokemon, don't catch them all ;)

src/Controllers/Admin/ConfigController.php
MyIgel: They are now (mostly) explicit variables :)

olebittner:
setting list config options (`REQUIRED_USER_FIELDS`, `LOCALES`) via ENV variables and then reloading the config page does disable the corresponding input field but still shows the old value.

olebittner:
`Default language` dropdown shows and allows selection of languages which are not part of the `Languages` list:
grafik

src/Http/SessionHandlers/DatabaseHandler.php
MyIgel: This only catches Exceptions and not Throwables so not all ;)
You gonna catch them all! <3

The reason why i ignore everything here is that the catch is only executed when something is really wrong and an exception ocures inside the app stack at an inconvenient point (normally when wanting to save the session while displaying another error) and thus any following exception would overlay the original problem.
If it still persists after that it would be fixed later so "intentionally left blank" is imho totally ok at that point.

MyIgel:
> setting list config options (`REQUIRED_USER_FIELDS`, `LOCALES`) via ENV variables and then reloading the config page does disable the corresponding input field but still shows the old value.

That looks to me like an interesting bug with the dropdown / values select logic (but should not prevent the merge)

MyIgel:
> `Default language` dropdown shows and allows selection of languages which are not part of the `Languages` list:

I could prevent that / validate it but that seems a bit overreacting? :D
(and its there to make reconfiguration easier / be less confusing)

olebittner:
It would be helpful if the config2docs output would contain the syntax for ENV variables especially for list like `REQUIRED_USER_FIELDS`, `LOCALES`.
It would also be helpful to have examples for the fields which have a `null` default i.e. for the timestamps like `buildup_start`.

PR #1597, opened 2025-12-20
Implement tags for shifts

MyIgel:
* Adds tags to shifts
* Can be filtered (better UI tbd; should be merged asap to prevent editing shifts afterwards at 39C3)


MyIgel:
A select is planned for a follow up patch; API is part of the ToDo list ;)

PR #1607, opened 2025-12-27
Define Timezone in iCal Export

### LINKED ISSUE #1604: Specify timezone in ics download
When downloading the ICS calendar of ones shifts, the file does not contain time zone information. This leads to incorrect import when the system's timezone is different to the event's time zone.

derrodirik:
Define the UTC Timezone in the Shift iCal export. This should fix #1604

### DIFF
--- a/tests/Unit/Controllers/FeedControllerTest.php
+++ b/tests/Unit/Controllers/FeedControllerTest.php
@@ -127,6 +127,47 @@ class FeedControllerTest extends ControllerTest
     }
 
     /**
+     * @covers \Engelsystem\Controllers\FeedController::ical
+     */
+    public function testIcalEmpty(): void
+    {
+        $this->request = $this->request->withQueryParams(['key' => 'fo0']);
+        $this->auth = new Authenticator(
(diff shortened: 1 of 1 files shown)


resources/views/api/ical.twig
```diff
@@ -4,6 +4,15 @@ BEGIN:VCALENDAR
 VERSION:2.0
 PRODID:-//-/{{ config('app_name') | replace(replacement) | raw }}//DE
 CALSCALE:GREGORIAN
+BEGIN:VTIMEZONE
+TZID:UTC
+BEGIN:STANDARD
+DTSTART:20250101T010000
```
MyIgel: That should be the first starting shift instead of a constant

MyIgel (changes requested):

MyIgel:
Hi, thx for the MR! Jus a small issue and it could be merged.

derrodirik:
Hi, if we do it that way, we need a fallback. So now, if the user has no shifts, the transition time is set to the start of the current day.

PR #1612, opened 2025-12-29
ui: prevent user nick icon/name from wrapping

teal-bauer:
## Summary
- Use CSS `background-image` for the angel icon instead of a separate `<span>` element with `&nbsp;`
- Keeps icon and name together as a single unit that won't break across lines
- Icon is slightly smaller (0.9em) for better visual balance

## Notes
- The `angel()` Twig macro is now unused and can be removed

### DIFF
--- a/includes/view/User_view.php
+++ b/includes/view/User_view.php
@@ -983,7 +983,8 @@ function User_Nick_render($user, $plain = false)
     }
 
     return render_profile_link(
-        '<span class="icon-icon_angel"></span>&nbsp;' . htmlspecialchars($user->displayName) . '</a>',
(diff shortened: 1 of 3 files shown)


MyIgel (changes requested):
The useful thing about the icon in its previous form is that its color automatically is the font color which can be seen esp. seen on `/users` and the 39c3 theme (21) -> the icon changed from the light font color to its black default color which has not enough contrast.

A better solution might be a span with nobreak around the previous icon?

teal-bauer:
Reworked - now just wraps icon+name in `<span class="text-nowrap">`. Keeps original icon.

PR #1619, opened 2026-01-05
feat(api): allow users to access own data without api privilege

teal-bauer:
## Summary

This PR allows authenticated users to access their own data via the `/api/v0-beta/users/self/*` endpoints without requiring the `api` privilege.

## Changes

Users can now access the following endpoints without the `api` privilege:
- `GET /api/v0-beta/users/self` - Get own user details
- `GET /api/v0-beta/users/self/angeltypes` - Get own angel types
- `GET /api/v0-beta/users/self/shifts` - Get own shifts
- `GET /api/v0-beta/users/self/worklogs` - Get own worklogs

## Implementation

Added `hasPermission()` method overrides in:
- `UsersController` - for `/users/{id}` and `/users/{id}/worklogs`
- `AngelTypeController` - for `/users/{id}/angeltypes`
- `ShiftsController` - for `/users/{id}/shifts`

The permission check allows access when the `user_id` parameter is `"self"` or matches the authenticated user's ID. For all other endpoints and other users' data, the `api` privilege is still required.

## Testing

All existing unit tests pass with these changes.

### DIFF
--- a/db/migrations/2026_04_22_000000_add_api_own_permission.php
+++ b/db/migrations/2026_04_22_000000_add_api_own_permission.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Engelsystem\Migrations;
+
+use Engelsystem\Database\Migration\Migration;
+use Illuminate\Database\Connection;
+use Illuminate\Database\Schema\Builder as SchemaBuilder;
+
+class AddApiOwnPermission extends Migration
+{
+    protected Connection $db;
+
+    public function __construct(SchemaBuilder $schema)
+    {
+        parent::__construct($schema);
+        $this->db = $this->schema->getConnection();
+    }
+
+    /**
+     * Run the migration
+     */
+    public function up(): void
+    {
+        $this->db->table('privileges')
+            ->insertOrIgnore([
+                'name' => 'api.own',
+                'description' => 'Access own data via the API',
+            ]);
+
+        $apiOwnId = $this->getPrivilegeId('api.own');
+
+        $groups = $this->db->table('groups')->select('id')->get();
+
+        $insertValues = [];
+        foreach ($groups as $group) {
+            $insertValues[] = ['group_id' => $group->id, 'privilege_id' => $apiOwnId];
+        }
+
+        if (!empty($insertValues)) {
+            $this->db->table('group_privileges')
+                ->insertOrIgnore($insertValues);
+        }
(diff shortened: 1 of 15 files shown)


MyIgel (changes requested):
I like the idea of allowing users to only view their "own" data (as a "soft launch" of the API) mostly but imho it would be  better to integrate it a bit more:
* Add `api.own` permission to all users to be able to disable it if needed and thus also show the info in the users setting
* The info endpoints should be viewable too (as in: which api pages are available -> needs additional checks before listing)
* Move the `hasPermission` check to a `OwnAuth` trait with an additional array of allowed "own" pages for the check
* Eventually only show the "available" api endpoints in the open api spec? by maybe adding a "own" tag? (just an idea to make it "easier" to understand what a user could do with the API)
* Calling `/users/self` without an login throws an error; that should be prevented too ;)
* I am missing tests or the feature

Some changes might be added features and thus need a new issue but the tests, access error, `OwnAuth` and `api.own` refactoring should be implemented

teal-bauer:
Reworked along the lines you suggested:

- New `api.own` privilege added in a migration and granted to every existing group, so behaviour is unchanged on upgrade. Admins can revoke it per group if they want to lock self-access down.
- The three inline `hasPermission()` overrides are now a single `OwnAuth` trait. Each controller just declares which action methods support self-access (`UsersController` -> `user`, `worklogs`; `AngelTypeController` -> `ofUser`; `ShiftsController` -> `entriesByUser`). The trait returns `null` for everything outside that allow-list so the regular permission pipeline runs.
- Layered access works as before for full-`api` users. Users with only `api.own` get the self routes and nothing else.
- `/users/self` unauthenticated used to 500 from a null deref. It now returns 401 via a small new `HttpUnauthorized` exception (mirrors `HttpForbidden`). Limited to self routes, so non-self unauth paths still 403.
- Settings page: `api.own` now also enables the API settings menu item, and the api section is gated so users with only `api.own` see a shorter "about" blurb pointing at `/api/v0-beta` and the OpenAPI spec. New translation strings for en_US and de_DE.
- Tests cover the trait directly (unauth self -> 401, unauth other -> null, no `api.own` -> null, self alias -> true, own id -> true, other id -> null, unknown method -> null) plus a wiring test on each of the three refactored controllers.

Deferred to follow-ups as we agreed: the info endpoint listing the routes a self-only user can hit, and the OpenAPI `own` tag.

Issue #1620, opened 2026-01-05
bug: mobile navbar-toggle hidden (pushed off screen) on wide content


escapedcat:
When the content is displaying i.e. a wide shift table the "hamburger icon" is moving of screen.

I'll try to fix this.

#### navbar-toggle missing in header

<img width="449" height="956" alt="Image" src="https://github.com/user-attachments/assets/15137af1-5fe1-4ab7-a190-de5a4569f181" />

#### need to scroll to right to see it

<img width="459" height="959" alt="Image" src="https://github.com/user-attachments/assets/15b47153-0fed-417a-bf8c-e3aaecfe90a3" />

### Desktop (useable)

<img width="1422" height="896" alt="Image" src="https://github.com/user-attachments/assets/f19dc092-7b1c-4819-beb8-41dff73eb20e" />

teal-bauer:
Sorry @escapedcat I didn't see your comment that you were going to work on this!

PR #1623, opened 2026-01-05
fix: mobile navbar hidden by wide content like shift calendar

escapedcat:
Fixes: #1620

I clicked around in the local dev app to see if this affecting anything else.
Can test more if there are known pitfalls.

Happy to adjust this.

### BEFORE
Image

### AFTER
image

- Add overflow-x: auto to #content to contain wide content
- Change shift-calendar.table-responsive to overflow-x: auto
- Wide content now scrolls within container instead of pushing navbar off-screen

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/resources/assets/themes/base.scss
+++ b/resources/assets/themes/base.scss
@@ -67,6 +67,12 @@ body {
   padding-top: 70px;
 }
 
+// Contain wide content within the main content area
+// so it doesn't cause page-level horizontal scroll that affects fixed navbar
+#content {
+  overflow-x: auto;
+}
+
 .footer a {
   color: $text-muted;
 }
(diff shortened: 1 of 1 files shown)


teal-bauer:
@escapedcat Since you asked for feedback on this PR, I just wanted to mention that when I was debugging a follow-up issue with my PR #1624, I was checking whether this alternative would have prevented this issue. Unfortunately it would probably cause the same unwanted clipping and scroll: `#content { overflow-x: auto; }` creates a clipping context on `#content`, so dropdowns will clip (vertically) and there will be ~15px unwanted scroll from Bootstrap `.row` negative margins. My patch had the same issues. I think the full fix will be a bit more complicated.

escapedcat:
Cool, thanks! 🤞

PR #1624, opened 2026-01-05
fix: keep navbar fixed when content overflows horizontally

### LINKED ISSUE #1620: bug: mobile navbar-toggle hidden (pushed off screen) on wide content

teal-bauer:
## Summary
- Fixes mobile navbar hamburger menu being pushed/scrolled off-screen when viewing wide content (e.g., shift calendar with many locations)
- Adds `overflow-x: clip` to body to prevent document-level horizontal scrolling
- Adds `overflow-x: auto` to `#content` to allow wide content to scroll within its container

Fixes #1620

### DIFF
--- a/resources/assets/themes/base.scss
+++ b/resources/assets/themes/base.scss
@@ -65,6 +65,13 @@ $navbar-height: 3.125rem;
 
 body {
   padding-top: 70px;
+  // Prevent horizontal scroll at body level so fixed navbar stays in place
+  overflow-x: clip;
+}
+
+#content {
+  // Allow wide content (e.g., shift calendar) to scroll horizontally within the content area
+  overflow-x: auto;
 }
 
 .footer a {


teal-bauer:
Navbar is now fixed, 100vw wide, and clipped, while `#content` gets to overflow horizontally:

Screen Shot 2026-01-05 at 18 23 18

escapedcat:
Did the same just before you :P #1623

teal-bauer:
@escapedcat Somehow I completely missed your comment in the original issue :~ Sorry for the duplication, I didn't intend to impinge on your fix ^^

I do think, just from a technical perspective, that this solution is a bit more generalized, but I'll leave the decision to the maintainers ^^

escapedcat:
All good, happens. Would have appreciated feedback on my PR but as long as this is being fixed I'm good. Closing my PR.

MyIgel:
@teal-bauer I just noticed that now all wide pages like the shift page scrolls too 0o
Can you have a look at this too?

teal-bauer:
Hmm, I did look at the shifts pages and they scrolled as intended, can you give me an example route (I can create my own test data) and device/form factor where this happens?

MyIgel:
On the shift page when using Firefox ( http://localhost:5080/shifts?action=view&shift_id=8110 ) I can scroll the content for ~ one margin:
Screenshot_20260112_183709

MyIgel:
In addition the dropdown when adding an angel to a angeltype ( http://localhost:5080/user-angeltypes?action=add&angeltype_id=145 ) is cut too 0o

teal-bauer:
Ah, this is on non-mobile breakpoints? I have an idea…

teal-bauer:
Ok so the issue is of course triggered the clipping boundary created by the overflow setting, so a quick fix would be to only have that on mobile breakpoints, something like

```
@media (max-width: map-get($grid-breakpoints, 'lg') - 0.02) {
    overflow-x: auto;
}
```

But the _cause_ is that something on that page is causing the page content to be wider than `#content`. I'll see if I can find a "proper" fix, I guess it's an interaction with the bootstrap base styles.

teal-bauer:
Bootstrap `.row` has -15px margins on each side, so when `overflow-x: auto` is on `#content`, you get -15px of scroll.
I tried subtracting the margin with some unholy `calc(...)` business, and that works, but the dropdowns are still clipped because `#content` on `/user-angeltypes?action=add&angeltype_id=1` is not full height, and the overflow creates the clipping context. I think an easier solution might be to just tell the navbar to stay put and limit its width - I'm testing that approach now.

teal-bauer:
Opened #1635 with a simpler approach - instead of messing with body/content overflow (which caused the dropdown clipping and row margin issues), it just constrains the navbar itself to viewport width. Much cleaner ^^

PR #1627, opened 2026-01-06
fix: anchor migration classname regex to start of string

### LINKED ISSUE #1621: bug: Migration wrapper removes _all_ digits and underscores from migration filename
The migration system uses `preg_replace('/\d+_/', '', $migration)` to deduce the class name from the file name, which removes ALL digit+underscore patterns, including e.g. the `2_` in `oauth2_`. So `oauth2_server_tables` becomes `oauthserver_tables`.

This just stumped me for way longer than it should have - I would propose changing the preg_replace to `preg_replace('/\A\d+_/', '', $migration)` to anchor it to the beginning of the filename.

teal-bauer:
## Summary
- Fixes regex that strips timestamp prefix from migration filenames
- Old regex `/\d+_/` removed ALL digit+underscore patterns, breaking migrations like `oauth2_server_tables` (became `oauthserver_tables`)
- New regex `/\A(?:\d+_)+/` anchors to start and matches the timestamp format precisely

Closes #1621

### DIFF
--- a/src/Database/Migration/Migrate.php
+++ b/src/Database/Migration/Migrate.php
@@ -154,7 +154,7 @@ class Migrate
     {
         require_once $file;
 
-        $className = Str::studly(preg_replace('/\d+_/', '', $migration));
+        $className = Str::studly(preg_replace('/^(?:\d+_)+/', '', $migration));
         /** @var Migration $class */
         $class = $this->app->make('Engelsystem\\Migrations\\' . $className);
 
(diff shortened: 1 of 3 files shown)


src/Database/Migration/Migrate.php
```diff
@@ -154,7 +154,7 @@ protected function migrate(string $file, string $migration, Direction $direction
     {
         require_once $file;
 
-        $className = Str::studly(preg_replace('/\d+_/', '', $migration));
+        $className = Str::studly(preg_replace('/\A(?:\d+_)+/', '', $migration));
```
MyIgel: I would prefer the easier to understand variation of

```suggestion
        $className = Str::studly(preg_replace('/^(?:\d+_)+/', '', $migration));
```

When you really create a migration with a new line i'm totally here to hear that story but otherwise i am on the side of "its much easier to understand" vs "capturing the edge case of an edge case that will not happen"

PR #1628, opened 2026-01-06
feat: mobile always show user menu items

escapedcat:
During Kongress I was constantly confused why I had to open "two" menus on mobile to find the "My shifts" link.
A dropdown in a dropdown? Why?

Trying to fix this with this PR. 

You can argue "why only the users menu". This was the most painful for me. Maybe language should be expanded too? I don't have an opinion on that.

Happy to adjust this.

### Before

image

Then click...

image

### After

image

### DIFF
--- a/resources/views/layouts/parts/navbar.twig
+++ b/resources/views/layouts/parts/navbar.twig
@@ -112,7 +112,7 @@
                     {{ menuUserHints() }}
 
                     {% include "layouts/parts/language_dropdown.twig" %}
-                    <li class="nav-item dropdown">
+                    <li class="nav-item dropdown dropdown-user-menu">
                         <a class="nav-link dropdown-toggle" href="#" role="button" data-bs-toggle="dropdown" aria-expanded="false">
                             {{ m.angel() }} {{ user.displayName }}
                         </a>
--- a/resources/assets/themes/base.scss
+++ b/resources/assets/themes/base.scss
@@ -222,6 +222,21 @@ table.table-sticky-header thead {
   }
 }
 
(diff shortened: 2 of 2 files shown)


resources/views/layouts/parts/navbar.twig
```diff
@@ -112,7 +112,7 @@
                     {{ menuUserHints() }}
 
                     {% include "layouts/parts/language_dropdown.twig" %}
-                    <li class="nav-item dropdown">
+                    <li class="nav-item dropdown dropdown--user-menu">
```
MyIgel: Technically the class does not need the doubled `--` here, everything else looks & works as expected.

(& will be merged after the 39C3 release)

escapedcat: True, it's somewhat [BEM](https://getbem.com/) influenced. Happy to adjust to whatever you prefer.

MyIgel: then i would prefer only one `-` & bem is relatively useless as the rest is mostly bootstrap (not that i would mind) :D

escapedcat: done

MyIgel:
Looks great, now i just have to find out if the ai :hankey: added more stuff than is needed or not and it can be merged :shrug: 

PS: Please dont use it, it takes > 1000% more review capacity to review ai-slopped-parrot- :hankey:...

MyIgel:
Thats the sub-set which is actually needed (as in: the same thats changed when actually expanding the menu) without the added random halluzinated "floof")

PR #1633, opened 2026-01-11
Update PHPUnit to v12

### LINKED ISSUE #1632: Update PHPUnit to v12
This project is currently using PHPUnit v9. Between version 9 and the current version 12, there have been several removals and deprecations that need to be addressed.

This update is a prerequisite to adopting Pest and then expanding into end-to-end testing.

  ## Goals

  The goal of this issue is to complete the PHPUnit v12 migration by ensuring:

  - All tests continue to pass
  - Test coverage remains similar
  - All deprecations / warnings / notices are resolved
  - Migrate phpunit.xml
  - Switch from phpunit annotations to attributes

Enterprize1:
Closes #1632 

These changes were produced by updating PHPUnit, running [Rector](https://github.com/rectorphp/rector-phpunit) and then fixing the remaining problems.

The main deprecations / removals are:

- Abstract classes that provide methods for other tests are not allowed to end in `Test` anymore, as PHPUnit assumes that these are tests themselves otherwise
- The `[Covers*]` attributes replace the `@covers` annotation (and must be defined on the class instead of an individual method)
- Same for other annotations, all can be expressed as attributes now (i.e. `Group`, `TestDox`)
- PHPUnit now wants test writers to differentiate between mocks and stubs. It warns if one creates a mock but then does not create any expectations on the created object. This does not work well with the current structure, where some mocks are created in superclasses and used only in some tests, so the `#[AllowMockObjectsWithoutExpectations]` escape hatch is used liberally in this PR.
-  The method `withConsecutive` was removed in PHPUnit 10: https://tomasvotruba.com/blog/how-to-upgrade-deprecated-phpunit-with-consecutive
- Any `DataProvider` now needs to be static
- The return type of `createMock`, `createStub`, etc. is now correct and does not require a `@var`-annotation anymore.


MyIgel:
Hi @Enterprize1,
thx for this MR! At the moment it hast some merge conflicts & we added newer tests that must be migrated first ;)
As i don't see any commands to migrate them by myself: Is there some time you know you could update the changes (so that we don't change too much in the engelsystem during that period)?

Enterprize1:
@MyIgel I just rebased to get to the current state. Another rebase should be quick if we find a good time for the update

MyIgel:
I rebased it again, updated some broken coverage annotations and implemented missing test cases -> imho it should be ready for merge now :D

PR #1635, opened 2026-01-13
fix: simplify navbar fix for wide content

### LINKED ISSUE #1620: bug: mobile navbar-toggle hidden (pushed off screen) on wide content

teal-bauer:
## Summary
- Replaces body/content overflow approach with direct navbar constraint
- Previous fix (#1624) caused dropdown clipping and unwanted scroll from Bootstrap row negative margins
- New approach: constrain navbar to viewport width directly

## Changes
```scss
.navbar.fixed-top {
  max-width: 100vw;
  overflow-x: clip;
}
```

This is much simpler and doesn't affect other elements like dropdowns.

## Testing
Tested on mobile viewport: shift calendar scrolls horizontally while navbar stays fixed at viewport width. Verified dropdowns (e.g., user selector on angeltype page) are not clipped. Desktop behavior unchanged.

Fixes #1620

### DIFF
--- a/resources/assets/themes/base.scss
+++ b/resources/assets/themes/base.scss
@@ -65,13 +65,12 @@ $navbar-height: 3.125rem;
 
 body {
   padding-top: 70px;
-  // Prevent horizontal scroll at body level so fixed navbar stays in place
-  overflow-x: clip;
 }
 
-#content {
-  // Allow wide content (e.g., shift calendar) to scroll horizontally within the content area
-  overflow-x: auto;
+// Keep navbar fixed at viewport width even when content is wider
+.navbar.fixed-top {
+  max-width: 100vw;
+  overflow-x: clip;
 }
 
 .footer a {


MyIgel:
When reproducing the issue in #1620 (opening the user page) on a small screen the buttons force the whole page to be wider and the menu scrolls out of view which is also not that optimal imho?

teal-bauer:
Hmm, it worked fine on the (wide) shifts page and it seems to work on the user page too, unless you mean a different route?

My shifts, menu open:
image

My shifts, menu closed:
image

Content scrolls, nav stays fixed:
image

Where do you see the bug still?

MyIgel:
Screenshot_20260116_230621

I just meant that it would be nice if the menu could stay in place instead of scrolling out but thats then probably for another iteration?

teal-bauer:
Hmm but it doesn't do that on my dev instance - see screenshot? Your screen doesn't have the 5px "margin" visible so it doesn't look like you are running the right CSS?

MyIgel:
I'm running commit 263ba4b which is the one from this MR so its the one that would be merged.

Is your user in some shifts / has some worklogs so that the table below is wider?

MyIgel:
Other pages with the problem are Arrive angels, Shifts overview etc.

At least on Firefox / Linux (Arch)

xuwhite:
This works in Safari but seems to be broken in Firefox and Chrome.

Firefox 
Screenshot 2026-01-29 at 18 55 51

Safari
Screenshot 2026-01-29 at 18 56 00

Chrome
Screenshot 2026-01-29 at 18 59 04

teal-bauer:
I hope I'll find time to look into it on the weekend, idk what's happening as I tested with Firefox and Chrome, I must've mixed up commits or stashed the wrong thing at some point...

teal-bauer:
This is Firefox on macOS: 

Users page
https://github.com/user-attachments/assets/85a5b4c6-7ad4-4776-b904-f0ec1deec16e

Shifts page with lots of content:
https://github.com/user-attachments/assets/3ef509df-4f54-4719-af0b-3f8fb2b76225

Same thing on Chrome on macOS and Firefox on Windows.
Do you have any ideas what could be different between the setups?

teal-bauer:
@xuwhite It looks like your might be setting the viewport different from the scaling or something? I have never seen anything like that happen. Does it happen when you use the "mobile view" t hing on the dev tools, or just resize your actual window smaller?

xuwhite:
> @xuwhite It looks like your might be setting the viewport different from the scaling or something? I have never seen anything like that happen. Does it happen when you use the "mobile view" t hing on the dev tools, or just resize your actual window smaller?

I used the "mobile view" thing on the dev tools in both firefox and chrome.

xuwhite:
Actually I tested it with my mobile devices (iPad and iPhone) on both firefox and chrome and it worked like it is supposed to. Maybe those dev tools are broken. @MyIgel could you please test it on Android?

MyIgel:
Screenshot_20260207-165443
Chrome / Android and also Firefox have the same problem "in real hardware" on my phone

MyIgel:
Anyhow; i'll merge it now to at least be able to release the 30C3 version

MyIgel:
The overflow also breaks safaris rendering due to a bug which cuts menu dropdown items :upside_down_face:, reverted in #1669

PR #1636, opened 2026-01-13
fix: show shifts on initial page load when all shifts are in past

### LINKED ISSUE #1430: Shift view is empty until new or newly logged-in user clicks "Filter" for the first time
The `Shifts` view is empty until a new (or newly logged-in) user clicks "Filter" for the first time:

https://github.com/user-attachments/assets/75928a4b-9178-428b-86a6-f64cae836b57

Is this on purpose?

I would've expected it to show all shifts until I've explicitly filtered them, not no shifts until I've applied filters. I think this might also have contributed to my [confusion](https://github.com/engelsystem/engelsystem/pull/1413#discussion_r1731180596) around the shifts view.

teal-bauer:
## Summary
- Fixes empty shifts view on initial page load when all shifts are in the past
- When today has no shifts, the filter now defaults to the first available day instead of today
- This prevents "No shifts found" until the user clicks Filter

The issue was that the date filter defaulted to today's timestamp, but the dropdown only shows days that actually have shifts. When all shifts are in the past, this caused a mismatch: the UI showed the first day with shifts, but the query filtered by today's date (which had none).

Tested on an instance where all shifts were in the past - shifts now display correctly on initial load.

Fixes #1430

### DIFF
--- a/includes/pages/user_shifts.php
+++ b/includes/pages/user_shifts.php
@@ -57,16 +57,24 @@ function update_ShiftsFilter_timerange(ShiftsFilter $shiftsFilter, $days)
 {
     $start_time = $shiftsFilter->getStartTime();
     if (is_null($start_time)) {
-        $now = (new DateTime())->format('Y-m-d');
-        if (in_array($now, $days)) {
+        $now = new DateTime();
+        $today = $now->format('Y-m-d');
+        if (in_array($today, $days)) {
             // Today has shifts, use current time
-            $start_time = time();
+            $start_time = $now->getTimestamp();
         } elseif (!empty($days)) {
-            // Today has no shifts, default to first day with shifts
-            $start_time = DateTime::createFromFormat('Y-m-d', $days[0])->getTimestamp();
+            // Today has no shifts, find the next upcoming day with shifts
+            $selectedDay = $days[0];
+            foreach ($days as $day) {
+                if ($day >= $today) {
+                    $selectedDay = $day;
+                    break;
+                }
+            }
+            $start_time = DateTime::createFromFormat('Y-m-d', $selectedDay)->getTimestamp();
         } else {
             // No days with shifts at all
-            $start_time = time();
+            $start_time = $now->getTimestamp();
         }
     }
 


includes/pages/user_shifts.php
```diff
@@ -58,13 +58,14 @@ function update_ShiftsFilter_timerange(ShiftsFilter $shiftsFilter, $days)
...
         $now = (new DateTime())->format('Y-m-d');
-        $first_day = DateTime::createFromFormat(
-            'Y-m-d',
-            in_array($now, $days) ? $now : ($days[0] ?? (new DateTime())->format('Y-m-d'))
-        )->getTimestamp();
-        if (time() < $first_day) {
-            $start_time = $first_day;
+        if (in_array($now, $days)) {
+            // Today has shifts, use current time
+            $start_time = time();
```
MyIgel: Imho explicitly using $now would be easier to read:

```php
        $now = new DateTime();
        if (in_array($now->format('Y-m-d'), $days)) {
            // Today has shifts, use current time
            $start_time = $now->getTimestamp();
```

includes/pages/user_shifts.php
```diff
@@ -58,13 +58,14 @@ function update_ShiftsFilter_timerange(ShiftsFilter $shiftsFilter, $days)
...
+        } elseif (!empty($days)) {
+            // Today has no shifts, default to first day with shifts
+            $start_time = DateTime::createFromFormat('Y-m-d', $days[0])->getTimestamp();
```
MyIgel: Generally it would be great to have some "the next day around" magic in that case (which of course would be a bit more advanced then just a simple check) as the first one might be pretty old when the Engelsystem gets used not only for a single short event but over months or years

teal-bauer: Yeah, in rescheduled we can define the "start of day" per event to take into account night shifts etc. maybe a setting like that would be useful here too. Otherwise snapping the view to the next "upcoming" shift day wouldn't be too hard either, I think.

teal-bauer:
- Using explicit `$now`
- When today has no shifts, it now finds the first day after today instead of always defaulting to `$days[0]`
- Falls back to `$days[0]` when all shifts are in the past, to show a non-empty default view (TBD)

includes/pages/user_shifts.php
```diff
@@ -57,15 +57,24 @@ function update_ShiftsFilter_timerange(ShiftsFilter $shiftsFilter, $days)
...
+        $now = new DateTime();
+        $today = $now->format('Y-m-d');
+        if (in_array($today, $days)) {
+            // Today has shifts, use current time
+            $start_time = $now->getTimestamp();
+        } elseif (!empty($days)) {
+            // Today has no shifts, find the next upcoming day with shifts
+            $selectedDay = $days[0];
+            foreach ($days as $day) {
+                if ($day >= $today) {
```
MyIgel, line 69: This compares strings with strings if i interpret it correctly; not sure if its a good idea to rely on PHPs magic conversions here :D

teal-bauer: Nothing magic about it, we know both sides are strings and we know their format is lexicographically well-ordered. Like, I don't mind changing this, but I guarantee you there is no magic conversion happening and this will never be a problem in practice.

teal-bauer: The alternative would be to either re-parse things into timestamps, or keep them as DateTime objects but then we have to format them everywhere we want to use them. So this seems like the simplest way to me

MyIgel: I think it should be fine too so lgtm

PR #1638, opened 2026-01-13
fix: sort users at database level before pagination

### LINKED ISSUE #1595: Fix pagination and sorting order for the admin "All users"-page
With the new pagination of the admin "All users"-page the sorting now works in an unexpected way. Instead of sorting all users by the chosen column and then paginating over the results it first paginates the users and then sorts only on the user-chosen page. For example it is not possible to find all users who have been freeloaded while using the pagination.

Also when clicking on a column header to sort I would expect the chosen page size to stay the same, currently it resets back to the default.

To achieve this we probably need to move the sorting into the database query, which seems to be non-trivial, as some columns are computed later.

teal-bauer:
The admin "All users" page was sorting in-memory after fetching a paginated result set. This meant clicking a column header like "freeloads" would only sort the users on the current page, making it impossible to find e.g. all users who have been freeloaded when browsing multiple pages.

Additionally, changing the sort order would reset the page size back to the default, losing any custom page size the user had selected.

This PR moves sorting into the SQL query instead of sorting in PHP after pagination. This ensures sorting works across all pages instead of only sorting the current page's results. Also preserves the page size parameter when clicking column headers to change sort order.

## Changes

Added a column map to translate user-facing column names to their actual database columns (with proper table prefixes for the joined tables). The query now joins `users_personal_data`, `users_contact`, and `users_state` so we can sort by columns in those tables. Freeloads count is computed as a subquery so it can also be sorted at the database level.

The view helper for column header links now passes through the `c` (page size) parameter when generating sort URLs.

Fixes #1595

### DIFF
--- a/includes/controller/users_controller.php
+++ b/includes/controller/users_controller.php
@@ -270,11 +270,11 @@ function users_list_controller()
     ];
 
     $order_by = 'name';
-    if ($request->has('OrderBy') && array_key_exists($request->input('OrderBy'), $columnMap)) {
-        $order_by = $request->input('OrderBy');
+    if ($request->query->has('OrderBy') && array_key_exists($request->query->get('OrderBy'), $columnMap)) {
+        $order_by = $request->query->get('OrderBy');
     }
 
-    $perPage = $request->get('c', config('display_users'));
+    $perPage = $request->query->get('c', config('display_users'));
     if ($perPage == 'all') {
         $perPage = PHP_INT_MAX;
     }


includes/controller/users_controller.php
```diff
@@ -250,26 +250,27 @@ function users_list_controller()
...
-            'force_active',
-            'force_food',
-            'got_goodie',
-            'shirt_size',
-            'planned_arrival_date',
-            'planned_departure_date',
-            'last_login_at',
-        ])
-    ) {
+    if ($request->has('OrderBy') && array_key_exists($request->input('OrderBy'), $columnMap)) {
```
xuwhite:
```suggestion
    if ($request->query->has('OrderBy') && array_key_exists($request->query->get('OrderBy'), $columnMap)) {
```
My IDE says $request->has() and $request->input() are deprecated.

includes/controller/users_controller.php
```diff
@@ -280,31 +281,27 @@ function users_list_controller()
...
+        ->leftJoin('users_personal_data', 'users.id', '=', 'users_personal_data.user_id')
+        ->leftJoin('users_contact', 'users.id', '=', 'users_contact.user_id')
+        ->leftJoin('users_state', 'users.id', '=', 'users_state.user_id')
+        ->selectSub(
+            ShiftEntry::selectRaw('COUNT(*)')
+                ->whereColumn('shift_entries.user_id', 'users.id')
+                ->whereNotNull('freeloaded_by'),
+            'freeloads'
+        )
+        ->orderBy($columnMap[$order_by]);
```
xuwhite:
```suggestion
        ->orderBy($columnMap[$order_by])
        ->orderBy('users.name');
```
Imho it looks more clean if the list is sub ordered by name.

includes/controller/users_controller.php
```diff
@@ -280,31 +281,27 @@ function users_list_controller()
...
+
+    $users = $query->paginate($perPage);
```
xuwhite: Why add this after the $query?

teal-bauer: Do you mean why the `->paginate` is on a separate line? Just find it cleaner that way

includes/controller/users_controller.php
```diff
@@ -250,61 +250,59 @@ function users_list_controller()
         throw_redirect(url('/'));
     }
 
+    // Map user-facing column names to actual database columns
+    $columnMap = [
+        'name' => 'users.name',
+        'first_name' => 'users_personal_data.first_name',
+        'last_name' => 'users_personal_data.last_name',
+        'dect' => 'users_contact.dect',
+        'arrived' => 'users_state.arrived',
```
MyIgel:
```suggestion
        'arrived' => 'users_state.arrival_date',
```

includes/controller/users_controller.php
```diff
@@ -250,61 +250,59 @@ function users_list_controller()
...
     }
 
-    $perPage = $request->get('c', config('display_users'));
+    $perPage = $request->query->get('c', config('display_users'));
     if ($perPage == 'all') {
         $perPage = PHP_INT_MAX;
     }
     $perPage = is_numeric($perPage) ? (int) $perPage : config('display_users');
 
     /** @var User[]|Collection|LengthAwarePaginator $users */
```
MyIgel, line 283: Doc string is now inconsistent

Issue #606, opened 2019-05-30
Restrict Usernames to Plain ASCII


jwacalex:
To prevent issues with unicode escaping or undefinable UTF-8 based nicknames, registration should only allow usernames match ``/^\w+$/``

herzmut:
I have mixed feelings about excluding names like céline, yıldırım, tannhäuser, any name with a character like underscore _, period . or dashes - as well as all of the emojis, and I suggest to deal with Unicode inputs properly.

jwacalex:
@herzmut the main issue is the "frontend" that has the issues. without a proper ascii representation it is currently not possible to find the angels because the names are at some random place in the lists.

MyIgel:
The inputs are not the problem ;) The issue is that when someone wants to search a username it can be complicated "the a with a round on it and then the l with a cross".

Imho not allowing them isn't the right way, we should find an option to maybe convert them on the fly to ascii only and use that as an additional search field.

herzmut:
On second degree search results, nicknames which are sorted with ascending Levenshtein distance could show up as suggestions. https://www.php.net/manual/en/function.levenshtein.php

jwacalex:
Regarding levenshtein i want to quote the following remark 

>  The levenshtein function processes each byte of the input string individually. Then for multibyte encodings, such as UTF-8, it may give misleading results.

https://www.php.net/manual/en/function.levenshtein.php#113702

herzmut:
Despite the fact that there are multibyte-compatible implementations, I quickly tested that proposition with the names mentioned above and found the results to be negligible: For the names mentioned above, each unicode character counted as 2 instead of 1 in their Levensthein distance result compared to their respective ASCII equivalent. However, another obstacle of Levensthein in PHP is the limitation on 255 characters.

MyIgel:
That only gets more complicated when using "non latin only" characters, the comparison function should be aware of it. But that needs some further investigation imho and maybe a test implementation comparing the various possibilities / methods

teal-bauer:
There are various foldings and normalized forms for Unicode strings, it wouldn't be a ton of work to make e.g. an input like `celine` match `céline`, or `yildirim` match `yıldırım`. It might not be possible to do on the fly in the DB, so adding a separate column for "normalized" nickname or a separate column only for searching containing stemmed and normalized forms of all searchable fields might be required / useful.

teal-bauer:
[...]

PR #1639, opened 2026-01-13
fix: use mb_stripos for case-insensitive search with umlauts

### LINKED ISSUE #1317: Case sensitive search for umlauts
The search function, e.g. on `/admin-active`, is case-sensitive with regards to umlaute. That is, searching for `ä` does not yield accounts  which only have `Ä` in their name; likewise, `ö` and `Ö` as well as `ü` and `Ü`.

teal-bauer:
PHP's `stristr()` function is not multibyte-aware, which means searching for lowercase umlauts (ä, ö, ü) won't match their uppercase equivalents (Ä, Ö, Ü). For example, searching for "ü" would fail to find users named "MÜLLER".

This affects the admin search pages for active users, free angels, and arrivals. The fix replaces `stristr()` with `mb_stripos()`, which handles multibyte characters correctly for case-insensitive matching.

Added a test to verify that umlaut searching works case-insensitively with `mb_stripos`. The application already uses UTF-8 throughout (database charset is utf8mb4, PHP's mb_internal_encoding is UTF-8), so no encoding issues to worry about.

Fixes #1317

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -231,9 +231,9 @@ function view(?string $template = null, array $data = []): Renderer|string
 
 /**
  * Normalize a string for search comparison.
- * Converts to ASCII and lowercase for diacritic-insensitive matching.
+ * Converts to ASCII and lowercase for diacritic- and compatibility-insensitive matching (e.g. "m²" → "m2").
  */
 function normalize_for_search(string $s): string
 {
-    return transliterator_transliterate('Any-Latin; Latin-ASCII; Lower', $s) ?: mb_strtolower($s);
+    return transliterator_transliterate('Any-NFKC; Any-Latin; Latin-ASCII; Lower', $s) ?: mb_strtolower($s);
 }
--- a/tests/Unit/Helpers/StringSearchTest.php
+++ b/tests/Unit/Helpers/StringSearchTest.php
@@ -66,6 +66,8 @@ class StringSearchTest extends TestCase
             'ASCII finds German ß' => ['Großmann', 'grossmann', true],
             'partial ASCII match' => ['Tannhäuser', 'hauser', true],
             'no match' => ['Schmidt', 'muller', false],
+            'NFKC compatibility: superscript 2' => ['42m²', '42m2', true],
+            'NFKC compatibility: superscript 3' => ['10m³', '10m3', true],
         ];
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
     },
     "require": {
         "php": ">=8.2.0",
+        "ext-intl": "*",
         "ext-json": "*",
         "ext-libxml": "*",
         "ext-mbstring": "*",


teal-bauer:
Extended to also add ASCII folding per #606 - searching "muller" now finds "Müller", "celine" finds "céline", etc. Uses PHP's Intl Transliterator.

tests/Unit/Helpers/StringSearchTest.php
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * @dataProvider provideAsciiFoldingCases
+     */
+    public function testAsciiFoldingSearch(
+        string $haystack,
+        string $needle,
+        bool $expectedFound
+    ): void {
+        $normalize = fn($s) => transliterator_transliterate('Any-Latin; Latin-ASCII; Lower', $s) ?: mb_strtolower($s);
```
MyIgel: Testing the third copy of an anonymous function is not the best idea as they will diff at some point.
Moving it to a helper (`src/helpers.php`) is imho the better place.

includes/pages/admin_active.php
```diff
@@ -241,7 +242,7 @@ function admin_active()
             $match = false;
             foreach ($tokens as $t) {
                 $t = trim($t);
-                if (!empty($t) && stristr($user->name, $t)) {
+                if (!empty($t) && mb_stripos($normalize($user->name), $normalize($t)) !== false) {
```
MyIgel: Using `\Illuminate\Support\Str::contains` here is more explicit and thus easier to read (without much overhead)

tests/Unit/Helpers/StringSearchTest.php
teal-bauer: Yeah I wasn't sure if this needed tests at all since it's standard library stuff. Not sure what the helper would contain though, the lambda is already pretty tiny?

includes/pages/admin_active.php
teal-bauer: But Illuminate Str::contains doesn't do any normalization? Or do you mean sth like `Str::contains($normalize(...), $normalize(...))`?

teal-bauer: Illuminate also does mb_strtolower for each needle, and then non-mb str_contains for each (needle,haystack), I am not sure those are all equivalent to mb_stripos?

teal-bauer: Checked PHP docs: after `mb_strtolower`, `str_contains` does byte comparison. So "MÜLLER" → "müller", but `str_contains("müller", "muller")` is false because the ü bytes (C3 BC) don't match u (75).

So `Str::contains` handles case but not diacritics. For this PR we need both - that's what the Transliterator does (strips diacritics before comparison).

teal-bauer:
Moved the normalize lambda to `normalize_for_search()` helper in `src/helpers.php`. Tests now cover the helper directly.

src/helpers.php
```diff
@@ -228,3 +228,12 @@ function view(?string $template = null, array $data = []): Renderer|string
...
     return $renderer->render($template, $data);
 }
+
+/**
+ * Normalize a string for search comparison.
+ * Converts to ASCII and lowercase for diacritic-insensitive matching.
+ */
+function normalize_for_search(string $s): string
+{
+    return transliterator_transliterate('Any-Latin; Latin-ASCII; Lower', $s) ?: mb_strtolower($s);
```
MyIgel: After some tests it would be great to add `Any-NFKC` beforehand so that for example `42m²` can be found by using `42m2`.
Also adding `"ext-intl": "*",` to `composer.json` makes it explicit that the php intl extension is used.

teal-bauer:
Added `Any-NFKC` before `Any-Latin`, and added `ext-intl` as explicit composer require, tests updated.

Issue #1145, opened 2023-05-16
Date and time is not fully visible


xuwhite:
I'm using a Mac book Air M1 with firefox
The dates in the shift view are not fully visible which is pretty annoying
<img width="1440" alt="Screenshot 2023-05-16 at 15 00 21" src="https://github.com/engelsystem/engelsystem/assets/73076134/54b3e8bf-953a-456a-bc15-6e9f271489e4">

PR #1640, opened 2026-01-13
fix: use content-based width for shift calendar time column

teal-bauer:
The time column in the shift calendar had a fixed 50px width which caused date and time labels to be cut off, particularly with longer locale-specific date formats.

This changes the column to use `max-content` sizing so it adapts to whatever content it needs to display.


MyIgel:
It would need some space to the right too:

Screenshot_20260117_231734

MyIgel:
Also it is not related to #1145 as the problem there are the date/time dropdown selects for searching

PR #1642, opened 2026-01-14
fix: prevent shifts filter date/time values from being cut off

### LINKED ISSUE #1145: Date and time is not fully visible

### LINKED ISSUE #1637: Shifts filter: date/time values cut off by picker icons
On the shifts page (`/user-shifts`), the date dropdown and time input values are partially hidden:

- Date select values are truncated (e.g., "2026-01-..." instead of full date)
- Time input values are cut off by the native browser clock icons (e.g., "13:0" instead of "13:01", "23:5" instead of "23:59")

This occurs on non-mobile desktop viewports. The column widths and native picker controls don't leave enough space for the full values to display.

teal-bauer:
## Summary
- Add CSS `min-width: max-content` to date selects and time inputs to prevent truncation
- Use `col-auto` instead of fixed-width grid columns so date/time groups size to content
- Add horizontal padding to the separator dash

On mobile, elements still stack as before.

Fixes #1637
Fixes #1145

### DIFF
--- a/resources/assets/js/forms.js
+++ b/resources/assets/js/forms.js
@@ -240,7 +240,7 @@ ready(() => {
       allowHTML: true,
       shouldSort: false,
       shouldSortItems: false,
-      removeItemButton: true,
+      removeItemButton: element.multiple,
       classNames: {
         containerInner: 'choices__inner form-control',
       },


xuwhite:
It fixes the mentioned problem but sadly it makes a problem that existed before worse. The cause is most probably that all shifts existing are dated in 2025. This happens sometimes if I click on the dropdown.

Before:
Screenshot 2026-01-29 at 18 32 07

Now: 
Screenshot 2026-01-29 at 18 30 26

It is out of scope here but it would be nice to get it fixed. (I will probably write another issue for that.)

teal-bauer:
@xuwhite Ok, I managed to reproduce your issue: There is a tiny, almost invisible (black on dark gray) [x] button overlapping the text that clears the selection, which causes Choices.js to mark everything unselected and then collapse to nothing. I'll implement a fix...

teal-bauer:
`removeItemButton: true` was set globally for all Choices.js selects, including single-selects like the date/time filters. This added a nearly invisible remove button (black × on dark themes) that, when clicked, cleared the selection and collapsed the container to ~42px — making it unusable.

Fix: `removeItemButton: element.multiple` — only multi-selects (location/type/filled filters) get the remove button. Single-selects can't be cleared, so the collapse can't happen.

xuwhite:
Thank you. Works like a charm. ☺️

PR #1643, opened 2026-01-14
fix: show all registration validation errors together

### LINKED ISSUE #1444: not all registration validation errors show up together
validateUniqueUsername(), validateUniqueEmail() are called after input validations and each other, so not all validation errors are shown together 

steps to reproduce:
- go to /register
- enter a nick and e-mail that's already taken
- leave one required input empty
- submit with register button

expected behaviour 
- all validation errors ("nick not unique", "e-mail not unique", "required input empty") are shown together 

real behaviour 
- only "required input empty" validation error is shown
- after entering the missing input and submitting with "register" button only "nick not unique" validation error is shown
- finally after changing the nick to a unique nick, "e-mail not unique" validation error is shown

teal-bauer:
Changed validator to collect all errors before throwing, so users see all validation problems at once instead of one at a time.

Fixes #1444

### DIFF
--- a/src/Factories/User.php
+++ b/src/Factories/User.php
@@ -154,11 +154,31 @@ class User
             $validationRules['tshirt_size'] = $this->isRequired('tshirt_size') . '|shirt-size';
         }
 
-        $data = $this->validate($rawData, $validationRules);
+        // Run validation but don't throw yet - collect all errors first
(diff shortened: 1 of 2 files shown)


MyIgel:
Fixed formatting, works great! <3

PR #1646, opened 2026-01-14
feat: add calendar subscription URL for iCal export

### LINKED ISSUE #413: iCal Export should trigger calendar abonnement instead of downloading an ics file

teal-bauer:
Adds a `content-disposition: inline` header for the iCal endpoint with file name containing engelsystem and user name.

~~Add a `subscribe` parameter to the iCal endpoint. When set, the Content-Disposition header is omitted, allowing calendar apps to subscribe to the feed instead of downloading it as a file.~~

~~The settings page now shows both URLs: the existing download link plus a new subscription URL that calendar apps can poll for updates.~~

Fixes #413

### DIFF
--- a/includes/pages/user_shifts.php
+++ b/includes/pages/user_shifts.php
@@ -382,7 +382,6 @@ function view_user_shifts()
                 'msg'  => msg(),
                 'tag_id'  => $tagId,
                 'shifts_table'  => $shiftCalendarRenderer->render(),
-                'ical_text'     => div('mt-3', ical_hint()),
                 'filter'        => __('Filter shifts'),
                 'filter_toggle' => __('shifts.filter.toggle'),
                 'expand_toggle_title' => __('shifts.toggle.title'),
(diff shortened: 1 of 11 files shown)


MyIgel:
Is it even useful to have the "forced download" then (besides of being able to set the file name which is defined by the url path which should be changeable?) also the button on the user page still downloads it and is most probably the "more" used one where it is, on the other side, also confusing to download the file imho

teal-bauer:
Honestly hadn't thought about it too much, just didn't want to change an existing feature "nebenbei". Some people might want to import ICS files, or operators might want to avoid the server load of subscription fetches possibly, but TBH this is such a cheap operation and easily cached that "display by default" is probably the sane thing to do. I can adapt it so the ICS is simply rendered by default, and if users want to download they can just use their browser's functionality like Shift-Click or Save Link As.

teal-bauer:
Done - removed the forced download and simplified the settings page to show a single "iCal Feed" URL.

> besides of being able to set the file name which is defined by the url path which should be changeable?

Not sure why the filename/path should be configurable, can you elaborate?

teal-bauer:
Now we give the browser a filename hint for saving, is that what you meant?

MyIgel:
I added a direct link on the "iCal feed" button to eventually encourage using it instead of downloading

xuwhite:
I unified the subscription feed and tried to improve the wording. Additionally I removed the old iCal export on /my-shifts that was just a duplicate of the buttons on top.

xuwhite:
Screenshot 2026-03-09 at 13 45 50
Screenshot 2026-03-09 at 13 46 12

teal-bauer:
If the UI was click-to-select or -to-copy, or an actual URL I can lonk-press, that would be super :D

MyIgel:
@xuwhite there is still a `%ical_text%` placeholder & i am not sure if its that useful to just remove the hint completely?
The link itself in code tags could be made clickable if adding an `<a href="...` inside the `<code>` and styling it accordingly.

xuwhite:
I rebased this PR. @MyIgel please take another look at it.

PR #1647, opened 2026-01-14
feat: send welcome email to newly registered users

### LINKED ISSUE #1546: Send Welcome E-Mail to newly registered Angels.
We received questions from multiple users wo were confused, that they didn't get a e-mail after registration.
Users appear to be so used to getting e-mail confirmation messages, that they think they misspelled their email if they don't get one from Engelsystem. 

I was also unsure, if our e-mail setup is broken, but later saw, that there is no template for such emails.

The feature request would be something like:
- Add template for welcome e-mail. (use text from existing registration confirmation e-mail)
- Send an e-mail when a new user is created.

teal-bauer:
Sends a welcome email to users when they register, confirming their account was created successfully.

Uses the existing email infrastructure and respects the user's email notification preferences.

Fixes #1546

### DIFF
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -169,10 +169,13 @@ msgid "email.user.welcome.subject"
 msgstr "Willkommen bei %s!"
 
 msgid "email.user.welcome"
-msgstr ""
-"Danke für deine Registrierung! Dein Konto wurde erfolgreich erstellt.\n"
-"\n"
-"Du kannst dich jetzt einloggen und für Schichten eintragen."
+msgstr "Danke für deine Registrierung! Dein Konto wurde erfolgreich erstellt."
+
+msgid "email.user.welcome.can_signup"
+msgstr "Du kannst dich jetzt einloggen und für Schichten eintragen."
+
+msgid "email.user.welcome.arrive_first"
+msgstr "Du kannst dich jetzt einloggen. Um dich für Schichten eintragen zu können, melde dich bitte vor Ort, wenn du angekommen bist."
 
 msgid "password.minimal_length"
 msgstr "Mindestlänge %d Zeichen"
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -165,6 +165,15 @@ msgstr ""
 "zurücksetzen kannst. Bitte benutze die E-Mail-Adresse, die du bei der Anmeldung "
 "verwendet hast."
 
+msgid "email.user.welcome.subject"
+msgstr "Willkommen bei %s!"
+
+msgid "email.user.welcome"
+msgstr ""
+"Danke für deine Registrierung! Dein Konto wurde erfolgreich erstellt.\n"
+"\n"
+"Du kannst dich jetzt einloggen und für Schichten eintragen."
+
 msgid "password.minimal_length"
 msgstr "Mindestlänge %d Zeichen"
 
--- a/resources/lang/de_DE/default.po
+++ b/resources/lang/de_DE/default.po
@@ -169,13 +169,13 @@ msgid "email.user.welcome.subject"
 msgstr "Willkommen bei %s!"
 
 msgid "email.user.welcome"
-msgstr "Danke für deine Registrierung! Dein Konto wurde erfolgreich erstellt."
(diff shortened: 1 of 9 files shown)


resources/lang/en_US/default.po
```diff
@@ -967,6 +967,15 @@ msgstr ""
...
 msgstr "Please visit %s to recover your password."
 
+msgid "email.user.welcome.subject"
+msgstr "Welcome to %s!"
+
+msgid "email.user.welcome"
+msgstr ""
+"Thank you for registering! Your account has been created successfully.\n"
+"\n"
+"You can now log in and sign up for shifts."
```
MyIgel: Shift sign up is only possible if configured (`!config('signup_requires_arrival') || config('autoarrive')`) and eventually shifts exist (although that would not be needed here as a feature toggle).

So the text should either be omitted or replaced by a generic "come to heaven to be marked as arrived" (equivalent to `render_user_arrived_hint`, better general wording welcome)

MyIgel (changes requested):

resources/lang/en_US/default.po
teal-bauer: Ah, fair point, so we should either check the configuration and have a specific message, or just limit ourselves to "You can now log in".

teal-bauer:
Updated to check `signup_requires_arrival` and `autoarrive` config and show appropriate message.

Side question: how chaos-specific should the default translations be? I kept this generic ("check in at the event") rather than mentioning "Heaven" / "Himmel" - but I noticed some other translations use these terms freely. Is there a preference for keeping defaults generic vs. using familiar chaos terminology?

src/Events/Listener/Users.php
```diff
@@ -0,0 +1,28 @@
...
+        protected LoggerInterface $log,
+        protected EngelsystemMailer $mailer
+    ) {
+    }
+
+    public function created(User $user): void
+    {
+        $this->mailer->sendViewTranslated(
+            $user,
+            'email.user.welcome.subject',
```
MyIgel, line 23: This resolves to `Welcome to [username]!`, i don't think that this was intended ^^

(data needs the app name before the username, added as an unnamed array value)

src/Events/Listener/Users.php
```diff
@@ -0,0 +1,28 @@
...
+            'emails/user-welcome',
+            ['username' => $user->displayName]
```
MyIgel: The username is not used in the mail text, eventually we could add it somewhere useful?

tests/Unit/Factories/UserTest.php
```diff
@@ -55,6 +56,10 @@ public function setUp(): void
         $this->app->instance(Authenticator::class, $this->app->make(Authenticator::class));
         $this->app->alias(Authenticator::class, 'authenticator');
 
+        $dispatcher = $this->createMock(EventDispatcher::class);
+        $dispatcher->method('dispatch')->willReturn([]);
```
MyIgel: Technically it would be `null` here ;)

resources/lang/de_DE/default.po
```diff
@@ -165,6 +165,18 @@ msgstr ""
...
+msgstr "Willkommen bei %s!"
+
+msgid "email.user.welcome"
+msgstr "Danke für deine Registrierung! Dein Konto wurde erfolgreich erstellt."
+
+msgid "email.user.welcome.can_signup"
+msgstr "Du kannst dich jetzt einloggen und für Schichten eintragen."
+
+msgid "email.user.welcome.arrive_first"
+msgstr "Du kannst dich jetzt einloggen. Um dich für Schichten eintragen zu können, melde dich bitte vor Ort, wenn du angekommen bist."
```
MyIgel:
```suggestion
msgstr "Du kannst dich jetzt einloggen. Um dich für Schichten eintragen zu können, melde dich bitte vor Ort an."
```
Would be a more fitting translation (eventually even without ` an`)?

MyIgel (changes requested):
> and respects the user's email notification preferences

It doesn't and that's imho the right way to do it -> A user should be notified that an account got created in any case.

MyIgel:
Generally texts should match those used anywhere else; the "Heaven" / "Help Desk" terminology is indeed a point that might need some "cleanup"

teal-bauer:
I think this is ready for another review?

PR #1648, opened 2026-01-14
feat: allow angel types to permit shift signup before arrival

### LINKED ISSUE #579: Allow Angeltypes to Signup before they are marked arrived
Sometimes angel groups need to sign up shifts before they are marked arrived. this is not possible with a global "cannot sign up when not arrived" flag, instead this should be on a angel type basis

teal-bauer:
Adds a per-angel-type setting `signup_allowed_before_arrival` that overrides the global `signup_requires_arrival` config. When enabled on an angel type, users can sign up for shifts of that type even before being marked as arrived.

This is useful for teams that want to let trusted members commit to shifts before the event starts, while keeping the arrival requirement in place for other angel types.

Fixes #579

### DIFF
--- a/includes/model/Shifts_model.php
+++ b/includes/model/Shifts_model.php
@@ -554,7 +554,7 @@ function Shift_signup_allowed_angel(
         return new ShiftSignupState(ShiftSignupStatus::NOT_YET, $free_entries);
     }
 
-    if (config('signup_requires_arrival') && !$user->state->arrived) {
+    if (config('signup_requires_arrival') && !$user->state->arrived && !$angeltype->shift_signup_before_arrival) {
         return new ShiftSignupState(ShiftSignupStatus::NOT_ARRIVED, $free_entries);
     }
 
--- a/includes/model/Shifts_model.php
+++ b/includes/model/Shifts_model.php
@@ -223,7 +223,8 @@ function NeededAngeltypes_by_ShiftsFilter(ShiftsFilter $shiftsFilter)
             `angel_types`.`id`,
             `angel_types`.`name`,
             `angel_types`.`restricted`,
-            `angel_types`.`shift_self_signup`
+            `angel_types`.`shift_self_signup`,
+            `angel_types`.`shift_signup_before_arrival`
         FROM `shifts`
         JOIN `needed_angel_types` ON `needed_angel_types`.`shift_id`=`shifts`.`id`
         JOIN `angel_types` ON `angel_types`.`id`= `needed_angel_types`.`angel_type_id`
@@ -241,7 +242,8 @@ function NeededAngeltypes_by_ShiftsFilter(ShiftsFilter $shiftsFilter)
             `angel_types`.`id`,
             `angel_types`.`name`,
             `angel_types`.`restricted`,
-            `angel_types`.`shift_self_signup`
+            `angel_types`.`shift_self_signup`,
+            `angel_types`.`shift_signup_before_arrival`
         FROM `shifts`
         JOIN `needed_angel_types` ON `needed_angel_types`.`shift_type_id`=`shifts`.`shift_type_id`
         JOIN `angel_types` ON `angel_types`.`id`= `needed_angel_types`.`angel_type_id`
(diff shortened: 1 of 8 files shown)


src/Models/AngelType.php
```diff
@@ -57,8 +58,9 @@ class AngelType extends BaseModel
         'contact_name'              => '',
         'contact_dect'              => '',
         'contact_email'             => '',
-        'restricted'                => true,
-        'requires_driver_license'   => false,
+        'restricted'                   => true,
+        'shift_signup_before_arrival'  => false,
+        'requires_driver_license'      => false,
         'requires_ifsg_certificate' => false,
```
xuwhite, line 64: Please either align all arrows or have only one space before each one. This looks weird. (I know this should be covered by the linter, sadly it's not.)

src/Models/AngelType.php
xuwhite: Do you want to do me a favor and delete those empty lines please?

src/Models/AngelType.php
```diff
@@ -90,8 +93,9 @@ class AngelType extends BaseModel
 
     /** @var array<string, string> */
     protected $casts = [ // phpcs:ignore
-        'restricted'                => 'boolean',
-        'requires_driver_license'   => 'boolean',
+        'restricted'                   => 'boolean',
+        'shift_signup_before_arrival'  => 'boolean',
+        'requires_driver_license'      => 'boolean',
         'requires_ifsg_certificate' => 'boolean',
```
xuwhite, line 99: Same arrow placing here.

xuwhite (commented):
CR: LGTM. I have some formatting comments.

teal-bauer:
All done!

includes/model/Shifts_model.php
```diff
@@ -554,7 +554,7 @@ function Shift_signup_allowed_angel(
         return new ShiftSignupState(ShiftSignupStatus::NOT_YET, $free_entries);
     }
 
-    if (config('signup_requires_arrival') && !$user->state->arrived) {
+    if (config('signup_requires_arrival') && !$user->state->arrived && !$angeltype->shift_signup_before_arrival) {
```
MyIgel, line 563: Thats not working as intended (and can't) here as its always false because the AngelType model is forceFilled by available values -> Please test the MR on all shift pages before creating

MyIgel (changes requested):

includes/view/AngelTypes_view.php
```diff
@@ -159,6 +159,18 @@ function AngelType_edit_view(AngelType $angeltype, bool $supporter_mode)
                                 __('angeltypes.restricted.info') . '"></span>',
                                 $angeltype->restricted
                             ),
+                        $supporter_mode
```
xuwhite: If the config is set to `signup_requires_arrival=false` this checkbox should be disabled and or not hidden and I think the info should be hidden

includes/model/Shifts_model.php
teal-bauer: I need to setup my dev env again, I unfortunately ran out of space for all the containers of my various projects :~ So it may take a few days to get this to a workable/testable state.

teal-bauer:
@MyIgel good catch, you were right. The check did read as always false on the shift calendar, but not because of `forceFill`'s allowlist behaviour. The issue is upstream: the raw `Db::select` queries in `NeededAngelTypes_by_shift`, `NeededAngeltypes_by_ShiftsFilter`, and `NeededAngeltype_by_Shift_and_Angeltype` only pulled `name`, `restricted`, and `shift_self_signup`. The new column was never in the result set, so `forceFill` kept the `$attributes` default of `false`. Added `shift_signup_before_arrival` to all six SELECTs.

The Shift detail page and the entry-create endpoint were already fine since they use a fully loaded Eloquent `AngelType`, but the calendar (which is where users hit signup) used the forceFilled one. Traced through all three paths to confirm.

@xuwhite done, the checkbox and its info tooltip now only render when `signup_requires_arrival` is enabled globally. The supporter-mode info row is hidden too. Also gated the controller write so the stored value persists if the global is toggled off and back on, rather than being silently reset to false by the absent form field.

PR #1649, opened 2026-01-14
feat(ical): add optional reminder alarms to iCal feed

### LINKED ISSUE #299: Include alarms in personal ics feed
The events in the iCal feed of your own shifts should include `VALARM`s so that a subscribed client automatically gets alerts for shifts.

This could be a user setting. At least some calendar clients provide an option to ignore alarms in subscribed calendars though.

teal-bauer:
Users can now enable 30-minute reminder alarms for their shifts in the iCal export. When enabled, each shift entry in the feed includes a VALARM component that triggers a reminder 30 minutes before the shift starts.

The setting is available on the API settings page alongside the existing iCal export options.

Closes #299


teal-bauer:
Could make the alarm timing configurable (15/30/60 min) later.

MyIgel:
Hmm somehow i would see it more as a get parameter to have notifications? On the other side its a more user specific thing so i am undecided.

A bigger "problem" is that the setting is only available when the user has api permissions which should imho be decoupled somehow but on the other side it does not fit on any of the other categories

teal-bauer:
I thought the `/settings/api` page where the iCal links are is accessible anyway, just not the remainder of the API?

But it would be easy enough to add a dropdown there and update the URLs on the fly, instead of persisting.

MyIgel:
Fair enough, i missed that so thats fine ^^

teal-bauer:
Ok so my understanding is correct that any user has access to the iCal & feed part of the API? Should we just put the dropdown on that page then and generate the links with query params?

teal-bauer:
Reworked to use a query param (?alarm=30) instead of storing it as a user setting. Simpler, no DB changes needed.

Not sure how to handle the migration file - I just deleted it since this branch hasn't been released. Is that ok?

xuwhite:
> Reworked to use a query param (?alarm=30) instead of storing it as a user setting. Simpler, no DB changes needed.
> 
> Not sure how to handle the migration file - I just deleted it since this branch hasn't been released. Is that ok?

Deleting the unused migration file is perfect, since it has never seen the main branch.

src/Controllers/SettingsController.php
xuwhite, line 36: Request is missing ;D 

```suggestion
        protected Response $response,
        protected Request $request
```

resources/views/pages/settings/api.twig
```diff
@@ -72,9 +72,57 @@
...
+                                if (checkbox.checked && minutes.value > 0) {
+                                    urlEl.textContent = baseUrl + '&alarm=' + minutes.value;
+                                } else {
+                                    urlEl.textContent = baseUrl;
+                                }
+                            }
+
+                            checkbox.addEventListener('change', updateUrl);
+                            minutes.addEventListener('input', updateUrl);
+                        })();
```
xuwhite: I would love if "ical_hide" would not collapse when clicking "Update URL".

xuwhite (changes requested):
I like this. Thank you very much. 
I have two rather small comments on this.
(I didn't test the functionality of the actual ical alarm.)

resources/views/pages/settings/api.twig
teal-bauer: As in preventDefault?

src/Controllers/SettingsController.php
teal-bauer: Huh. I am confused how this worked in my tests. Would it just silently swallow the error? Or would it inherit the Request from BaseController somehow? :o I'll look into this...

teal-bauer: Yeah, no idea what happened there. I added Request now :D

resources/views/pages/settings/api.twig
teal-bauer: Ah, you mean due to the GET request and full-page reload. Sure, I can whip something up.

teal-bauer: @xuwhite Are you running without JS or is the JS not working for some reason? I think it should hide the button and update via JS request normally?

MyIgel: Also there shouldn't be any JS on the page itself -> could you move it to a .js file? ;)

teal-bauer:
I think this should cover it, I'll check again when I've gotten some more sleep... if I don't forget.

MyIgel:
Looks like the noscript part is never shown (as the element is never shown because the buttons to show the URLs/Keys can't be pressed) -> imho it is not an "important core functionality" of the engelsystem and thus we could ignore the possibility of disabled JS here

xuwhite:
I added a yarn lint fix commit. Since the pre commit hook wasn't happy.

teal-bauer:
I think this is ready for another review?

resources/views/pages/settings/api.twig
```diff
@@ -72,9 +72,37 @@
...
+                            <div class="form-check mb-0">
+                                <input type="checkbox" class="form-check-input" name="ical_alarm_enabled"
+                                    id="ical_alarm_enabled" value="1"
+                                    {% if ical_alarm_enabled %}checked{% endif %}>
+                                <label class="form-check-label" for="ical_alarm_enabled">
+                                    {{ __('settings.api.ical_alarm_notify') }}
+                                </label>
+                            </div>
+                            <input type="number" class="form-control form-control-sm"
+                                style="width: 80px" name="ical_alarm_minutes" id="ical_alarm_minutes"
```
MyIgel, line 95: Styling should be implemented by adding classes / in css

resources/views/pages/settings/api.twig
```diff
@@ -72,9 +72,37 @@
...
+                        {% set icalParams = icalParams|merge({'alarm': ical_alarm_minutes}) %}
+                    {% endif %}
+                    <p>
+                        <code id="ical-url">{{ url('/ical', icalParams) }}</code>
+                    </p>
+                    <form method="get" action="{{ url('/settings/api') }}" class="mt-2" id="ical-alarm-form"
+                        data-base-url="{{ url('/ical', {'key': user.api_key}) }}">
+                        <div class="d-flex align-items-center gap-2 flex-wrap">
+                            <div class="form-check mb-0">
+                                <input type="checkbox" class="form-check-input" name="ical_alarm_enabled"
```
MyIgel, line 87: It would be better to use macros from `resources/views/macros/form.twig` here if possible and eventually for the numbers and button too

MyIgel (changes requested):

PR #1650, opened 2026-01-14
Fixes for 39C3 release

### LINKED ISSUE #1595: Fix pagination and sorting order for the admin "All users"-page

MyIgel:
* Log location name when changing shift
* Log token on JWT error
* Show message on own shifts page when freeloaded
* 39c3 design:
  * Fix selected bg color in searchable drowdown
  * Fix color of progress bar on shift
* Fix users sorting in shift
* Allow searching users by contact infos
* Fix string input length validation date detection (broke when saving `p3000` as worklog description)
* Fix en_US plural texts, update de_DE text
* Fix saving user while email human gets enabled which resulted in db error (email can't be null)
* Fix /users paginated sorting & cleanup, closes #1595

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -7149,16 +7149,16 @@
         },
         {
             "name": "phpunit/phpunit",
-            "version": "9.6.29",
+            "version": "9.6.34",
             "source": {
                 "type": "git",
                 "url": "https://github.com/sebastianbergmann/phpunit.git",
-                "reference": "9ecfec57835a5581bc888ea7e13b51eb55ab9dd3"
+                "reference": "b36f02317466907a230d3aa1d34467041271ef4a"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/sebastianbergmann/phpunit/zipball/9ecfec57835a5581bc888ea7e13b51eb55ab9dd3",
-                "reference": "9ecfec57835a5581bc888ea7e13b51eb55ab9dd3",
+                "url": "https://api.github.com/repos/sebastianbergmann/phpunit/zipball/b36f02317466907a230d3aa1d34467041271ef4a",
+                "reference": "b36f02317466907a230d3aa1d34467041271ef4a",
                 "shasum": ""
             },
             "require": {
@@ -7180,7 +7180,7 @@
                 "phpunit/php-timer": "^5.0.3",
                 "sebastian/cli-parser": "^1.0.2",
                 "sebastian/code-unit": "^1.0.8",
-                "sebastian/comparator": "^4.0.9",
+                "sebastian/comparator": "^4.0.10",
                 "sebastian/diff": "^4.0.6",
                 "sebastian/environment": "^5.1.5",
                 "sebastian/exporter": "^4.0.8",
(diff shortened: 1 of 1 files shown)


includes/controller/shifts_controller.php
```diff
@@ -191,6 +191,7 @@ function shift_edit_controller()
                 'Updated shift \'' . $shifttypes[$shifttype_id] . ', ' . $title
                 . '\' from ' . $start->format('Y-m-d H:i')
                 . ' to ' . $end->format('Y-m-d H:i')
+                . ' in ' . $locations[$rid]
```
xuwhite, line 194: please also add this to shift deletions

resources/lang/en_US/default.po
```diff
@@ -16,6 +16,18 @@ msgstr ""
 "Language: en_US\n"
 "X-Poedit-SearchPath-0: .\n"
 
+# Translated plural form must exist to be shown
+msgid "There are unconfirmed angels in %d angel type. Angel type that needs approval:"
+msgid_plural "There are unconfirmed angels in %d angel types. Angel types that need approvals:"
+msgstr[0] "There are unconfirmed angels in %d angel type. Angel type that needs approval:"
+msgstr[1] "There are unconfirmed angels in %d angel types. Angel types that need approvals:"
```
xuwhite: Angel types that need approvals: => Angel types that need approval:

xuwhite:
for the theme commit please also fix this
Screenshot 2026-01-15 at 18 58 58

xuwhite:
Screenshot 2026-01-15 at 19 04 55
please add more space above the error

xuwhite:
The progress bar color doesn't work for me
Screenshot 2026-01-15 at 19 09 07

includes/view/User_view.php
```diff
@@ -317,7 +322,7 @@ function User_view_shiftentries($needed_angel_type)
         . '">' . htmlspecialchars($needed_angel_type['name']) . '</a>:</b> ';
 
     $shift_entries = [];
-    foreach (collect($needed_angel_type['users'])->sortBy('name') as $user_shift) {
+    foreach (collect($needed_angel_type['users'])->sortBy('name', SORT_STRING | SORT_FLAG_CASE) as $user_shift) {
```
xuwhite, line 326: please sort by name after sorting by params

xuwhite:
I'm done with my review 🎉

includes/view/User_view.php
MyIgel: not the right place but sure

xuwhite:
Screenshot 2026-01-17 at 14 18 44
please also darken "X"

xuwhite:
Screenshot 2026-01-17 at 14 20 29
This is in desperate need of contrast too

PR #1652, opened 2026-01-15
feat: email notifications for news comments and questions

### LINKED ISSUE #1550: Add email notification for questions and comments
It would be nice to have email notification for questions and comments on news. I realize that it's not entirely clear who should get these emails, maybe it makes sense to tie that to the shift coordinator role?

teal-bauer:
Notify news authors when someone comments on their news item, and notify users with `question.edit` permission when new questions are asked.

Resolves #1550

### DIFF
--- a/db/factories/User/SettingsFactory.php
+++ b/db/factories/User/SettingsFactory.php
@@ -24,6 +24,7 @@ class SettingsFactory extends Factory
             'email_goodie'     => $this->faker->boolean(),
             'email_shiftinfo' => $this->faker->boolean(),
             'email_news'      => $this->faker->boolean(),
+            'email_questions' => $this->faker->boolean(),
             'mobile_show'     => $this->faker->boolean(),
         ];
     }
--- a/db/migrations/2026_03_06_000000_add_email_questions_to_users_settings.php
+++ b/db/migrations/2026_03_06_000000_add_email_questions_to_users_settings.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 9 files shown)


MyIgel:
I think the notifications should be configurable (what notifications a user gets) even if that means adding more config options (`email_news_comments` / `email_questions_asked`?) to a users page (but only if they can create news / answer questions).
Imho the default should be enabled for news comments and disabled for questions to not spam every shico? And when a news get updated the updating user should also be notified?

xuwhite:
> I think the notifications should be configurable (what notifications a user gets) even if that means adding more config options (`email_news_comments` / `email_questions_asked`?) to a users page (but only if they can create news / answer questions). Imho the default should be enabled for news comments and disabled for questions to not spam every shico? And when a news get updated the updating user should also be notified?

I totally agree. For me it would be fine if only the user in `news.user_id` would get notified. Each user with permission should be able to configure messages in the user profile settings.

teal-bauer:
@MyIgel

> when a news gets updated the updating user should also be notified

Sorry, could you clarify who should be notified? Did you mean the original news author gets notified when someone else edits their post, or something else?

> email_news default

All other email preferences currently default to off (opt-in). Should we change `email_news` to default to true for new users (existing users unaffected), or keep it consistent with the rest?

teal-bauer:
> I totally agree. For me it would be fine if only the user in `news.user_id` would get notified. Each user with permission should be able to configure messages in the user profile settings.

Ok, but that's already how it works in this PR, unless I'm misunderstanding you?

Issue #1585, opened 2025-11-30
Menu not working without JS


MyIgel:
The Engelsystem should be somehow useable without JS and such the menu should too

PR #1653, opened 2026-01-15
feat(profile): add shift filter by state

### LINKED ISSUE #399: Feature Request: User view for next shifts
From a user perspective it is quite hard to have a quick look at my next shifts.
For the next shift I can click the button with the minute coundown. However while doing a shift that button is not productive because it shows the current shift.
The most appealing way for me would be a filter by already done, running and subscribed shifts, maybe even cancelled shifts In the view which appears after clicking my username.

teal-bauer:
Users can now filter their shifts by state (upcoming, running, completed) on the profile page. Filter buttons are displayed above the shifts table.

Closes #399

### DIFF
--- a/includes/controller/users_controller.php
+++ b/includes/controller/users_controller.php
@@ -159,6 +159,11 @@ function user_controller()
         }
     }
 
+    $shift_filter = $request->input('shift_filter', '');
+    if (!in_array($shift_filter, ['upcoming', 'running', 'completed'])) {
+        $shift_filter = '';
+    }
+
     $shifts = Shifts_by_user($user_source->id, true);
     foreach ($shifts as $shift) {
         // TODO: Move queries to model
@@ -232,6 +237,7 @@ function user_controller()
                 || $is_ifsg_supporter
                 || auth()->can('user.drive.edit')
                 || $is_drive_supporter,
+            $shift_filter,
         ),
     ];
 }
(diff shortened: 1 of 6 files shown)


MyIgel:
I am not completely convinced by that feature as implemented because the engelsystem end-user-facing-frontend (everything that users without special permissions can use) once had the "goal" of also being able to be used without JS (currently slightly broken: #1585) so not having those buttons without js (or using forms & filtering in php) might be better?

teal-bauer:
Hm, I didn't see any note to that effect in https://github.com/engelsystem/engelsystem/blob/main/DEVELOPMENT.md or https://github.com/engelsystem/engelsystem/blob/main/CONTRIBUTING.md or I wouldn't have gone this route. TBH I personally think the "no javascript" ship has sailed. On that page we have all the data available in the frontend, so doing a full request cycle would be wasteful and worse UX, but I can rebuild this feature to be fully server-side or attempt some progressive enhancement (although I don't know if Engelsystem has provisions for that built-in, in Rails I would just put Stimulus+Turbo to use...)

teal-bauer:
Rebuilt with server-side filtering + progressive enhancement:

- Filter buttons are now links with `?shift_filter=upcoming|running|completed`
- Server filters before rendering (works without JS)
- JS enhances with instant client-side filtering when starting from unfiltered view
- If page loads already filtered, JS falls back to normal navigation (can't show rows that aren't in the DOM)

xuwhite:
I love it 😍. (only functional review)
One comment:

Please add more contrast to the completed button.
Screenshot 2026-03-06 at 16 08 10

PR #1654, opened 2026-01-15
feat: add token auto-confirm and bulk actions for restricted angel types

### LINKED ISSUE #522: Auto-accept/approve some angels for restricted angel types
Approving all angels manually for restricted angel types can be a lot of work, especially since it is a two/three step process (click, confirm, scroll down again). I'd like to enter a list of nicknames ahead of time, and if angels on that list request to join the angel type, they should be granted this permission directly. That way we could just whitelist people allowed on the "experienced angel" type.

teal-bauer:
Implements two features for managing restricted angel type memberships:

1. **Token-based auto-confirmation**: admins can set a self_signup_token on restricted angel types. Users entering the token when joining are automatically confirmed without manual approval.

2. **Bulk approve/reject UI**: supporters can now select multiple pending users via checkboxes and confirm or deny them in a single action.

Closes #522

### DIFF
--- a/includes/controller/user_angeltypes_controller.php
+++ b/includes/controller/user_angeltypes_controller.php
@@ -494,7 +494,7 @@ function user_angeltypes_bulk_controller(): array
         throw_redirect(url('/angeltypes'));
     }
 
-    $selectedIds = $request->postData('selected_ids');
+    $selectedIds = $request->request->all('selected_ids');
     if (empty($selectedIds) || !is_array($selectedIds)) {
         error(__('angeltype.bulk.none_selected'));
         throw_redirect(url('/angeltypes', ['action' => 'view', 'angeltype_id' => $angeltype->id]));
--- a/includes/view/AngelTypes_view.php
+++ b/includes/view/AngelTypes_view.php
@@ -750,7 +750,10 @@ function AngelType_view_info(
                 continue;
             }
             $checkbox = '<input type="checkbox" name="selected_ids[]" value="' . $member->pivot->id . '" class="form-check-input user-select-checkbox">';
-            $unconfirmed_with_checkboxes[] = array_merge(['select' => $checkbox], (array) $member);
+            // Convert member to array preserving the dynamic 'name' property set earlier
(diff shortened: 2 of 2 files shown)


MyIgel:
Without having a look at the code too much: we already have the "QR Code" feature so another token to join would be imho confusing and feels a lot like duplicated maintenance work?

teal-bauer:
I haven't used the QR feature, but entering a short string on signup feels much simpler than making a QR code available. OTOH scanning a QR code at a meeting might be quicker (if it's a direct signin link), but it's probably not something you'd ever type. So for me these are complementing each other?

teal-bauer:
Ok so looking at the QR code feature, that gives you a full link with a time-limited JWT, so definitely not something you can easily "tell" someone, whereas the text token is for use cases like "Sign up as Translation Angel, enter token `C3LINGO`" in a chat. I think both have their own use cases.

xuwhite:
Thanks for the PR. I wanted to do this for a long time but never did. I just tried it and have some comments.

Could you please add some contrast for the checkboxes? They disappear on every other row.

Screenshot 2026-03-06 at 15 26 46

There is something wrong in this branch. The dropdown on /user-angeltypes?action=add is cut off in both safari and firefox. Can you please take a look at it?

Screenshot 2026-03-06 at 15 28 23

The table header has a checkbox too, but it does nothing. 
As a user I would expect this checkbox to be "select all" / "deselect all". 
I would love if this would work and would actually select and deselect all checkboxes when clicked, that would eliminate the need for the "confirm all" and "deny all" buttons. 

If we keep those buttons I would prefer to keep the coloring and have them in a different group (maybe even at the bottom of the table, but I am not sure about that).

Screenshot 2026-03-06 at 15 34 56

MyIgel:
The "cut of" add user page is a remainder of the menu fixes and already fixed in main; as this MR is not rebased it still persists.

Imho adding this as a plugin would be "better" as it still duplicates the qr feature somehow (and sending a sign up link vs a shorter code is imho not that much of a difference?)

PR #1656, opened 2026-01-18
fix(shifts): add right padding to time column

teal-bauer:
Follow-up to #1640 - adds padding between time text and border.

### DIFF
--- a/resources/assets/themes/base.scss
+++ b/resources/assets/themes/base.scss
@@ -312,6 +312,7 @@ table.table-sticky-header thead {
     flex-shrink: 0;
     width: max-content;
     min-width: max-content;
+    padding-right: 0.5rem;
   }
 
   .shift-card {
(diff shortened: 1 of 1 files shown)


MyIgel (changes requested):
When setting the pdding in that way the "table" header has a "break" too which is imho not that useful; in addition the "overlaying" content is half tee-through which is even more confusing:

Image
Image

teal-bauer:
@MyIgel good catches, thanks. Scoped the padding to `.tick` children only so the `.header` keeps its existing `padding: 5px 5px 5px 16px` and doesn't break. Also gave `.lane.time` a solid background via `var(--es-table-sticky-bg, var(--bs-body-bg))` (same pattern as `.table-sticky-header thead`), since `.lane`'s `$table-striped-bg` is semi-transparent and was letting scrolled content bleed through.

MyIgel:
After a rebase the css has no effect on the design 0o

xuwhite:
Screenshot 2026-07-25 at 22 35 29
It now looks like this.

PR #1664, opened 2026-02-18
introduce biome linter, update packages

xuwhite:
- remove eslint and prettier
- update packages
- use composer audit in ci


weeman1337 (approved):
I would prefer moving to the Vite stack including https://oxc.rs/

But no strong opinion tbh

PR #1671, opened 2026-03-07
Hide build warnings to be fixed by Bootstrap 6

MyIgel:
* import warnings can only be usably fixed by bootstrap 6 as far as i understand
* the others are annoying and could be fixed before but might need some effort


xuwhite:
Screenshot 2026-03-19 at 18 24 34
Pre commit hook says no.

PR #1673, opened 2026-03-08
Settings: Show info as text

MyIgel:
Although I think it looks way more cluttered, its an option to implement it that way:

Screenshot_20260308_164717
Screenshot_20260308_164635

### DIFF
--- a/resources/assets/themes/base.scss
+++ b/resources/assets/themes/base.scss
@@ -76,6 +76,14 @@ body {
   color: $text-muted;
 }
 
+.container {
+  &.config {
+    .form-check-label {
+      font-weight: $form-label-font-weight;
+    }
+  }
+}
+
 .text-inherit {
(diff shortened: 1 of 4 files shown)


teal-bauer:
I think that is MUCH better UX. Trading very little, otherwise mostly unused space, for obvious hints that don't require hovering over a comparatively tiny target? That's a big win :)

xuwhite:
could you please add font-weight 700 to the checkbox label?

Screenshot 2026-03-19 at 17 20 59


right now the config options have different font weights
Screenshot 2026-03-19 at 17 21 09

MyIgel:
@xuwhite could you please add that?

xuwhite:
> @xuwhite could you please add that?

Done

MyIgel:
Screenshot_20260320_200049

PR #1679, opened 2026-04-01
Show day of week on dates

MyIgel:
image
image

### DIFF
--- a/config/app.php
+++ b/config/app.php
@@ -155,6 +155,10 @@ return [
                 'teardown_end' => [
                     'type' => 'datetime-local',
                 ],
+                'enable_date_day' => [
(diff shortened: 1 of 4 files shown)


xuwhite:
Please add an option to the profile settings to disable this feature

MyIgel:
Imho its enough to have it configurable per event (and disabled by default as it is)
I would generally prefer not to introduce configs that change the UI just by a small bit as starting that would imho introduce too much variability / other independent state changes on the long run

xuwhite:
Please make it possible to disable this feature in a plugin.

MyIgel:
The event handling definitely needs a rewrite; its a mess that way (but legacy code so :shrug:)

PR #1684, opened 2026-04-04
Shifts filter: Unify shown shifts

MyIgel:
* Unify and fix shifts filter for angel types and locations
* Angel type & location shifts table: Use buttons (makes semantically no sense...)


includes/sys_template.php
xuwhite: this function is not used anymore please remove it.

PR #1689, opened 2026-04-25
Add support for plugins

MyIgel:
This adds the basis for Plugins to be added to the Engelsystem.
The most important basic things are already there, some other parts need further work and extension to be useable.

Current features:
* Management page under settings to install, uninstall and update plugins
  * Plugins have some metadata about heir state and namespaces in their `composer.json` (Plugins should use their own namespace); dependency resolution between plugins and / or env features is currently missing
* Adding custom controllers & routes
* Settings extension and configuration of own pages possible
* Adding and overwriting translations
* Registration of event handlers
* Adding middleware from config
* Per-plugin migrations (Plugins should use a unique prefix for their tables / columns)
* Extending .twig templates (even multiple plugins can extend the same template!, more blocks needed)
* Added demo plugin to showcase currently implemented functionality

Related documentations PR: https://github.com/engelsystem/engelsystem-doc/pull/11

At the moment the best way to extend core functionality is to wrap their classes via the app container; more events and thus event handling needed at some point.
A small working Plugin can be found at https://github.com/MyIgel/TranslationMapper

Needed features / tbd:
* "Theme"-Plugins (minimum config, easy configurability etc.)
* More events at useful parts of the Engelsystem with predefined interfaces which should preferably not change as often as engelsystem internal code / interfaces
* Refactor templates to be better reusable and especially extendable (more `{% block %}` elements to be extended / hooked to etc.)
* More Interfaces to extend of (allows easier wrapping of classes)
* Some possibility to have an overview about engelsystem plugins (discoverability, filter, semi-central-plugin-hub, ratings etc.) with "klick to install" instead of copying them to the `resources/plugins` directory
* Allow plugins to define dependencies (other plugins, system env, libraries etc.)
* tbd: Extending CSS/JS, eventually hooking sql queries etc.


src/Plugins/PluginServiceProvider.php
```diff
@@ -0,0 +1,217 @@
...
+
+        if ($isEnabled) {
+            $this->registerPluginProviders($plugin);
+            $this->registerPluginMiddleware($plugin);
+            $this->registerPluginEventHandlers($plugin);
+            $this->registerPluginConfigOptions($plugin);
+        }
+    }
+
+    protected function getNamespace(array $autoload, string $pluginPath): ?array
```
Enterprize1: This could just be typed as `array` as only the contents of the array may be null

src/Controllers/Admin/PluginsController.php
```diff
@@ -0,0 +1,206 @@
...
+        );
+        $baseDir = $plugin->getNamespacePath() . 'Migrations';
+        if (is_dir($baseDir)) {
+            $this->migration->setNamespace($plugin->getNamespace() . 'Migrations\\');
+            $this->migration->setOutput(function (string $message): void {
+                if (Str::startsWith($message, 'Skipping ')) {
+                    return;
+                }
+                $this->log->info($message);
+            });
```
Enterprize1, line 82: This changes the `migration` singleton and while this will probably be fine and no code will ever access the variable afterwards it would feel safer if we reset the fields afterwards or allow passing `namespace` and `output` to the `run(...)` call

resources/plugins/DemoPlugin/src/Controller.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
```
Enterprize1, line 2: There is a `declare(strict_types=1);` missing here

src/Plugins/PluginServiceProvider.php
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        if (!$isEnabled) {
+            $this->app->alias(DisabledPlugin::class, $namespacedPlugin);
+        }
+
+        $plugin = $this->app->make(
+            class_exists($namespacedPlugin) ? $namespacedPlugin : Plugin::class,
+            ['pluginInfo' => $pluginInfo],
+        );
```
Enterprize1, line 160: At first I thought this would already create the plugin and call the constructor, but if I understand this correctly now the alias above changes the instantiated class to the `DisabledPlugin`.

A structure like `if (!$isEnabled) {} elseif (class_exists) {} else {}` would communicate the control flow better I think

src/Plugins/PluginServiceProvider.php
```diff
@@ -0,0 +1,217 @@
...
+    {
+        $composerFile = $pluginPath . 'composer.json';
+
+        if (!is_readable($composerFile)) {
+            return;
+        }
+
+        $pluginInfo = json_decode(file_get_contents($composerFile), true);
+        if (!$pluginInfo || empty($pluginInfo['name']) || ($pluginInfo['type'] ?? null) !== 'engelsystem-plugin') {
+            return;
```
Enterprize1, line 77: We could have some kind of log or other notification here to surface composer.json errors instead of silently skipping the plugin

resources/plugins/DemoPlugin/lang/de_DE.po
Enterprize1, line 1: Generating the translations produces untracked .mo files in the plugin. Adding a .gitignore preventing this to the demo plugin would be a good template I think

src/Controllers/Admin/PluginsController.php
MyIgel: As this is the same "action" as "running migrate" (as in: it configures the migration to run and the runs the migrations) resetting it would be imho "wrong" (for example if some other plugins want to do things with it afterwards)

src/Plugins/PluginServiceProvider.php
MyIgel: I am not sure how that would make it much clearer without repeating code?

if the plugin is not enabled the DisabledPlugin class is used every time.
If its enabled it creates an instance, either of the plugins specific class  or the generic `Plugin` otherwise.

src/Plugins/PluginServiceProvider.php
MyIgel: I thought about this already but it is imho not the right point as its called on every request and thus would spam any log for random (semi-)empty folders (`$RECYCLE.BIN` anyone?)

src/Plugins/Plugin.php
```diff
@@ -0,0 +1,177 @@
...
+declare(strict_types=1);
+
+namespace Engelsystem\Plugins;
+
+use FastRoute\RouteCollector;
+
+class Plugin
+{
+    protected string $name;
+    protected string $pluginName;
```
olebittner, line 14: what is the difference between ```name``` and ```pluginName```? I mean I see that ```name``` is pulled from the json and ```pluginName``` is extracted from the path, but why have both if only ```pluginName``` is ever used?

src/Plugins/PluginServiceProvider.php
```diff
@@ -0,0 +1,217 @@
...
+        );
+        if (!$pluginNamespace) {
+            $pluginNamespace = $pluginName . '\\';
+            $pluginNamespacePath = $pluginPath;
+        }
+
+        $pluginInfo['plugin_name'] = $pluginName;
+        $pluginInfo['path'] = $pluginPath;
+        $pluginInfo['namespace'] = $pluginNamespace;
+        $pluginInfo['namespace_path'] = $pluginNamespacePath;
```
olebittner, line 92: won't this only contain the first namespace while it is possible to register multiple via the json? Depending on what the intention behind this info is, I think an array similar to the authors structure would be better.

src/Controllers/Admin/PluginsController.php
```diff
@@ -0,0 +1,206 @@
...
+
+        $this->log->info('Uninstalled plugin "{name}"', ['name' => $plugin->getPluginName()]);
+        $this->addNotification('plugin.uninstall.success');
+    }
+
+    protected function handleUpdate(AbstractPlugin $plugin, Plugin $model, ?string $baseDir): void
+    {
+        if ($baseDir) {
+            $this->migration->run($baseDir, Direction::UP);
+        }
```
olebittner, line 174: I'm not sure about the update functionality, because the plugin if once enabled stays enabled regardless of changes in the version numbers. This means the update plugin code will be executed, even if the migrations did not run yet. On the other hand the update button is shown, even if clicking it won't do anything if the new version doesn't provide any additional migrations.

Maybe it would be better if plugins are only enabled if the version matches, this ensures plugin code is only executed after migrations are run. The UI would then offer the choice to update the plugin (and thereby re-enabled it) or uninstall it.

src/Plugins/Plugin.php
MyIgel: The "name" is the one from composer (`vendorNamespace/pluginName`) and because i wanted to have at least some "compatibility" between composer field names it is still there; the pluginName is effectively the "plugin directory name" (`pluginName`) and thus the one thats currently shown to the user

src/Plugins/PluginServiceProvider.php
MyIgel: Thats basically the "main" namespace; the others are loaded (`PluginServiceProvider::registerNamespaces`) but indeed not referenced here.
Do you see a need to have them all?

src/Controllers/Admin/PluginsController.php
MyIgel: The "it must be initialised" part is still true for the "run the update" call but i see the "dont load code" part too; not sure which way would be better (and for example not clicking update would be bad too as it might disable important parts/features of the page, especially as there is not always anything to do?)

olebittner: I guess it won't matter if there are no new migrations. But can we expect the new version of the plugin to handle running without its migrations be executed first.

src/Plugins/PluginServiceProvider.php
olebittner: No particular need, just wanted to point out the difference

olebittner:
when two plugin register the same route you get a ```FastRoute \ BadRouteException```which makes sense, but you cannot easily recover, as the the plugin config page can no longer be opened and you cannot disable or uninstall the plugin.

src/Controllers/Admin/PluginsController.php
xuwhite: Please add this to the documentation @MyIgel

src/Plugins/Plugin.php
xuwhite: Please add a comment to this to explain it.

MyIgel:
> when two plugin register the same route you get a `FastRoute \ BadRouteException`which makes sense, but you cannot easily recover, as the the plugin config page can no longer be opened and you cannot disable or uninstall the plugin.

I'll add some graceful error handling & logging here at some later point :D

PR #1694, opened 2026-06-06
Fix multiple error messages

MyIgel:
* Fix sorting on users list
* Fix admin active when shift has no time
* Fix shifts when shift created_at is not set
* Imply hint permission when view is granted
* Fix error when themes are not configured


xuwhite:
please add a description

includes/controller/users_controller.php
```diff
@@ -258,9 +258,9 @@ function users_list_controller()
         'arrived' => 'arrived',
         'got_voucher' => 'users_state.got_voucher',
         'active' => 'users_state.active',
-        'force_active' => 'users_state.force_active',
-        'force_food' => 'users_state.force_food',
-        'got_goodie' => 'users_state.got_goodie',
+        'force_active' => 'users_state.force_active_by',
```
xuwhite, line 261:
```suggestion
        'force_active_by' => 'users_state.force_active_by',
```
Please make it consistent.

MyIgel: Not part of this fix

xuwhite: pleeeeeeaaaase

MyIgel: nope

